PR #83, opened 2021-06-24
TRA-3317 Tap and show outbound traffic

nimrod-up9:
image

Adds `--direction` flag to `mizu tap` command.
With `--direction any` outgoing traffic from tapped pods is recorded and displayed.
With `--direction in` only ingoing traffic is recorded. This is the same behavior we had before. It is the default.


cli/kubernetes/provider.go
```diff
@@ -226,19 +226,30 @@ func (provider *Provider) RemoveDaemonSet(ctx context.Context, namespace string,
 	return provider.clientSet.AppsV1().DaemonSets(namespace).Delete(ctx, daemonSetName, metav1.DeleteOptions{})
 }
 
-func (provider *Provider) ApplyMizuTapperDaemonSet(ctx context.Context, namespace string, daemonSetName string, podImage string, tapperPodName string, aggregatorPodIp string, nodeToTappedPodIPMap map[string][]string, linkServiceAccount bool) error {
+func (provider *Provider) ApplyMizuTapperDaemonSet(ctx context.Context, namespace string, daemonSetName string, podImage string, tapperPodName string, aggregatorPodIp string, nodeToTappedPodIPMap map[string][]string, linkServiceAccount bool, direction string) error {
```
gadotroee, line 229: I think it will be more clear of the parameter here will be boolean to say if we want anydirection.. 
and not comparing to "any" inside..

gadotroee (commented):
one comment and one thing we will talk about

gadotroee (approved):

gadotroee (approved):

PR #85, opened 2021-06-27
Fix crash when there are no pods matching the regex

nimrod-up9:
1. CLI will not crash if there are no pods matching the regex. Instead it will print a warning and continue to watch the pods.
2. If the number of tapped pods decreases to 0 (at a later stage), CLI will delete mizu-tapper daemonset.
3. Attempting to delete resources that are not installed will not cause an error.

### DIFF
--- a/cli/cmd/tapRunner.go
+++ b/cli/cmd/tapRunner.go
@@ -113,19 +113,26 @@ func getMizuApiFilteringOptions(tappingOptions *MizuTapOptions) (*shared.Traffic
 }
 
 func createMizuTappers(ctx context.Context, kubernetesProvider *kubernetes.Provider, nodeToTappedPodIPMap map[string][]string, tappingOptions *MizuTapOptions) error {
-	if err := kubernetesProvider.ApplyMizuTapperDaemonSet(
-		ctx,
-		mizu.ResourcesNamespace,
-		mizu.TapperDaemonSetName,
-		tappingOptions.MizuImage,
-		mizu.TapperPodName,
-		fmt.Sprintf("%s.%s.svc.cluster.local", aggregatorService.Name, aggregatorService.Namespace),
-		nodeToTappedPodIPMap,
-		mizuServiceAccountExists,
-		tappingOptions.Direction,
-	); err != nil {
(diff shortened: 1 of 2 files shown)


cli/kubernetes/provider.go
```diff
@@ -102,7 +102,6 @@ func (provider *Provider) CreateMizuAggregatorPod(ctx context.Context, namespace
 			},
 			DNSPolicy:                     core.DNSClusterFirstWithHostNet,
 			TerminationGracePeriodSeconds: new(int64),
-			// Affinity: TODO: define node selector for all relevant nodes for this mizu instance
```
gadotroee, line 105: why do you remove this comment? 
I think we do need it (maybe it is already done?)

gadotroee: ok I see it is done

cli/cmd/tapRunner.go
```diff
@@ -45,6 +46,22 @@ func RunMizuTap(podRegexQuery *regexp.Regexp, tappingOptions *MizuTapOptions) {
 		currentlyTappedPods = matchingPods
 	}
 
+	var namespacesStr string
+	if targetNamespace != "" {
```
gadotroee, line 50: I think you can use mizu.K8sAllNamespaces (the const)

cli/cmd/tapRunner.go
```diff
@@ -45,6 +46,22 @@ func RunMizuTap(podRegexQuery *regexp.Regexp, tappingOptions *MizuTapOptions) {
...
+		namespacesStr = fmt.Sprintf("namespace \"%s\"", targetNamespace)
+	} else {
+		namespacesStr = "all namespaces"
+	}
+	fmt.Printf("Tapping pods in %s\n", namespacesStr)
+
+	if len(currentlyTappedPods) == 0 {
+		var suggestionStr string
+		if targetNamespace != "" {
```
gadotroee, line 59: here too (change ot mizu.K8sAllNamespaces)

gadotroee (commented):
small comment

gadotroee (approved):

PR #104, opened 2021-07-12
Mizu tap - adding logs for troubleshooting

IgorGov:


cli/cmd/tapRunner.go
```diff
@@ -337,7 +340,7 @@ func syncApiStatus(ctx context.Context, cancel context.CancelFunc, tappingOption
 		default:
 			err = controlSocket.SendNewTappedPodsListMessage(currentlyTappedPods)
 			if err != nil {
-				fmt.Printf("error Sending message via control socket %s\n", err)
+				fmt.Printf("error Sending message via control socket %v, error: %s\n", controlSocketStr, err)
```
gadotroee, line 343: maybe this should be debug

gadotroee: Just because it is CLI and it will be visible to the user (and I'm not sure we want to write the websocket url

gadotroee (approved):

gadotroee (changes requested):
Are you sure you commit everything? 
I think the we missing changes in go.mod / go.sum in cli

cli/cmd/tapRunner.go
IgorGov: changed it to debug

gadotroee (approved):

PR #107, opened 2021-07-13
TRA-3437 long running

RamiBerm:


api/pkg/database/main.go
```diff
@@ -34,12 +33,19 @@ var (
...
+}
+
 func GetEntriesTable() *gorm.DB {
 	return DB.Table("mizu_entries")
 }
 
 func initDataBase(databasePath string) *gorm.DB {
-	temp, _ := gorm.Open(sqlite.Open(databasePath), &gorm.Config{})
+	temp, _ := gorm.Open(sqlite.Open(databasePath), &gorm.Config{
+		Logger: logger.Default.LogMode(logger.Silent),
```
RamiBerm, line 47: gorm's logger prints full entry objects on expected errors (locked db when pruning), this makes mizu stuck

api/main.go
```diff
@@ -126,18 +127,38 @@ func getTrafficFilteringOptions() *shared.TrafficFilteringOptions {
...
 }
 
+func isHealthCheckByUserAgent(message *tap.OutputChannelItem) bool {
+	for _, header := range message.HarEntry.Request.Headers {
+		if strings.ToLower(header.Name) == "user-agent" {
+			for _, userAgent := range userAgentsToFilter {
+				if strings.Contains(strings.ToLower(header.Value), userAgent) {
+					return true
+				}
+				return false
```
nimrod-up9, line 152: Should be outside of the inner for loop.

api/pkg/api/main.go
```diff
@@ -179,3 +180,22 @@ func getServiceNameFromUrl(inputUrl string) (string, string) {
 func CheckIsServiceIP(address string) bool {
 	return k8sResolver.CheckIsServiceIP(address)
 }
+
+// gives a rough estimate of the size this will take up in the db, good enough for maintaining db size limit accurately
+func getEstimatedEntrySizeBytes(mizuEntry models.MizuEntry) int {
```
nimrod-up9, line 185: This is going to get out of sync with the real size as soon as we add more fields in the DB.
For example, `getEstimatedEntrySizeBytes` does not take `EstimatedSizeBytes` into account.

api/pkg/database/main.go
nimrod-up9: What other way do we have to know about the errors?

api/pkg/database/size_enforcer.go
```diff
@@ -0,0 +1,119 @@
...
+	"time"
+)
+
+const percentageOfMaxSizeBytesToPrune = 15
+const defaultMaxDatabaseSizeBytes = 200 * 1000 * 1000
+
+func StartEnforcingDatabaseSize() {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Printf("Error creating filesystem watcher for db size enforcement: %v\n", err) // TODO: make fatal
```
nimrod-up9, line 21: Jira ticket instead of TODO.

api/pkg/database/size_enforcer.go
```diff
@@ -0,0 +1,119 @@
...
+				fmt.Printf("filesystem watcher encountered error:%v\n", err)
+			}
+		}
+	}()
+
+	err = watcher.Add(DBPath)
+	if err != nil {
+		log.Printf("Error adding %s to filesystem watcher for db size enforcement: %v\n", DBPath, err) //TODO: make fatal
+	}
+	<-done
```
nimrod-up9, line 60: Why does `StartEnforcingDatabaseSize` need to hang around after creating the go routines and the watcher?

api/pkg/database/size_enforcer.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+		var entry models.MizuEntry
+		err = DB.ScanRows(rows, &entry)
+		if err != nil {
+			fmt.Printf("Error scanning db row: %v\n", err)
+			continue
+		}
+
+		entryIdsToRemove = append(entryIdsToRemove, entry.ID)
+		bytesToBeRemoved += int64(entry.EstimatedSizeBytes)
```
nimrod-up9, line 108: Why use EstimatedSizeBytes? Can't we get the exact size of the row? Is it slower?

shared/go.mod
```diff
@@ -2,4 +2,8 @@ module github.com/up9inc/mizu/shared
 
 go 1.16
 
-require github.com/gorilla/websocket v1.4.2
+require (
+    github.com/gorilla/websocket v1.4.2
```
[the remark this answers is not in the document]
RamiBerm: socket_client.go contains shared code that uses github.com/gorilla/websocket v1.4.2

api/pkg/database/size_enforcer.go
RamiBerm: changed to fatal as this no longer needs to be immediately demoed

api/pkg/database/size_enforcer.go
RamiBerm: fair point

api/pkg/database/size_enforcer.go
RamiBerm: Its "estimated" because I doubt sqlite rows are exactly the size of their contents in bytes, there's likely some overhead and there may also be some overhead per column.

Haven't found a way to get accurate byte count of rows and this estimated count works well enough

api/pkg/database/main.go
RamiBerm: Implemented a truncating logger

api/pkg/database/size_enforcer.go
```diff
@@ -0,0 +1,119 @@
...
+
+	done := make(chan bool)
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return // closed channel
+				}
+				if event.Op&fsnotify.Write == fsnotify.Write {
```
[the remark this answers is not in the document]
RamiBerm: yep fixed

shared/go.mod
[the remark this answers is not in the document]
nimrod-up9: ok, missed that.

cli/cmd/tap.go
```diff
@@ -49,12 +50,13 @@ Supported protocols are HTTP and gRPC.`,
 			return errors.New(fmt.Sprintf("%s is not a valid regex %s", args[0], err))
 		}
 
-		mizuTapOptions.MaxEntriesDBSizeBytes, err = shared.HumanReadableToBytes(humanMaxEntriesDBSize)
+		mizuTapOptions.MaxEntriesDBSizeBytes = 200 * 1000 * 1000
```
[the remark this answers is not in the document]
RamiBerm: Whoops good catch 😅

nimrod-up9 (dismissed):

nimrod-up9 (approved):

PR #109, opened 2021-07-13
TRA-3440 - add confirmation message and info on hover (tooltip) chagne analyze to analysis

gadotroee:


api/main.go
```diff
@@ -83,6 +84,11 @@ func main() {
 func hostApi(socketHarOutputChannel chan<- *tap.OutputChannelItem) {
 	app := fiber.New()
 
+	app.Use(cors.New(cors.Config{
+		AllowOrigins: "*",
+		AllowMethods: "*",
+		AllowHeaders: "*",
+	}))
```
nimrod-up9, line 91: I don't know enough about CORS to tell if this is somehow a security risk.
I guess that because Mizu server does not authenticate, then we are not making anything worse by allowing CORS?

cli/cmd/tap.go
```diff
@@ -26,22 +27,32 @@ type MizuTapOptions struct {
...
 var tapCmd = &cobra.Command{
 	Use:   "tap [POD REGEX]",
 	Short: "Record ingoing traffic of a kubernetes pod",
 	Long: `Record the ingoing traffic of a kubernetes pod.
 Supported protocols are HTTP and gRPC.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		RunMizuTap(regex, mizuTapOptions)
+		return nil
+	},
+	PreRunE: func(cmd *cobra.Command, args []string) error {
```
nimrod-up9, line 45: 👍

cli/uiUtils/confirmation.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"bufio"
+	"fmt"
+	"github.com/up9inc/mizu/cli/mizu"
+	"log"
+	"os"
+	"strings"
+)
+
+func AskForConfirmation(s string) bool {
```
nimrod-up9, line 12: It's great that you made a reusable function.

cli/uiUtils/confirmation.go
```diff
@@ -0,0 +1,26 @@
...
+	reader := bufio.NewReader(os.Stdin)
+
+	fmt.Printf(mizu.Magenta, s)
+
+	response, err := reader.ReadString('\n')
+	if err != nil {
+		log.Fatal(err)
+	}
+	response = strings.ToLower(strings.TrimSpace(response))
+	if response == "" || response == "y" || response == "yes" {
```
nimrod-up9, line 22: I would also accept "Y".

ui/src/App.tsx
```diff
@@ -3,23 +3,78 @@ import './App.sass';
...
 import {HarPage} from "./components/HarPage";
+import Tooltip from "./components/Tooltip";
+import {makeStyles} from "@material-ui/core/styles";
 
 
+const useStyles = makeStyles(() => ({
+    tooltip: {
+        backgroundColor: "#3868dc",
+        color: "white",
+        fontSize: 13,
```
nimrod-up9, line 14: I think we usually use css or sass for styles.

ui/src/components/HarPage.tsx
```diff
@@ -43,16 +43,24 @@ const isKubeProxy = () => {
     return window.location.href.indexOf("/api/v1/namespaces/") > -1;
 }
 
+const isLocalDevelopment = () => {
+    return window.location.href.indexOf(":3000") > -1;
+}
```
nimrod-up9, line 48: Not a very robust criterion.

ui/src/components/HarPage.tsx
```diff
@@ -43,16 +43,24 @@ const isKubeProxy = () => {
...
+
 const getMizuApiUrl = () => {
     if (isKubeProxy()) {
         return window.location.href;
+    } else if (isLocalDevelopment) {
+        return "http://localhost:8899/api/v1/namespaces/default/services/mizu-collector:80/proxy"
     }
     return window.location.origin;
 };
```
nimrod-up9, line 57: What's going on here?

ui/src/components/Tooltip.tsx
```diff
@@ -4,6 +4,7 @@ import React from "react";
 export interface TooltipProps extends MUITooltipProps {
     variant?: 'default' | 'wide' | 'fit';
     isSimple?: boolean;
+    classes?: any;
```
nimrod-up9, line 7: Can you be more specific than `any`?

cli/uiUtils/confirmation.go
gadotroee: it is lowered before

api/main.go
gadotroee: yes, it is mainly to be able to access mizu when starting the site locally and want to work against server that running in another port

ui/src/App.tsx
gadotroee: I think we are doing it like this sometimes when it is done using useStyles

ui/src/components/HarPage.tsx
gadotroee: not relevant now

ui/src/components/Tooltip.tsx
gadotroee: I see it is any in our site a lot when using the classes

api/pkg/controllers/entries_controller.go
```diff
@@ -150,6 +150,10 @@ func UploadEntries(c *fiber.Ctx) error {
 	if err := validation.Validate(uploadRequestBody); err != nil {
 		return c.Status(fiber.StatusBadRequest).JSON(err)
 	}
+	fmt.Println("---a---")
+	fmt.Println(uploadRequestBody)
+	fmt.Println("---b---")
```
nimrod-up9, line 155: Debugging leftovers?

cli/cmd/tapRunner.go
```diff
@@ -261,6 +261,7 @@ func portForwardApiPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
 					}
 					rlog.Debugf("Sending get request to %v\n", u.String())
 					if response, err := http.Get(u.String()); err != nil || response.StatusCode != 200 {
+						fmt.Printf(u.String())
```
nimrod-up9, line 264: Debugging leftover?

If this is a print you want to keep, then it needs to be clear what is being printed here.
What is u?
```go
fmt.Printf("Sending traffic to: %s\n", u.String())
```

cli/kubernetes/proxy.go
```diff
@@ -40,15 +39,26 @@ func StartProxy(kubernetesProvider *Provider, mizuPort uint16, mizuNamespace str
 	return server.Serve(l)
 }
 
-func GetMizuCollectorProxiedHostAndPath(mizuPort uint16, mizuNamespace string, mizuServiceName string) string {
-	return fmt.Sprintf("localhost:%d/api/v1/namespaces/%s/services/%s:80/proxy", mizuPort, mizuNamespace, mizuServiceName)
+const mizuServicePort = 80
```
nimrod-up9, line 42: Consts belong at the top of the file.

cli/kubernetes/proxy.go
```diff
@@ -40,15 +39,26 @@ func StartProxy(kubernetesProvider *Provider, mizuPort uint16, mizuNamespace str
...
+
+func getMizuCollectorProxiedHostAndPath(mizuNamespace string, mizuServiceName string) string {
+	return fmt.Sprintf("/api/v1/namespaces/%s/services/%s:%v/proxy/", mizuNamespace, mizuServiceName, mizuServicePort)
```
nimrod-up9, line 45: Use `%d` here instead of `%v`, because this is not just a log message. It is important that we print an int here.
If for some reason `mizuServicePort` changes to an object, then we will notice it.

nimrod-up9 (approved):

PR #120, opened 2021-07-18
Missing request body

nimrod-up9:
For requests with multipart/form-data or application/x-www-form-urlencoded media-types, martian/har will parse the request body and place the parameters in harRequest.PostData.Params instead of harRequest.PostData.Text (as the HAR spec requires it). Mizu currently only looks at PostData.Text. Therefore, instead of letting martian/har set the content of PostData, always copy the request body to PostData.Text.

### DIFF
--- a/tap/har_writer.go
+++ b/tap/har_writer.go
@@ -1,9 +1,11 @@
 package tap
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


gadotroee:
I see there is params in the HAR spec.. 
http://www.softwareishard.com/blog/har-12-spec/#params
Am I missing something?

gadotroee (approved):

PR #124, opened 2021-07-20
Adding telemetry reports

IgorGov:


cli/mizu/telemetry.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/romana/rlog"
+	"net/http"
+)
+
+func ReportRun(cmd string, args interface{}) {
+	if Branch != "main" {
+		rlog.Debugf("reporting only on main branch")
+		return
+	}
+	url := "https://us-east4-up9-prod.cloudfunctions.net/mizu-telemetry"
```
RamiBerm, line 16: move to const outside of func

RamiBerm, line 16: move to const outside of func

RamiBerm (changes requested):

RamiBerm (approved):

PR #126, opened 2021-07-21
Add option to disable sensitive data redaction by flag and fix CPU leaks

RamiBerm:
* Introduces the `--no-redact` flag to the `tap` command
* Update ikisocket depedency to latest version (fixes a serious cpu leak)
* Fix potential CPU leak with time.After


cli/cmd/tapRunner.go
```diff
@@ -269,7 +270,7 @@ func portForwardApiPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
 				}
 			}
 
-		case <-time.After(25 * time.Second):
+		case <- timeAfter:
```
IgorGov, line 273: this is the main fix for CPU leaks, right?

IgorGov (approved):

api/go.mod
```diff
@@ -3,14 +3,15 @@ module mizuserver
...
+	github.com/antoniodipinto/ikisocket v0.0.0-20210719144512-dce3f3fbbd04
 	github.com/beevik/etree v1.1.0
 	github.com/djherbis/atime v1.0.0
 	github.com/fasthttp/websocket v1.4.3-beta.1 // indirect
+	github.com/fsnotify/fsnotify v1.4.9
 	github.com/go-playground/locales v0.13.0
 	github.com/go-playground/universal-translator v0.17.0
 	github.com/go-playground/validator/v10 v10.5.0
-	github.com/gofiber/fiber/v2 v2.8.0
+	github.com/gofiber/fiber/v2 v2.10.0
```
IgorGov, line 14: why not latest v2.15.0?

gadotroee (approved):

api/go.mod
RamiBerm: We're bound to the fiber version that ikisocket uses

cli/cmd/tapRunner.go
RamiBerm: The main issue was fixed in [ikisocket itself](https://github.com/antoniodipinto/ikisocket/pull/15/files), I suspect that time.After call here also causes a leak, if it does its a much smaller one

IgorGov (approved):

PR #131, opened 2021-07-22
UI api helper

lirazyehezkel:


ui/src/helpers/api.js
```diff
@@ -0,0 +1,51 @@
...
+// When working locally (with npm run start) change to:
+// export const MizuWebsocketURL = `ws://localhost:8899${mizuAPIPathPrefix}/ws`;
+export const MizuWebsocketURL = `ws://${window.location.host}${mizuAPIPathPrefix}/ws`;
+
+export default class Api {
+    constructor() {
+        this.client = null;
+        // When working locally (with npm run start) change to:
+        // this.api_url = `http://localhost:8899/${mizuAPIPathPrefix}/api/`;
+        this.api_url = `${window.location.origin}${mizuAPIPathPrefix}/api/`;
```
gadotroee, line 14: camelCase?

ui/src/helpers/api.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    init = () => {
```
gadotroee, line 17: why not init this in the constructor and then just use this.client (that will be already initialize)?

gadotroee (changes requested):
small comments

gadotroee (approved):

PR #134, opened 2021-07-24
Introducing new logger, logging debug to file and info to stderr

IgorGov:


gadotroee:
I see you changed from rlog. 
I opened PR that added "global flag" and it is nice in the way of adding debug so we probably want to merge those changes somehow (I think you solution is better - just need to understand how will it work with my changes (that are in progres now).

Also mizu config file will be names .mizu and I see you added .mizu folder.. it might cause conflict in some way so lets check and see if the config file should move or the log should move

gadotroee:
one thing that I want to think about from now (even if we are not implement it) is log rotation..
I don't want to have increasing file with no limit (although it should not increase too much)
maybe it can be easy to get this rotation (max time to save logs / max size for the log).

cli/cmd/tapRunner.go
```diff
@@ -324,19 +323,19 @@ func portForwardApiPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
-					rlog.Debugf("Sending get request to %v\n", u.String())
+					mizu.Log.Debugf("Sending get request to %v\n", u.String())
 					if response, err := http.Get(u.String()); err != nil || response.StatusCode != 200 {
-						fmt.Printf("error sending upload entries req, status code: %v, err: %v\n", response.StatusCode, err)
+						mizu.Log.Infof("error sending upload entries req, status code: %v, err: %v\n", response.StatusCode, err)
 					} else {
-						fmt.Printf(mizu.Purple, "Traffic is uploading to UP9 for further analsys")
-						fmt.Println()
+						mizu.Log.Infof(mizu.Purple, "Traffic is uploading to UP9 for further analsys")
+						mizu.Log.Info()
```
gadotroee, line 331: change to \n in the previous line

gadotroee (approved):

PR #149, opened 2021-07-28
TRA-3420 inform user of tls traffic

RamiBerm:


tap/passive_tapper.go
```diff
@@ -429,6 +429,21 @@ func startPassiveTapper(harWriter *HarWriter, outboundLinkWriter *OutboundLinkWr
...
 			assembler.AssembleWithContext(packet.NetworkLayer().NetworkFlow(), tcp, &c)
 			assemblerMutex.Unlock()
+
+			sourceIP := packet.NetworkLayer().NetworkFlow().Src().String()
+			destinationIP := packet.NetworkLayer().NetworkFlow().Dst().String()
+			destinationPort := int(tcp.DstPort)
+
+			if streamFactory.getStreamProps(sourceIP, destinationIP, destinationPort).isTapTarget {
+				fmt.Printf("%s -> %s:%d isTapTarget:%t", sourceIP, destinationIP, destinationPort, streamFactory.getStreamProps(sourceIP, destinationIP, destinationPort).isTapTarget)
+				if IsTLSHandshakePacket(tcp.Payload) {
```
nimrod-up9, line 439: Won't there only be a handshake when the there's a new stream?
Why do we need to run this for every single tcp packet?
Also, is it possible for the handshake to be split between several tcp packets? In that case we would need to look for the handshake after tcp assembly.

ui/src/App.tsx
```diff
@@ -15,11 +17,40 @@ const useStyles = makeStyles(() => ({
...
+    const [addressesWithTLS, setAddressesWithTLS] = useState(new Set());
+
+    useEffect(() => {
+        (async () => {
+            const recentTLSLinks = await api.getRecentTLSLinks();
+
+            if (recentTLSLinks?.length > 0) {
+                for (const tlsDestAddress of recentTLSLinks) {
+                    addressesWithTLS.add(tlsDestAddress)
+                }
```
nimrod-up9, line 38:
```typescript
addressesWithTLS = [...addressesWithTLS, ...tlsDestAddress]
```

ui/src/App.tsx
```diff
@@ -15,11 +17,40 @@ const useStyles = makeStyles(() => ({
...
+                setAddressesWithTLS(new Set(addressesWithTLS));
+                setShowTLSWarning(true);
+            }
+
+        })();
+    }, []);
+
+    const onTLSDetected = (destAddress: string) => {
+        addressesWithTLS.add(destAddress);
+        setAddressesWithTLS(new Set(addressesWithTLS));
```
nimrod-up9, line 48: Why do we need two ways to `setAddressesWithTLS`?

tap/passive_tapper.go
```diff
@@ -429,6 +429,21 @@ func startPassiveTapper(harWriter *HarWriter, outboundLinkWriter *OutboundLinkWr
 			assemblerMutex.Lock()
 			assembler.AssembleWithContext(packet.NetworkLayer().NetworkFlow(), tcp, &c)
 			assemblerMutex.Unlock()
+
+			sourceIP := packet.NetworkLayer().NetworkFlow().Src().String()
+			destinationIP := packet.NetworkLayer().NetworkFlow().Dst().String()
+			destinationPort := int(tcp.DstPort)
+
+			if streamFactory.getStreamProps(sourceIP, destinationIP, destinationPort).isTapTarget {
+				fmt.Printf("%s -> %s:%d isTapTarget:%t", sourceIP, destinationIP, destinationPort, streamFactory.getStreamProps(sourceIP, destinationIP, destinationPort).isTapTarget)
```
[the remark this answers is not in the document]
RamiBerm: whoops

ui/src/App.tsx
RamiBerm: This is for notifications from websocket, the previous one is the result of the http request done on page load

nimrod-up9 (approved):

PR #156, opened 2021-08-02
Print Mizu URL only after the watched pods

IgorGov:


cli/cmd/tapRunner.go
```diff
@@ -345,6 +335,25 @@ func portForwardApiPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
 	}
 }
 
+func requestForAnalysis(tappingOptions *MizuTapOptions) {
```
gadotroee, line 338: I liked that it change to function

gadotroee (approved):

PR #160, opened 2021-08-03
TRA-3415 ignore unready pods

RamiBerm:


cli/kubernetes/provider.go
```diff
@@ -635,3 +635,7 @@ func loadKubernetesConfiguration(kubeConfigPath string) clientcmd.ClientConfig {
 		},
 	)
 }
+
+func IsPodRunning(pod *core.Pod) bool {
```
IgorGov, line 639: why starts with capital letter? is it exposed method?

cli/cmd/tapRunner.go
```diff
@@ -257,19 +253,15 @@ func watchPodsForTapping(ctx context.Context, kubernetesProvider *kubernetes.Pro
 		}
 	}
 	restartTappersDebouncer := debounce.NewDebouncer(updateTappersDelay, restartTappers)
-	timer := time.AfterFunc(time.Second*10, func() {
-		mizu.Log.Debugf("Waiting for URL...")
+
+	go func() {
```
IgorGov, line 257: this one should be moved after the first call to "updateCurrentlyTappedPods"

cli/kubernetes/provider.go
RamiBerm: fixed

cli/cmd/tapRunner.go
RamiBerm: fixed

IgorGov (approved):

PR #161, opened 2021-08-03
extracted mizu folder to const

RoyUP9:


gadotroee:
Done as part of https://github.com/up9inc/mizu/pull/162

PR #162, opened 2021-08-03
Adding logs and fixing several issues

IgorGov:


cli/cmd/config.go
```diff
@@ -29,6 +29,5 @@ var configCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(configCmd)
-
-	configCmd.Flags().StringVarP(&outputFileName, "file", "f", "", "Save content to local file")
+	configCmd.Flags().BoolVarP(&regenerateFile, "file", "f", false, "Save content to local file")
```
gadotroee, line 32: I thought it will be nice to have -f to save to file, but wit can be achieved using " > filepath.yaml", so the only change that can be is to rename the flag because it will be better to call it regenerae IMO, but I will accept this as well

cli/mizu/config.go
```diff
@@ -29,14 +29,14 @@ func InitConfig(cmd *cobra.Command) error {
 	}
 
 	if err := mergeConfigFile(); err != nil {
-		Log.Infof(uiUtils.Red, "Invalid config file")
-		return err
+		Log.Errorf("Could not load config file, error %v", err)
+		Log.Fatalf("You can regenerate the file using `mizu config -f` or just remove it %v", GetConfigFilePath())
 	}
```
gadotroee, line 34: If you are changing the flag you will need to fix this as well

cli/mizu/logger.go
```diff
@@ -35,5 +34,6 @@ func InitLogger() {
 
 	logging.SetBackend(backend1Leveled, backend2Formatter)
 
+	Log.Debugf(" ===================== \n\n\n\n")
```
gadotroee, line 37: Maybe \n twice is enough.. 
Also Im thinking if we need to know the origin of this (of mizu is running and user start another mizu in second terminal it can confuse us) - maybe the solution will be to add something to the log when we are doing this Init and it will be written in the logs and we will have context - even process ID or info about the process will help because it will be another one in the second command

IgorGov: in the log file we have process id

IgorGov: https://github.com/up9inc/mizu/blob/1ab45d4e2ce0b0825c165fc23e8947a9ff2f4b1d/cli/mizu/logger.go#L13

cli/cmd/tapRunner.go
```diff
@@ -53,34 +53,31 @@ func RunMizuTap() {
...
+		return
+	}
+
 	if len(currentlyTappedPods) == 0 {
 		var suggestionStr string
 		if targetNamespace != mizu.K8sAllNamespaces {
 			suggestionStr = "\nSelect a different namespace with -n or tap all namespaces with -A"
 		}
 		mizu.Log.Infof("Did not find any pods matching the regex argument%s", suggestionStr)
+		return
```
gadotroee, line 76: why do we return here? 
maybe he want to keep it and pod will start later?

cli/cmd/tapRunner.go
```diff
@@ -53,34 +53,31 @@ func RunMizuTap() {
...
-		mizu.Log.Infof("Mizu is available at http://%s", <-urlReadyChan)
-	}()
-
 	var namespacesStr string
 	if targetNamespace != mizu.K8sAllNamespaces {
 		namespacesStr = fmt.Sprintf("namespace \"%s\"", targetNamespace)
 	} else {
 		namespacesStr = "all namespaces"
 	}
+	go mizu.CheckNewerVersion()
```
gadotroee, line 62: Is. it ok that it is running now in another routine? (will it write the message of new version in the middle?)

cli/mizu/logger.go
gadotroee: Oh, my bad for not checking it..

cli/mizu/config.go
```diff
@@ -29,33 +29,34 @@ func InitConfig(cmd *cobra.Command) error {
...
 
 	if err := mergeConfigFile(); err != nil {
-		Log.Infof(uiUtils.Red, "Invalid config file")
-		return err
+		Log.Errorf("Could not load config file, error %v", err)
+		Log.Fatalf("You can regenerate the file using `mizu config -f` or just remove it %v", GetConfigFilePath())
```
gadotroee, line 33: -r

cli/cmd/tapRunner.go
IgorGov: it not what actually happen, this is how the code behaves, I just made it more clear for the next reader

cli/cmd/tapRunner.go
IgorGov: good catch

gadotroee (approved):

PR #166, opened 2021-08-04
Tapper stats in stats tracker

gadotroee:


tap/stats_tracker.go
```diff
@@ -2,34 +2,51 @@ package tap
 
 import (
 	"sync"
+	"time"
 )
 
 type AppStats struct {
-	matchedMessages int
+	MatchedMessages      int       `json:"matchedMessages"`
+	PacketsCount         int       `json:"packetsCount"`
```
IgorGov, line 10: int64

IgorGov: totalPacketsCount

IgorGov: can we measure the packets count from last dump?

IgorGov (approved):

PR #169, opened 2021-08-05
added allowed set flag

RoyUP9:


cli/mizu/config.go
```diff
@@ -95,7 +105,7 @@ func initFlag(f *pflag.Flag) {
 
 	if f.Name == SetCommandName {
 		if setError := mergeSetFlag(sliceValue.GetSlice()); setError != nil {
-			Log.Infof(uiUtils.Red, "Invalid set argument")
+			Log.Warningf(uiUtils.Red, fmt.Sprintf("Invalid set argument, error: %v", setError))
```
IgorGov, line 108: we can print here the available options

IgorGov: allowedSetFlags

IgorGov: The log will look like:
`Invalid set argument, error: invalid set key AAA`
Can we remove the duplication, make more user friendly

cli/mizu/configStruct.go
```diff
@@ -6,6 +6,13 @@ import (
 	"github.com/up9inc/mizu/cli/mizu/configStructs"
 )
 
+const (
+	AgentImageConfigName             = "agent-image"
+	MizuResourcesNamespaceConfigName = "mizu-resources-namespace"
```
IgorGov, line 11: not related to this commit, but can you search for all "mizu-resources-namespace" accurances in the code 
they are mainly in errors or logs, and use this const there

cli/mizu/config.go
```diff
@@ -117,6 +127,11 @@ func mergeSetFlag(setValues []string) error {
 		}
 
 		argumentKey, argumentValue := split[0], split[1]
+
+		if !Contains(allowedSetFlags, argumentKey) {
+			return errors.New(fmt.Sprintf("invalid set key %s, allowed set flags: \"%s\"", argumentKey, strings.Join(allowedSetFlags, "\", \"")))
```
IgorGov, line 132: you call it in different names, "key" <-> "flag"

IgorGov (approved):

PR #173, opened 2021-08-06
Add option to supply user agents to ignore via config 

gadotroee:
right now it is only kube-probe and prometheus and now it will be default those and changed by user


cli/cmd/tap.go
```diff
@@ -65,7 +65,7 @@ func init() {
 	tapCmd.Flags().BoolP(configStructs.AllNamespacesTapName, "A", defaultTapConfig.AllNamespaces, "Tap all namespaces")
 	tapCmd.Flags().StringP(configStructs.KubeConfigPathTapName, "k", defaultTapConfig.KubeConfigPath, "Path to kube-config file")
 	tapCmd.Flags().StringArrayP(configStructs.PlainTextFilterRegexesTapName, "r", defaultTapConfig.PlainTextFilterRegexes, "List of regex expressions that are used to filter matching values from text/plain http bodies")
-	tapCmd.Flags().Bool(configStructs.HideHealthChecksTapName, defaultTapConfig.HideHealthChecks, "hides requests with kube-probe or prometheus user-agent headers")
+	tapCmd.Flags().StringArray(configStructs.IgnoredUserAgentsTapName, defaultTapConfig.HealthChecksUserAgentHeaders, "hides requests with kube-probe or prometheus user-agent headers")
```
IgorGov, line 68: Need to update the description

cli/mizu/configStructs/tapConfig.go
```diff
@@ -27,21 +27,21 @@ const (
...
+	AnalysisDestination          string   `yaml:"dest" default:"up9.app"`
+	SleepIntervalSec             int      `yaml:"upload-interval" default:"10"`
+	PodRegexStr                  string   `yaml:"regex" default:".*"`
+	GuiPort                      uint16   `yaml:"gui-port" default:"8899"`
+	Namespaces                   []string `yaml:"namespaces"`
+	Analysis                     bool     `yaml:"analysis" default:"false"`
+	AllNamespaces                bool     `yaml:"all-namespaces" default:"false"`
+	KubeConfigPath               string   `yaml:"kube-config"`
+	PlainTextFilterRegexes       []string `yaml:"regex-masking"`
+	HealthChecksUserAgentHeaders []string `yaml:"ignored-user-agents" default:"[\"kube-probe\",\"prometheus\"]"`
```
IgorGov, line 39: Notice that you are changing the functionality here.
The default was not ignoring the health checks

IgorGov, line 39: Need to update readme with an example how to filer these defaults

cli/cmd/tap.go
gadotroee: updated

cli/mizu/configStructs/tapConfig.go
gadotroee: I know I forget to change it back -  the default should be empty array

gadotroee: updated with example and explanation

cli/cmd/tap.go
gadotroee: I ended up with removing this flag (this option will be available only with editing from the config file)

cli/mizu/configStructs/tapConfig.go
```diff
@@ -27,21 +26,21 @@ const (
...
+	AnalysisDestination          string   `yaml:"dest" default:"up9.app"`
+	SleepIntervalSec             int      `yaml:"upload-interval" default:"10"`
+	PodRegexStr                  string   `yaml:"regex" default:".*"`
+	GuiPort                      uint16   `yaml:"gui-port" default:"8899"`
+	Namespaces                   []string `yaml:"namespaces"`
+	Analysis                     bool     `yaml:"analysis" default:"false"`
+	AllNamespaces                bool     `yaml:"all-namespaces" default:"false"`
+	KubeConfigPath               string   `yaml:"kube-config"`
+	PlainTextFilterRegexes       []string `yaml:"regex-masking"`
+	HealthChecksUserAgentHeaders []string `yaml:"ignored-user-agents" default:"[]"`
```
IgorGov, line 38: few comments:
* lets enable use to set it by using the --set command
* lets call it user agent black list
* in readme we can say something: "Mizu can ignore HTTP traffic by matching specific User-Agent header in request. For instance if you not interested seeing health checks from 'kube-probe' and 'prometheus' you can run 'mizu tap --set user-agent-bl=kube-probe --set user-agent-bl=prometheus

gadotroee: 1. I don't want to enable it vi a --set (it doesn't make any sense) [it will create confusion because it can't be the same name of parameter and we do not support it [in --set it should be singular `ignored-user-agent` and in settings it is `ignored-user-agents` (because in YAML we see list and in --set you are putting only one every tim
2. black list is term of the past we are not using black list and ignored is better IMO.
3. is not relevant after what I say in 

I can do the changes (it is not big) but let me know what do you thing

IgorGov (approved):

PR #174, opened 2021-08-08
Redesign test rules entry component

seltonfiuza:


ui/src/components/style/HarEntry.module.sass
```diff
@@ -24,12 +24,40 @@
...
+
+.ruleFailureRowSelected
   border: 1px $failure-color solid
   border-left: 5px $failure-color solid
+  margin-left: 10px
+  margin-right: 3px
+
+.ruleNumberTextFailure
+  color: #DB2156
+  font-family: Source Sans Pro;
```
lirazyehezkel, line 46: sass files shouldn't include ';' char

seltonfiuza: Fixed!

agent/pkg/models/models.go
```diff
@@ -218,7 +220,7 @@ func (fewp *FullEntryWithPolicy) UnmarshalData(entry *MizuEntry) error {
 
 func RunValidationRulesState(harEntry har.Entry, service string) ApplicableRules {
 	numberOfRules, resultPolicyToSend := rules.MatchRequestPolicy(harEntry, service)
-	statusPolicyToSend, latency := rules.PassedValidationRules(resultPolicyToSend, numberOfRules)
-	ar := NewApplicableRules(statusPolicyToSend, latency)
+	statusPolicyToSend, latency, numberOfRules := rules.PassedValidationRules(resultPolicyToSend, numberOfRules)
```
IgorGov, line 223: why "PassedValidationRules" can just return a "ApplicableRules" object?

agent/pkg/models/models.go
```diff
@@ -56,12 +56,14 @@ type BaseEntryDetails struct {
 type ApplicableRules struct {
 	Latency int64 `json:"latency,omitempty"`
 	Status  bool  `json:"status,omitempty"`
+	NumberOfRules int `json:"numberOfRules,omitempty"`
 }
 
-func NewApplicableRules(status bool, latency int64) ApplicableRules {
+func NewApplicableRules(status bool, latency int64, number int) ApplicableRules {
```
IgorGov, line 62: do we really need this function?
isn't it more simple and readable:
return ApplicableRules{ Status = status, Latency = latency, NumberOfRules = number)

IgorGov (approved):
Approving, will fix the comments in separate PR

PR #181, opened 2021-08-08
separated permissions section into separate file

haiut:
README modified

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,317 +39,15 @@ Pick one from the [Releases](https://github.com/up9inc/mizu/releases) page.
 
 ## Prerequisites
 1. Set `KUBECONFIG` environment variable to your Kubernetes configuration. If this is not set, Mizu assumes that configuration is at `${HOME}/.kube/config`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -39,317 +39,15 @@ Pick one from the [Releases](https://github.com/up9inc/mizu/releases) page.
...
-```
+For detailed list of k8s permissions see [PERMISSIONS](PERMISSIONS.md) document
 
-See `examples/roles` for example `clusterroles`. 
 
 ## How to Run
 
 1. Find pods you'd like to tap to in your Kubernetes cluster
-2. Run `mizu tap PODNAME` or `mizu tap REGEX` 
+2. Run `mizu tap` or `mizu tap PODNAME`
```
gadotroee, line 50: shouldn't it be pod regex?

RoyUP9 (approved):

README.md
haiut: correct, but I intentionally removed regex from this line to keep it easy to understand and simple
We already have example with regex below, so I don't think it's a problem

PR #200, opened 2021-08-11
Cli pkg refactor (2)

IgorGov:


cli/config/configStruct.go
```diff
@@ -0,0 +1,35 @@
+package config
+
+import (
+	"fmt"
+	configStructs2 "github.com/up9inc/mizu/cli/config/configStructs"
```
RoyUP9, line 5: why?

cli/config/config_test.go
```diff
@@ -1,7 +1,7 @@
-package mizu_test
+package config_test
 
 import (
-	"github.com/up9inc/mizu/cli/mizu"
+	config3 "github.com/up9inc/mizu/cli/config"
```
RoyUP9, line 4: why?

IgorGov: what do you mean why? the number 3 is bothering you?

RoyUP9 (approved):

PR #205, opened 2021-08-11
added set hierarchy, removed allowed set flags

RoyUP9:


cli/config/config.go
```diff
@@ -109,7 +103,10 @@ func initFlag(f *pflag.Flag) {
 
 	sliceValue, isSliceValue := f.Value.(pflag.SliceValue)
 	if !isSliceValue {
-		mergeFlagValue(configElem, f.Name, f.Value.String())
+		flagPath := getFlagPath(f.Name)
```
IgorGov, line 106:
```suggestion
		flagPath := []string {cmdName, f.Name}
```

IgorGov (approved):

PR #208, opened 2021-08-12
added resources to config

RoyUP9:


gadotroee (changes requested):
you requested to not approve

gadotroee (approved):

PR #212, opened 2021-08-15
Github actions workflow parallel improvements 

gadotroee:


gadotroee:
It will not work because each job runs in a new runner

PR #215, opened 2021-08-15
Improve cloud resources cleanup

nimrod-up9:
Always try to delete all resources.

Pseudocode:

```
for resource in resources:
    exists = k8s.get(resource)
    if exists:
        k8s.delete(resource)
Do this for all resources except namespace. If in “restricted namespace mode”, do not delete namespace because it is one of the user’s namespace.
```

Pseudocode with error handling:

```
for resource in resources:
    try:
        exists = k8s.get(resource)
        if exists:
            k8s.delete(resource)
    except err:
        if err != "failed to get resource":
            log(err)
```

If there are errors, print a summery when done deleting everything that could be deleted.
```
Failed to remove the following resources, for more info check logs at /Users/nimrod/.mizu/mizu_cli.log:
- Pod mizu-api-server in namespace user1
```
and in the log:
```
2021-09-22T17:33:52.376+03:00 DEBUG ‚> 5833 tapRunner.go:368 logDeletionError ‚> Error removing Pod mizu-api-server in namespace user1: insufficient permissions: pods "mizu-api-server" is forbidden: User "user1" cannot delete resource "pods" in API group "" in the namespace "user1". supply the required permission or control Mizu's access to namespaces by setting mizu-resources-namespace in the config file or setting the tapped namespace with --set mizu-resources-namespace=<NAMEPSACE>
```

"get errors" are only printed to the log file, not reported in CLI's stdout. Why? If the user doesn’t have permissions to manage a resource, they probably know it and expect mizu to work without it if it is an optional resource like clusterrole or configmap. “get” will fail at each mizu tap. It is unnecessary to alert every time. On the other hand it is not likely that a user has permissions to “create” a resource, but don’t have permissions to “get” that resource. Therefore, the situation where a user created a resource which they are unable to “get” is not likely to be common.


gadotroee (approved):

gadotroee (approved):

cli/cmd/tapRunner.go
```diff
@@ -252,77 +251,97 @@ func updateMizuTappers(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
+	if err := kubernetesProvider.RemoveRole(ctx, config.Config.MizuResourcesNamespace, mizu.RoleName); err != nil && !isErrForbiddenGet(err) {
+		level := getLogLevel(isErrForbiddenGet(err))
+		logger.Logf(level, uiUtils.Error, fmt.Sprintf("Error removing Role %s in namespace %s: %v", mizu.RoleName, config.Config.MizuResourcesNamespace, errormessage.FormatError(err)))
 	}
 
-	if !config.Config.IsNsRestrictedMode() {
-		waitUntilNamespaceDeleted(removalCtx, cancel, kubernetesProvider)
+	if err := kubernetesProvider.RemoveRoleBinding(ctx, config.Config.MizuResourcesNamespace, mizu.RoleBindingName); err != nil && !isErrForbiddenGet(err) {
+		level := getLogLevel(isErrForbiddenGet(err))
+		logger.Logf(level, uiUtils.Error, fmt.Sprintf("Error removing RoleBinding %s in namespace %s: %v", mizu.RoleBindingName, config.Config.MizuResourcesNamespace, errormessage.FormatError(err)))
```
IgorGov, line 325: why do we need "uiUtils.Error" if it will be printed to the file only "debug" (will look strange with additional chars)

cli/cmd/tapRunner.go
```diff
@@ -252,77 +251,97 @@ func updateMizuTappers(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
 	}
 }
 
+func isErrForbiddenGet(err error) bool {
+	if status := k8serrors.APIStatus(nil); errors.As(err, &status) {
+		return strings.Contains(status.Status().Message, "cannot get resource")
```
IgorGov, line 331: can't we check by status code?

cli/cmd/tapRunner.go
```diff
@@ -252,77 +251,97 @@ func updateMizuTappers(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
+			}
+		} else {
+			defer waitUntilNamespaceDeleted(ctx, cancel, kubernetesProvider)
 		}
+	}
 
-		if err := kubernetesProvider.RemoveService(removalCtx, config.Config.MizuResourcesNamespace, mizu.ApiServerPodName); err != nil {
-			logger.Log.Errorf(uiUtils.Error, fmt.Sprintf("Error removing Service %s in namespace %s: %v", mizu.ApiServerPodName, config.Config.MizuResourcesNamespace, errormessage.FormatError(err)))
-		}
+	if err := kubernetesProvider.RemovePod(ctx, config.Config.MizuResourcesNamespace, mizu.ApiServerPodName); err != nil {
```
IgorGov, line 288: can't we store what was already created, if we can't remove it (not matter of what reason) just print error?

cli/cmd/tapRunner.go
nimrod-up9: Status code is 403 for both GET and DELETE requests.
This is the jsonified status of a failed GET operation:
```json
  {
    "metadata": {},
    "status": "Failure",
    "message": "configmaps \"mizu-policy\" is forbidden: User \"user1\" cannot get resource \"configmaps\" in API group \"\" in the namespace \"user1\"",
    "reason": "Forbidden",
    "details": {
      "name": "mizu-policy",
      "kind": "configmaps"
    },
    "code": 403
  }
```
and this is the jsonified error of a failed DELETE operation:
```json
  {
    "metadata": {},
    "status": "Failure",
    "message": "clusterroles.rbac.authorization.k8s.io \"mizu-cluster-role\" is forbidden: User \"user1\" cannot get resource \"clusterroles\" in API group \"rbac.authorization.k8s.io\" at the cluster scope",
    "reason": "Forbidden",
    "details": {
      "name": "mizu-cluster-role",
      "group": "rbac.authorization.k8s.io",
      "kind": "clusterroles"
    },
    "code": 403
  }
```

The only difference is in the message.

cli/cmd/tapRunner.go
nimrod-up9: I want the cleanup to be as simple as possible.
It could be called at any stage, even before resource creation (there may be leftover resources from a previous run).

cli/cmd/tapRunner.go
nimrod-up9: 1. Where do you see additional chars? When I open the log file I see colors, not control chars.
2. We already print colors to the debug log. If this is an issue, then maybe we should write a new Log.Debug method that filters out non ASCII chars or control chars? [not in this PR]
3. I think it would be better to have `Log.Errorf` add the color, and not have `uiUtils.Error`spread all over the code. I'm not sure how to do this yet, and anyway it's for this PR.
4. I agree that it would be weird to have red message that start with the word "Error" but in DEBUG log level. On the other hand I tried to write generic reusable code.

gadotroee (dismissed):

IgorGov (approved):

PR #223, opened 2021-08-16
Feature/tra 3533 ssl connection pop up

imceZZ:


ui/src/App.tsx
```diff
@@ -118,8 +118,9 @@ const App = () => {
             </div>
             <HarPage setAnalyzeStatus={setAnalyzeStatus} onTLSDetected={onTLSDetected}/>
             <Snackbar open={showTLSWarning && !userDismissedTLSWarning}>
-                <MuiAlert elevation={6} variant="filled" onClose={() => setUserDismissedTLSWarning(true)} severity="warning">
-                    Mizu is detecting TLS traffic{addressesWithTLS.size ? ` (directed to ${Array.from(addressesWithTLS).join(", ")})` : ''}, this type of traffic will not be displayed.
+                <MuiAlert classes={{ filledWarning: 'customWarningStyle' }} elevation={6} variant="filled" onClose={() => setUserDismissedTLSWarning(true)} severity="warning">
+                    Mizu is detecting TLS traffic, this type of traffic will not be displayed.
+                    {addressesWithTLS.size ?  <ul className="httpsDomains"> {Array.from(addressesWithTLS, address => <li>{address}</li>)} </ul>: null}
```
lirazyehezkel, line 123: you can use:
{addressesWithTLS.size && <ul className="httpsDomains"> {Array.from(addressesWithTLS, address => <li>{address}</li>)} </ul>}

looks a bit better IMO

imceZZ: hi @lirazyehezkel  I tried your suggestion but this is what I get check 0 (zero) at the end

Screenshot 2021-08-17 at 18 36 55

if I make it addressesWithTLS.size > 0 && then everything is ok

imceZZ: this it happens when Set is empty if I leave addressesWithTLS.size && and the set is not empty then everything is OK

lirazyehezkel: Hmm are you sure? it seems like it should work:
image

if it doesn't work for some reason let's do: addressesWithTLS.size > 0 &&

imceZZ: @lirazyehezkel ok I changed to addressesWithTLS.size > 0 && but please before approving make a real test because if in new Set is present just one domain like this new Set('https://google.com') and that domain is not in array like this new Set(['https://google.com']) then in return function it will return char by char meaning h,t,t,p,s etc...

lirazyehezkel: where do you get from the domains list? let's make sure it will be always inside an array

PR #228, opened 2021-08-18
Fix: View command fail sporadically

IgorGov:


RoyUP9 (approved):
don't we want something without sleep in the future?

PR #235, opened 2021-08-19
fixed namespace restricted validation

RoyUP9:


cli/cmd/tapRunner.go
```diff
@@ -73,6 +73,14 @@ func RunMizuTap() {
 
 	targetNamespaces := getNamespaces(kubernetesProvider)
 
+	if config.Config.IsNsRestrictedMode() {
+		if len(targetNamespaces) != 1 || !mizu.Contains(targetNamespaces, config.Config.MizuResourcesNamespace) {
```
IgorGov, line 77: why contains? you reach this case only when len(targetNamespaces) == 1 

```suggestion
		if len(targetNamespaces) != 1 || targetNamespaces[0] != config.Config.MizuResourcesNamespace {
```

IgorGov (approved):

PR #243, opened 2021-08-20
Introducing API server provider

IgorGov:
- Added additional logs.
- Introducing API server provider (singleton) which incapsulates all request to mizu server
- Added connectivity test to API server in loop till it ready instead of sleep


cli/cmd/tapRunner.go
```diff
@@ -261,23 +253,26 @@ func updateMizuTappers(ctx context.Context, kubernetesProvider *kubernetes.Provi
 	return nil
 }
 
-func cleanUpMizu(kubernetesProvider *kubernetes.Provider) {
-	telemetry.ReportAPICalls(config.Config.Tap.GuiPort)
-	cleanUpMizuResources(kubernetesProvider)
-}
-
-func cleanUpMizuResources(kubernetesProvider *kubernetes.Provider) {
+func finishMizuException(kubernetesProvider *kubernetes.Provider) {
```
gadotroee, line 256: why finishMizuException? it is not necessarily means it throws exception from what I understand (t will just be called in the end of the function because we used 	defer finishMizuException(kubernetesProvider) (line 103)

cli/cmd/fetchRunner.go
```diff
@@ -1,96 +1,23 @@
...
-		err := extractAndWriteFile(f)
-		if err != nil {
-			return err
-		}
+		logger.Log.Errorf("Failed fetch data from API server %v", err)
+		return
 	}
 
-	return nil
+	_ = fsUtils.Unzip(zipReader, config.Config.Fetch.Directory)
```
gadotroee, line 22: maybe better to catch this and write to log?

cli/apiserver/provider.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+type apiServerProvider struct {
+	url     string
+	isReady bool
+}
+
+var Provider = apiServerProvider{}
+
+func (provider *apiServerProvider) Init(url string, retries int) error {
```
gadotroee, line 25: I would name it InitWithRetries.
because when I read it from outside (when we call this function) it is not clear what is the number after the url. 
it is not mandatory but maybe easier if we see the word retries in the function name

cli/apiserver/provider.go
```diff
@@ -0,0 +1,167 @@
...
+		} else {
+			logger.Log.Debugf("connection test to api server passed successfully")
+			break
+		}
+		retriesLeft -= 1
+		time.Sleep(time.Second)
+	}
+
+	if retriesLeft == 0 {
+		return fmt.Errorf("couldn't reach the api server after %v retries", retries)
```
gadotroee, line 42: do we want to return nil here? 
or maybe just change the isReady = false (I guess it is the default but maybe it is better to be specific and set the value to false?)

gadotroee (commented):
small comments, overall looks good IMO

cli/apiserver/provider.go
IgorGov: why return nil? it returns error because it failed to connect. the function sets isReady=True when it actually ready (all other cases doing nothing

gadotroee (approved):

PR #244, opened 2021-08-23
added tap acceptance tests, fixed duplicate namespace problem

RoyUP9:


gadotroee (approved):

acceptanceTests/Makefile
```diff
@@ -1,2 +1,2 @@
 test: ## Run acceptance tests.
-	@go test ./...
+	@go test ./... -timeout 0
```
IgorGov, line 2: lets make it hour

.github/workflows/pr_validation.yml
```diff
@@ -1,9 +1,19 @@
 name: PR validation
+
 on:
   pull_request:
     branches:
       - 'develop'
       - 'main'
+  push:
+    branches:
+      - 'develop'
```
IgorGov, line 10: lets check develop only tests (and not build)

PR #257, opened 2021-09-02
Fix the memory exhaustion by optimizing max. AMQP message size and `GOGC`

mertyildiran:

Sets maximum AMQP message size to 16MB

Adds one more fail-fast that checks max. header frame size of AMQP.

Sets `GOGC` to `12800` to trigger garbage collector less.

Does improvements in `startMemoryProfiler` method:
- Introduces `MEMORY_PROFILING_DUMP_PATH` environment variable.
- Introduces `MEMORY_PROFILING_TIME_INTERVAL` environment variable.

### Why setting `GOGC` to `12800`?

`GOGC`'s value is the criteria for Go's garbage collector triggering. It means the change in percentage of maximum allocated space compared to previous collection.

The dissectors constantly do allocations and quickly free them. This causes a lot of garbage collector triggering if it's set to `100`. A value like `12800` makes it less triggered. It actually reduces avg. memory usage because GC is less executed unnecessarily therefore jobs don't pile up.

Go runs its garbage collector at least every 2 ~seconds~ minutes no matter the value of `GOGC`. Also [see](https://github.com/up9inc/mizu/pull/257#issuecomment-913847182).

I've optimized it to `12800` by hand.

### Load Tests

I've run Sock Shop load tests for more than an hour with `--users 5 --spawn-rate 5 --headless` parameters. The peak memory usage was 312MB:

```
	User time (seconds): 906.36
	System time (seconds): 94.91
	Percent of CPU this job got: 25%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 1:05:19
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 312088
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 6173314
	Voluntary context switches: 25437585
	Involuntary context switches: 110789
	Swaps: 0
	File system inputs: 0
	File system outputs: 55216
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

Screenshot from 2021-09-03 03-52-28

### DIFF
--- a/tap/extensions/amqp/main.go
+++ b/tap/extensions/amqp/main.go
@@ -78,16 +78,11 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 	var lastMethodFrameMessage Message
 
 	for {
-		// fmt.Printf("lastMethodFrameMessage: %v\n", lastMethodFrameMessage)
 		frame, err := r.ReadFrame()
 		if err == io.EOF {
 			// We must read until we see an EOF... very important!
 			return errors.New("AMQP EOF")
-		} else if err != nil {
-			// 	return err
 		}
-
-		// fmt.Printf("frame: %+v\n", frame)
 
 		switch f := frame.(type) {
 		case *HeartbeatFrame:
--- a/tap/extensions/amqp/main.go
+++ b/tap/extensions/amqp/main.go
@@ -78,18 +78,16 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 	var lastMethodFrameMessage Message
 
 	for {
+		// fmt.Printf("lastMethodFrameMessage: %v\n", lastMethodFrameMessage)
 		frame, err := r.ReadFrame()
 		if err == io.EOF {
 			// We must read until we see an EOF... very important!
 			return errors.New("AMQP EOF")
 		} else if err != nil {
-			switch err.(type) {
-			case *Error:
-				if err.(*Error).Code == MaxHeaderFrameSizeError {
-					return err
-				}
-			}
+			// 	return err
 		}
+
+		// fmt.Printf("frame: %+v\n", frame)
 
 		switch f := frame.(type) {
 		case *HeartbeatFrame:
--- a/tap/extensions/amqp/main.go
+++ b/tap/extensions/amqp/main.go
@@ -83,8 +83,12 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 			// We must read until we see an EOF... very important!
 			return errors.New("AMQP EOF")
 		} else if err != nil {
-			// TODO: Causes ignoring some methods. Return only in case of a certain error. But what?
-			return err
+			switch err.(type) {
+			case *Error:
+				if err.(*Error).Code == MaxHeaderFrameSizeError {
+					return err
+				}
+			}
 		}
 
 		switch f := frame.(type) {
@@ -101,6 +105,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 			case *BasicDeliver:
 				eventBasicDeliver.Properties = header.Properties
 			default:
+				frame = nil
 			}
 
 		case *BodyFrame:
@@ -115,6 +120,8 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 				eventBasicDeliver.Body = f.Body
 				emitAMQP(*eventBasicDeliver, amqpRequest, basicMethodMap[60], connectionInfo, emitter)
 			default:
+				body = nil
+				frame = nil
 			}
 
 		case *MethodFrame:
@@ -200,6 +207,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 				emitAMQP(*eventConnectionClose, amqpRequest, connectionMethodMap[50], connectionInfo, emitter)
 
 			default:
+				frame = nil
 
 			}
 
--- a/tap/extensions/amqp/spec091.go
+++ b/tap/extensions/amqp/spec091.go
@@ -10,7 +10,6 @@ package main
 
 import (
 	"encoding/binary"
-	"fmt"
 	"io"
 )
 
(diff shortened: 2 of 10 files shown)


tap/extensions/amqp/main.go
```diff
@@ -78,15 +78,17 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 	var lastMethodFrameMessage Message
 
 	for {
+		// fmt.Printf("lastMethodFrameMessage: %v\n", lastMethodFrameMessage)
```
nimrod-up9, line 81: Either remove comment or print in DEBUG or TRACE level.
Same comment for line 90.

tap/extensions/amqp/main.go
```diff
@@ -78,15 +78,17 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
...
 		frame, err := r.ReadFrame()
 		if err == io.EOF {
 			// We must read until we see an EOF... very important!
 			return errors.New("AMQP EOF")
 		} else if err != nil {
-			// TODO: Causes ignoring some methods. Return only in case of a certain error. But what?
-			return err
+			// 	return err
```
nimrod-up9, line 87: Was this line commented out on purpose?
If it was, then remove the `else` clause.

tap/extensions/amqp/spec091.go
```diff
@@ -2854,7 +2856,8 @@ func (r *AmqpReader) parseMethodFrame(channel uint16, size uint32) (f frame, err
 			mf.Method = method
 
 		default:
-			return nil, fmt.Errorf("Bad method frame, unknown method %d for class %d", mf.MethodId, mf.ClassId)
+			// log.Printf("Bad method frame, unknown method %d for class %d", mf.MethodId, mf.ClassId)
```
nimrod-up9, line 2859: Either remove comment or print in DEBUG or TRACE level.
Same comment for the rest of the changes in this file.

nimrod-up9 (commented):
`12800` seems like a very high value.
If the heap is at 100MB, then the garbage collector will be triggered when the heap grows by 100MB * 12800% = 13GB.
If I understand correctly, for all practical purposes, this is the same as turning off the GC (which we don't want to do).

Can you please direct me to where it says that the GC runs every 2 seconds regardless of GOGC? I tried searching for it in blogs, the docs and in the source code but I couldn't find it (maybe I'm searching for the wrong thing).

A value of `1000` could be reasonable. Can you expand on how you hand tuned GOGC?

Is there another solution? Is it possible to make AMQP reuse its buffers instead of releasing them and reallocating?

mertyildiran:
> Can you please direct me to where it says that the GC runs every 2 seconds regardless of GOGC? I tried searching for it in blogs, the docs and in the source code but I couldn't find it (maybe I'm searching for the wrong thing).

My bad that it's actually 2 minutes, not 2 seconds. It is here in the source code of Go, defined in nanoseconds:

```go
// forcegcperiod is the maximum time in nanoseconds between garbage
// collections. If we go this long without a garbage collection, one
// is forced to run.
//
// This is a variable for testing purposes. It normally doesn't change.
var forcegcperiod int64 = 2 * 60 * 1e9
```
https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/proc.go#L5230-L5235

and it's used in here:

```go
	case gcTriggerTime:
		if gcController.gcPercent < 0 {
			return false
		}
		lastgc := int64(atomic.Load64(&memstats.last_gc_nanotime))
		return lastgc != 0 && t.now-lastgc > forcegcperiod
```

https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/mgc.go#L552

But there is another `case` statement there:

```go
	case gcTriggerCycle:
		// t.n > work.cycles, but accounting for wraparound.
		return int32(t.n-work.cycles) > 0
```

https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/mgc.go#L553-L555

Which checks if the number of already ran GC trigger cycles are bigger than the work cycles and the work cycles' value is updated on this line:

```asm
TEXT runtime·procyield(SB),NOSPLIT,$0-0
	MOVL	cycles+0(FP), AX
again:
	PAUSE
	SUBL	$1, AX
	JNZ	again
	RET
```

https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/asm_amd64.s#L605

whenever `procyield` method is called and it's called from many different places.

mertyildiran:
> `12800` seems like a very high value.
> If the heap is at 100MB, then the garbage collector will be triggered when the heap grows by 100MB * 12800% = 13GB.
> If I understand correctly, for all practical purposes, this is the same as turning off the GC (which we don't want to do).

Well not really. Setting `GOGC=12800` does not mean we disabled garbage collection. There are other mechanisms that are triggering garbage collection in Go. [See](https://github.com/up9inc/mizu/pull/257#issuecomment-913847182).

> A value of `1000` could be reasonable. Can you expand on how you hand tuned GOGC?

Maybe. I've hand tuned the value by trying the powers of two times 100 and looking at the memory profiling results.

> Is there another solution? Is it possible to make AMQP reuse its buffers instead of releasing them and reallocating?

I don't see another solution as of this moment. The issue is not AMQP specific. All protocol dissectors quickly allocate and free memory in rapid succession, concurrently. If we were implementing this in a language like C, the freeing would be guaranteed with the call to `free(x)` but since Go is garbage collected the garbage collection's itself becomes a problem that needs to be fixed.

A garbage collected language is not really suitable for implementing real-time systems. See these great answers:
- https://softwareengineering.stackexchange.com/a/285336/236831
- https://stackoverflow.com/a/3560023/2104879

A solution would be doing memory pooling using `sync.Pool` and not relying on Go's GC at all. But trying to convert every allocation to pooling would be a pain that no one can imagine.

mertyildiran:
@nimrod-up9 this is the alternative solution:

https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap-26c2462549a2/

```go
func main() {

	// Create a large heap allocation of 10 GiB
	ballast := make([]byte, 10<<30)

	// Application execution continues
	// ...
}
```

allocate 10 GiB of memory and never worry about GC triggers.

tap/extensions/amqp/main.go
mertyildiran: @nimrod-up9 removed with https://github.com/up9inc/mizu/commit/1a6d32e6614e4587883b27dda599f011054b8605

tap/extensions/amqp/main.go
mertyildiran: @nimrod-up9 yes it was intentional.

The compiler probably ignores empty `else if` outcomes but removed it with https://github.com/up9inc/mizu/commit/1a6d32e6614e4587883b27dda599f011054b8605 anyway.

tap/extensions/amqp/spec091.go
mertyildiran: @nimrod-up9 removed them with https://github.com/up9inc/mizu/commit/1a6d32e6614e4587883b27dda599f011054b8605

nimrod-up9:
> > Can you please direct me to where it says that the GC runs every 2 seconds regardless of GOGC? I tried searching for it in blogs, the docs and in the source code but I couldn't find it (maybe I'm searching for the wrong thing).
> 
> @nimrod-up9 My bad that it's actually 2 minutes, not 2 seconds. It is here in the source code of Go, defined in nanoseconds:
> 
> ```go
> // forcegcperiod is the maximum time in nanoseconds between garbage
> // collections. If we go this long without a garbage collection, one
> // is forced to run.
> //
> // This is a variable for testing purposes. It normally doesn't change.
> var forcegcperiod int64 = 2 * 60 * 1e9
> ```
> 
> https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/proc.go#L5230-L5235
> 
> and it's used in here:
> 
> ```go
> 	case gcTriggerTime:
> 		if gcController.gcPercent < 0 {
> 			return false
> 		}
> 		lastgc := int64(atomic.Load64(&memstats.last_gc_nanotime))
> 		return lastgc != 0 && t.now-lastgc > forcegcperiod
> ```
> 
> https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/mgc.go#L552
> 
> But there is another `case` statement there:
> 
> ```go
> 	case gcTriggerCycle:
> 		// t.n > work.cycles, but accounting for wraparound.
> 		return int32(t.n-work.cycles) > 0
> ```
> 
> https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/mgc.go#L553-L555
> 
> Which checks if the number of already ran GC trigger cycles are bigger than the work cycles and the work cycles' value is updated on this line:
> 
> ```assembly
> TEXT runtime·procyield(SB),NOSPLIT,$0-0
> 	MOVL	cycles+0(FP), AX
> again:
> 	PAUSE
> 	SUBL	$1, AX
> 	JNZ	again
> 	RET
> ```
> 
> https://github.com/golang/go/blob/a1938435d6361dcbc93a15ce0ace28748a45b85d/src/runtime/asm_amd64.s#L605
> 
> whenever `procyield` method is called and it's called from many different places.

Thanks for the explanation.

nimrod-up9 (approved):
Let's see how this behaves in develop.

PR #258, opened 2021-09-03
Fix body size, receive (elapsed time) and timestamps

mertyildiran:

Fixes the body size. It was in the `har.Entry.Response.BodySize` before https://pkg.go.dev/github.com/google/martian/har#Response https://github.com/up9inc/mizu/blob/17fa163ee310d94c3c6efea0509b014312290842/agent/pkg/models/models.go#L206

fetched through
https://github.com/up9inc/mizu/blob/7f2021c312ddc5249ba2f02dc7c6332acb12bb21/ui/src/components/HarEntryDetailed.tsx#L33

Fixes the receive (elapsed time). It's https://pkg.go.dev/github.com/google/martian/har#Timings Calculated just as before https://github.com/up9inc/mizu/blob/17fa163ee310d94c3c6efea0509b014312290842/tap/har_writer.go#L101

Fixes the timestamps. Now they come from https://github.com/up9inc/mizu/blob/71d5b10f56ca0e6738330f341ffb9c7a3cbc071d/tap/tcp_reader.go#L72 just as before.

Didn't bring back the status text part. Since it's now available in the details section and it's HTTP specific:

Screenshot from 2021-09-03 04-05-31

### DIFF
--- a/tap/api/api.go
+++ b/tap/api/api.go
@@ -68,10 +68,14 @@ type OutputChannelItem struct {
 	Pair           *RequestResponsePair
 }
 
+type SuperTimer struct {
+	CaptureTime time.Time
+}
+
 type Dissector interface {
 	Register(*Extension)
 	Ping()
-	Dissect(b *bufio.Reader, isClient bool, tcpID *TcpID, counterPair *CounterPair, emitter Emitter) error
+	Dissect(b *bufio.Reader, isClient bool, tcpID *TcpID, counterPair *CounterPair, superTimer *SuperTimer, emitter Emitter) error
 	Analyze(item *OutputChannelItem, entryId string, resolvedSource string, resolvedDestination string) *MizuEntry
 	Summarize(entry *MizuEntry) *BaseEntryDetails
 	Represent(entry *MizuEntry) (protocol Protocol, object []byte, bodySize int64, err error)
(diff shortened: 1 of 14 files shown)


tap/api/api.go
```diff
@@ -103,6 +107,7 @@ type MizuEntry struct {
 	RequestSenderIp     string `json:"requestSenderIp" gorm:"column:requestSenderIp"`
 	Service             string `json:"service" gorm:"column:service"`
 	Timestamp           int64  `json:"timestamp" gorm:"column:timestamp"`
+	ElapsedTime         int64  `json:"elapsed_time" gorm:"column:elapsedTime"`
```
gadotroee, line 110: I see others are in camelCase (in the json annotation), is there any reason it is different?)

tap/api/api.go
```diff
@@ -117,6 +122,7 @@ type MizuEntry struct {
 type MizuEntryWrapper struct {
 	Protocol       Protocol  `json:"protocol"`
 	Representation string    `json:"representation"`
+	BodySize       int64     `json:"body_size"`
```
gadotroee, line 125: Same camelCase as elapsed_time comment

gadotroee (commented):
Looks good IMO, left 2 small comments

tap/api/api.go
mertyildiran: @gadotroee I'm sorry, I'm used to snake_case when I see a JSON field. Fixed it with https://github.com/up9inc/mizu/commit/516a3e59b062f864ad54f5d7ee3a4f8e4a8a2259

tap/api/api.go
mertyildiran: @gadotroee I'm sorry, I'm used to snake_case when I see a JSON field. Fixed it with https://github.com/up9inc/mizu/commit/516a3e59b062f864ad54f5d7ee3a4f8e4a8a2259

gadotroee (approved):
It is ok, maybe snake case is better is this is what you used to see, I just try to check that we have one type of styling so it will not confuse someone that some of the fields are camelCase and some are snake_case. 
Anyway I approved it now and I'm happy that we are making develop stable as soon as we can.

PR #260, opened 2021-09-03
Stop the hanging Goroutines by dropping the old, unidentified TCP streams

mertyildiran:

Closes the hanging TCP message channels after a dynamically aligned timeout (base `500000` milliseconds).

https://github.com/up9inc/mizu/blob/aa24ee37791ca00741dbfb0f78e9334d6a3b344f/tap/passive_tapper.go#L98

Since all ports are listened now, in most cases,

https://github.com/up9inc/mizu/blob/f13152d619004de98d15b191e4ca3e1a3e37b97b/tap/tcp_stream.go#L166

is not called. Therefore the channels do not close:

https://github.com/up9inc/mizu/blob/f13152d619004de98d15b191e4ca3e1a3e37b97b/tap/tcp_reader.go#L52

and in some cases Goroutines pile up. Because `Read` does not return (falls into infinite loop):

https://github.com/up9inc/mizu/blob/f13152d619004de98d15b191e4ca3e1a3e37b97b/tap/tcp_reader.go#L69

This PR closes the TCP message channels based on the formula below:

https://github.com/up9inc/mizu/blob/aa24ee37791ca00741dbfb0f78e9334d6a3b344f/tap/passive_tapper.go#L214-L225

_**500000** x **CPU core count** / (**Number of currently running Goroutines** - **Base Goroutine count**) (ms)_

Moved from `Packets` to `NextPacket`:

https://github.com/up9inc/mizu/blob/f13152d619004de98d15b191e4ca3e1a3e37b97b/tap/passive_tapper.go#L388

The documentation says:

> This method is the most flexible, and exposes errors that may be encountered by the underlying PacketDataSource. It's also the fastest in a tight loop, since it doesn't have the overhead of a channel read/write. However, it requires the user to handle errors, most importantly the io.EOF error in cases where packets are being read from a file. - [source](https://pkg.go.dev/github.com/google/gopacket#hdr-Reading_With_NextPacket_Function)

### Test Results

It stabilizes arround these values:

```
Total dropped: 1780 Goroutine metric: 0.003213 Total Goroutines: 3748 Timeout (ms): 1606
```

no matter what the load is or the resources are.

On my machine;

For the `locust --host http://192.168.49.2:30001 -f load_dev.py --users 100 --spawn-rate 100 --headless` the drop rate was 0.1% - 0.5%. 

```
2021/09/05 10:18:34 passive_tapper.go:386: app stats - {"processedBytes":258225879,"packetsCount":626898,"tcpPacketsCount":574718,"reassembledTcpPayloadsCount":176064,"tlsConnectionsCount":0,"matchedPairs":0,"droppedTcpStreams":9797}
```

For `locust --host http://192.168.49.2:30001 -f load_dev.py --users 50 --spawn-rate 50 --headless` it was ignorable.

Better machines would have less drop rate.

### Conclusion

`ReassemblyComplete` is not guaranteed to be called. Especially the reassembly does not finish for TLS connections in most cases. I tried putting `return 0, io.EOF` to here:

https://github.com/up9inc/mizu/blob/39ea02250ef61162e57a9e2bb247725a99b1a732/tap/tcp_reader.go#L80

and putting `layers.LayerTypeTLS` check in here:

https://github.com/up9inc/mizu/blob/39ea02250ef61162e57a9e2bb247725a99b1a732/tap/passive_tapper.go#L426-L427

But they didn't work. Therefore I came up with this solution. It might not be the best solution but it prevents the overloads at least.

_There are [issues](https://github.com/google/gopacket/issues?q=is%3Aissue+TLS) like [this one](https://github.com/google/gopacket/issues/905) in `gopacket` related to TLS that worth noting._

_Even though we rule out the TLS connections. There could still be a need for such a dropping mechanism. Since the issue is **not TLS specific**._

---

**Addition:** [See](https://github.com/up9inc/mizu/pull/260#issuecomment-913079684).

### DIFF
--- a/tap/extensions/amqp/main.go
+++ b/tap/extensions/amqp/main.go
@@ -41,7 +41,7 @@ func (d dissecting) Ping() {
 
 const amqpRequest string = "amqp_request"
 
-func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, counterPair *api.CounterPair, superTimer *api.SuperTimer, emitter api.Emitter) error {
+func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, counterPair *api.CounterPair, superTimer *api.SuperTimer, superIdentifier *api.SuperIdentifier, emitter api.Emitter) error {
 	r := AmqpReader{b}
 
 	var remaining int
@@ -78,6 +78,10 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 	var lastMethodFrameMessage Message
 
 	for {
+		if superIdentifier.Protocol != nil && superIdentifier.Protocol != &protocol {
+			return errors.New("Identified by another protocol")
+		}
+
 		frame, err := r.ReadFrame()
 		if err == io.EOF {
 			// We must read until we see an EOF... very important!
@@ -110,9 +114,11 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 			switch lastMethodFrameMessage.(type) {
 			case *BasicPublish:
 				eventBasicPublish.Body = f.Body
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventBasicPublish, amqpRequest, basicMethodMap[40], connectionInfo, superTimer.CaptureTime, emitter)
 			case *BasicDeliver:
 				eventBasicDeliver.Body = f.Body
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventBasicDeliver, amqpRequest, basicMethodMap[60], connectionInfo, superTimer.CaptureTime, emitter)
 			default:
 			}
@@ -134,6 +140,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 					NoWait:     m.NoWait,
 					Arguments:  m.Arguments,
 				}
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventQueueBind, amqpRequest, queueMethodMap[20], connectionInfo, superTimer.CaptureTime, emitter)
 
 			case *BasicConsume:
@@ -146,6 +153,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 					NoWait:      m.NoWait,
 					Arguments:   m.Arguments,
 				}
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventBasicConsume, amqpRequest, basicMethodMap[20], connectionInfo, superTimer.CaptureTime, emitter)
 
 			case *BasicDeliver:
@@ -165,6 +173,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 					NoWait:     m.NoWait,
 					Arguments:  m.Arguments,
 				}
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventQueueDeclare, amqpRequest, queueMethodMap[10], connectionInfo, superTimer.CaptureTime, emitter)
 
 			case *ExchangeDeclare:
@@ -178,6 +187,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 					NoWait:     m.NoWait,
 					Arguments:  m.Arguments,
 				}
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventExchangeDeclare, amqpRequest, exchangeMethodMap[10], connectionInfo, superTimer.CaptureTime, emitter)
 
 			case *ConnectionStart:
@@ -188,6 +198,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 					Mechanisms:       m.Mechanisms,
 					Locales:          m.Locales,
 				}
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventConnectionStart, amqpRequest, connectionMethodMap[10], connectionInfo, superTimer.CaptureTime, emitter)
 
 			case *ConnectionClose:
@@ -197,6 +208,7 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 					ClassId:   m.ClassId,
 					MethodId:  m.MethodId,
 				}
+				superIdentifier.Protocol = &protocol
 				emitAMQP(*eventConnectionClose, amqpRequest, connectionMethodMap[50], connectionInfo, superTimer.CaptureTime, emitter)
 
 			default:
--- a/tap/extensions/amqp/main.go
+++ b/tap/extensions/amqp/main.go
@@ -32,7 +32,7 @@ func init() {
 type dissecting string
 
 func (d dissecting) Register(extension *api.Extension) {
-	extension.Protocol = protocol
+	extension.Protocol = &protocol
 }
 
 func (d dissecting) Ping() {
--- a/tap/passive_tapper.go
+++ b/tap/passive_tapper.go
@@ -226,7 +226,6 @@ func closeTimedoutTcpStreamChannels() {
 			if !stream.isClosed && stream.superIdentifier.Protocol == nil && time.Now().After(streamWrapper.createdAt.Add(time.Duration(dynamicStreamChannelTimeoutNs))) {
 				stream.Close()
 				statsTracker.incDroppedTcpStreams()
-				fmt.Printf("Dropped an unidentified TCP stream because of load. Total dropped: %d Goroutine metric: %f Total Goroutines: %d Timeout (ms): %d\n", statsTracker.appStats.DroppedTcpStreams, metric, n, dynamicStreamChannelTimeoutMs)
 				rlog.Debugf("Dropped an unidentified TCP stream because of load. Total dropped: %d Goroutine metric: %f Total Goroutines: %d Timeout (ms): %d\n", statsTracker.appStats.DroppedTcpStreams, metric, n, dynamicStreamChannelTimeoutMs)
 			}
 			return true
--- a/tap/passive_tapper.go
+++ b/tap/passive_tapper.go
@@ -223,24 +223,28 @@ func closeTimedoutTcpStreamChannels() {
 			metric := float64(cpuCount) / float64(diff)
 			dynamicStreamChannelTimeoutMs := int(float64(baseStreamChannelTimeoutMs) * metric)
 			dynamicStreamChannelTimeoutNs := dynamicStreamChannelTimeoutMs * 1000000
-			if !stream.isClosed && stream.superIdentifier.Protocol == nil && time.Now().After(streamWrapper.createdAt.Add(time.Duration(dynamicStreamChannelTimeoutNs))) {
-				stream.Close()
-				statsTracker.incDroppedTcpStreams()
-				rlog.Debugf("Dropped an unidentified TCP stream because of load. Total dropped: %d Goroutine metric: %f Total Goroutines: %d Timeout (ms): %d\n", statsTracker.appStats.DroppedTcpStreams, metric, n, dynamicStreamChannelTimeoutMs)
-			} else if stream.superIdentifier.Protocol != nil && !stream.superIdentifier.IsClosedOthers {
-				for i := range stream.clients {
-					reader := &stream.clients[i]
-					if reader.extension.Protocol != stream.superIdentifier.Protocol {
-						reader.Close()
-					}
+			if stream.superIdentifier.Protocol == nil {
+				if !stream.isClosed && time.Now().After(streamWrapper.createdAt.Add(time.Duration(dynamicStreamChannelTimeoutNs))) {
+					stream.Close()
+					statsTracker.incDroppedTcpStreams()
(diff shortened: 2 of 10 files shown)


tap/passive_tapper.go
```diff
@@ -354,7 +384,14 @@ func startPassiveTapper(outputItems chan *api.OutputChannelItem) {
 		startMemoryProfiler()
 	}
 
-	for packet := range source.Packets() {
+	for {
+		packet, err := source.NextPacket()
+		if err == io.EOF {
+			break
```
gadotroee, line 390: If I understand correctly - this error should happen only when reading the packets from file and we reach the end?
so if we are reading from interfact (like the default "any") will we reach this "break"?

mertyildiran: @gadotroee tested with all the supported protocols and Sock Shop load test, didn't experience an `EOF`. I think `EOF` is not possible for Linux network interfaces. Do we support reading traffic from `.pcap` files? In any case, it can be turned into `continue` as a safe bet.

I've pushed more commits and updated the PR description. Please also review them.

tap/stats_tracker.go
```diff
@@ -23,6 +24,7 @@ type StatsTracker struct {
 	reassembledTcpPayloadsCountMutex sync.Mutex
```
mertyildiran, line 24: @gadotroee why are there multiple `sync.Mutex`es here instead of just one in `AppStats` struct?

mertyildiran:
I've added one more criteria with my last commit https://github.com/up9inc/mizu/commit/cba52468fb9aee60aabf1486162a725dad03d27d that checks if the TCP stream is identified by a protocol. Because without that "the dropping solution" was invalid for long lasting TCP connections like AMQP has.

If a TCP stream is not identified in the time frame that's dynamically determined by the formula then it's dropped:

https://github.com/up9inc/mizu/blob/cba52468fb9aee60aabf1486162a725dad03d27d/tap/passive_tapper.go#L220

tap/stats_tracker.go
mertyildiran: For increments/decrements using https://pkg.go.dev/sync/atomic might yield better performance.

mertyildiran:
`runtime.NumCPU` factor is actually open to discussion. We might want to take CPU clock speed into account. or maybe instead of CPU core count and clock speed, Kubernetes resources info of the pod can be the factor. I believe the drop rate should be tested in a real cloud environment with different CPU core counts. Hence we can fine tune the formula.

tap/stats_tracker.go
gadotroee: yes we already taught about replacing it with atomic, the code will be much better and as you said might bring better performance. you can see that one of the counters that was added lately used this atomic (I think) - but if you want to add it - your are more than welcome to do it

mertyildiran: @gadotroee let's do that in a later PR. Because it's not the concern of this PR. I just wanted to note that.

mertyildiran:
I can confirm that piling up Goroutines is an issue in https://github.com/up9inc/mizu/commit/17fa163ee310d94c3c6efea0509b014312290842 before the merge of https://github.com/up9inc/mizu/pull/224

Tap into Sock Shop `./cli/bin/mizu__ tap -n sock-shop` and run `locust --host http://192.168.49.2:30001 -f load_dev.py --users 100 --spawn-rate 100 --headless`. You will see that number of Goroutines will reach to **10000+** and continue to grow.

Same issue is also valid for https://github.com/up9inc/mizu/commit/d4436d9f152ae2db4cda4ffd479294947063679c It just triples the number of Goroutines so it reaches to **30000+** arround the same time.

The number of piling up Goroutines would never reduce or stabilize in both of the commits if the Kubernetes cluster is experiencing a high traffic load constantly.

The memory cost of 10000 Goroutines is 47MB in a 64-bit machine. More Goroutines lead to slower garbage-collection which makes the memory situation exponentially worse. Infinitely growing number of Goroutines eventually makes it run out of memory. (got a crash in https://github.com/up9inc/mizu/commit/17fa163ee310d94c3c6efea0509b014312290842 too)

#### This PR;

Introduces a solution of **dropping the old, unidentified TCP streams** in the favor of **stabilizing the number of Goroutines** and **being port independent**.

After this PR, the number of Goroutines stabilizes arround 3000 - 4000 in the same exact scenario.

```
2021/09/05 10:18:34 passive_tapper.go:386: app stats - {"processedBytes":258225879,"packetsCount":626898,"tcpPacketsCount":574718,"reassembledTcpPayloadsCount":176064,"tlsConnectionsCount":0,"matchedPairs":0,"droppedTcpStreams":9797}
```

*`matchedPairs` value will be fixed with another PR. It's not recorded right now.*

mertyildiran:
We might also want to take *TCP Keep-Alive* and *HTTP Keep-Alive* into account.

tap/extensions/amqp/main.go
```diff
@@ -78,6 +78,10 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 	var lastMethodFrameMessage Message
 
 	for {
+		if superIdentifier.Protocol != nil && superIdentifier.Protocol != &protocol {
+			return errors.New("Identified by another protocol")
+		}
```
nimrod-up9, line 83: Can checking and setting the protocol be done in one place for all protocols? Before calling to Dissect maybe?

tap/passive_tapper.go
```diff
@@ -198,8 +201,41 @@ func startMemoryProfiler() {
 	}()
 }
 
+func closeTimedoutTcpStreamChannels() {
+	cpuCount := runtime.NumCPU()
+	// +1 Stats tracker Goroutine
+	// +1 Main packet capture loop
+	baseGoroutineCount := runtime.NumGoroutine() + 2
+	for {
+		runtime.Gosched()
```
nimrod-up9, line 210: Why the call to `Gosched`? If this is because this loop doesn't block on anything, then as I understand it this isn't needed in new versions of go (I think 1.16 is included).
Anyway, it is healthier to add a sleep period. This function routine does need to run all the time.

tap/passive_tapper.go
```diff
@@ -94,6 +95,8 @@ var ownIps []string             // global
 var hostMode bool               // global
 var extensions []*api.Extension // global
 
+const baseStreamChannelTimeoutMs int = 5000 * 100
```
nimrod-up9, line 98: Use `time.Millisecond` or `time.Nanosecond`.

tap/passive_tapper.go
```diff
@@ -198,8 +201,41 @@ func startMemoryProfiler() {
...
+			diff := n - baseGoroutineCount
+			if diff < 1 {
+				diff = 1
+			}
+			// `metric` tries to stabilize arround 1.0.
+			// `metric` means what percentage of the load can be actually handled by the given resources.
+			metric := float64(cpuCount) / float64(diff)
+			dynamicStreamChannelTimeoutMs := int(float64(baseStreamChannelTimeoutMs) * metric)
+			dynamicStreamChannelTimeoutNs := dynamicStreamChannelTimeoutMs * 1000000
+			if !stream.isClosed && stream.superIdentifier.Protocol == nil && time.Now().After(streamWrapper.createdAt.Add(time.Duration(dynamicStreamChannelTimeoutNs))) {
```
nimrod-up9, line 226: What does it mean that `stream.superIdentifier.Protocol == nil`?
If we have `stream.superIdentifier.Protocol` non-nil for an inactive stream, will it ever close?

nimrod-up9: I think the algorithm can be simplified by setting the maximum allowed nil-protocol-time to a const (the argument to `After`).
Do your tests lead to a different conclusion?

tap/tcp_reader.go
```diff
@@ -59,6 +60,7 @@ type tcpReader struct {
 	extension          *api.Extension
 	emitter            api.Emitter
 	counterPair        *api.CounterPair
+	sync.Mutex
```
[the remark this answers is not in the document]
mertyildiran: @nimrod-up9 there are two fields `isClosed` and `msgQueue` together needs to be guarded in concurrent calls to the following methods:

https://github.com/up9inc/mizu/blob/6a6575a72793f102311116edd237c797515420f2/tap/tcp_stream.go#L152-L170

https://github.com/up9inc/mizu/blob/6a6575a72793f102311116edd237c797515420f2/tap/tcp_stream.go#L184-L216

Since `tcpReader` struct is not exported there is no difference between embedding or not in terms of accessibility. There is also no performance difference as far as I know.

tap/extensions/amqp/main.go
mertyildiran: @nimrod-up9 no, because the dissectors run in parallel. Once a dissector sets `superIdentifier.Protocol` hence it becomes non-nil, the other protocols stop reading with these checks.

The read loops start to happen with the `Dissect` call so the protocol identification happens inside those concurrent read loops. This check is a way to provide a communication between the protocol dissectors without blocking them.

tap/passive_tapper.go
mertyildiran: @nimrod-up9 if `stream.superIdentifier.Protocol == nil` is `true` it means that the TCP stream is not yet identified by any protocols. If it's `false` it means the TCP stream is identified by a protocol.

If `stream.superIdentifier.Protocol` is non-nil then yes the TCP stream is not dropped. The other protocols break their read loop as it's explained in here https://github.com/up9inc/mizu/pull/260#discussion_r703020684

`stream.superIdentifier.Protocol == nil` criteria is there because the AMQP streams are long lasting and shouldn't be dropped. Please see https://github.com/up9inc/mizu/pull/260#issuecomment-913079684

mertyildiran: @nimrod-up9 the algorithm cannot be simplified because it scales vertically and the timeout is rearranged based on the load. 

The formula tries to answer this question: "Do we have enough resources to handle that amount Goroutines?" If the answer to this question is no then it drops the **old, unidentified TCP streams**.

There is a room for the improvement in the formula though. Please see;
- https://github.com/up9inc/mizu/pull/260#issuecomment-913082888
- https://github.com/up9inc/mizu/pull/260#issuecomment-913126666

tap/passive_tapper.go
mertyildiran: @nimrod-up9 if we make it `time.Duration` then it has to be converted back to `int/float` in here:

https://github.com/up9inc/mizu/blob/6a6575a72793f102311116edd237c797515420f2/tap/passive_tapper.go#L224

While the value is converted to `time.Duration` eventually in here:

https://github.com/up9inc/mizu/blob/6a6575a72793f102311116edd237c797515420f2/tap/passive_tapper.go#L226

So converting it to `time.Duration` will probably increase the amount CPU instructions needed to run.

tap/passive_tapper.go
mertyildiran: @nimrod-up9 yes, since it's like a non-critical housekeeping method basically the loop does not need to run so aggressively. I also thought to add a `time.Sleep` call but couldn't decide the duration. But now I see that `time.Sleep` provides better resource efficiency. So I switched to `time.Sleep` with https://github.com/up9inc/mizu/commit/77eaed598975ef6f0a8bfb550e9b59a61f17a5e8

tap/extensions/amqp/main.go
mertyildiran: @nimrod-up9 Actually, I've realized that there can and should be a single place that closes the channels of other protocols in case of an identification for better cancellation. In addition to those checks. So I've added that with https://github.com/up9inc/mizu/commit/fb4f4c9a4c7224636ceb5d5b6903603a8bd7d0d3

tap/passive_tapper.go
```diff
@@ -198,8 +201,59 @@ func startMemoryProfiler() {
...
+			n := runtime.NumGoroutine()
+			diff := n - baseGoroutineCount
+			if diff < 1 {
+				diff = 1
+			}
+			// `metric` tries to stabilize arround 1.0.
+			// `metric` means what percentage of the load can be actually handled by the given resources.
+			metric := float64(cpuCount) / float64(diff)
+			dynamicStreamChannelTimeoutMs := int(float64(baseStreamChannelTimeoutMs) * metric)
+			dynamicStreamChannelTimeoutNs := dynamicStreamChannelTimeoutMs * 1000000
```
nimrod-up9, line 225: `dynamicStreamChannelTimeoutNs` is difficult to reason about.
A constant timeout makes more sense to me. We can say that if a protocol was not detected in under X ms and/or after Y packets, then the stream should be dropped.
I expect each routine to be blocked most of the time, waiting for traffic, so a time-based criterion is more suitable than guessing the effective CPU cycles dedicated to dissector.

tap/passive_tapper.go
```diff
@@ -198,8 +201,59 @@ func startMemoryProfiler() {
 	}()
 }
 
+func closeTimedoutTcpStreamChannels() {
+	cpuCount := runtime.NumCPU()
+	// +1 Stats tracker Goroutine
+	// +1 Main packet capture loop
+	baseGoroutineCount := runtime.NumGoroutine() + 2
+	for {
+		time.Sleep(1 * time.Millisecond)
```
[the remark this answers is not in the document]
mertyildiran: @nimrod-up9 changed it to 10ms with https://github.com/up9inc/mizu/commit/e92410688c651d34fc53f162f31aa5277e4cdb72

tap/passive_tapper.go
mertyildiran: @nimrod-up9 replaced the formula with hard timeout 5000 milliseconds and 4000 maximum number of Goroutines https://github.com/up9inc/mizu/commit/e92410688c651d34fc53f162f31aa5277e4cdb72

nimrod-up9 (approved):

PR #266, opened 2021-09-07
Temporarily fix the filtering into its original state

mertyildiran:

### DIFF
--- a/ui/src/components/EntriesList.tsx
+++ b/ui/src/components/EntriesList.tsx
@@ -54,8 +54,8 @@ export const EntriesList: React.FC<EntriesListProps> = ({entries, setEntries, fo
     const filterEntries = useCallback((entry) => {
         if(methodsFilter.length > 0 && !methodsFilter.includes(entry.method.toLowerCase())) return;
         if(pathFilter && entry.path?.toLowerCase()?.indexOf(pathFilter) === -1) return;
-        if(statusFilter.includes(StatusType.SUCCESS) && entry.statusCode >= 400) return;
-        if(statusFilter.includes(StatusType.ERROR) && entry.statusCode < 400) return;
+        if(statusFilter.includes(StatusType.SUCCESS) && entry.status_code >= 400) return;
+        if(statusFilter.includes(StatusType.ERROR) && entry.status_code < 400) return;
         return entry;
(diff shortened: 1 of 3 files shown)


ui/src/components/EntriesList.tsx
```diff
@@ -54,8 +54,8 @@ export const EntriesList: React.FC<EntriesListProps> = ({entries, setEntries, fo
     const filterEntries = useCallback((entry) => {
         if(methodsFilter.length > 0 && !methodsFilter.includes(entry.method.toLowerCase())) return;
         if(pathFilter && entry.path?.toLowerCase()?.indexOf(pathFilter) === -1) return;
-        if(statusFilter.includes(StatusType.SUCCESS) && entry.statusCode >= 400) return;
-        if(statusFilter.includes(StatusType.ERROR) && entry.statusCode < 400) return;
+        if(statusFilter.includes(StatusType.SUCCESS) && entry.status_code >= 400) return;
```
gadotroee, line 57: I think that better solution is to change the status_code to be statusCode.
(in line 137 (tap/api/api.go).

Not sure what is using this as the snake case formation (and maybe it will be harder) but we are using cameCase for json keys.

mertyildiran: @gadotroee I used snake_case in a lot of places. Without realizing the project's JSON style guide of choice is camelCase. There are also PascalCase(s) coming from the dissectors since there are structs which don't have `json:"payload"` definition.

Fixing the JSON styling of the project is not the concern of this PR. It should be done by a separate PR. This PR is fixing the filtering feature.

gadotroee: I understand and see how this pretty small PR (nice of that) fixing the issue (because it accessed incorrect field before).
I'm approving this PR and it will be good to do a PR that fix the styling (it can be also be fixing to snake case if it is easier) but it will be much better to have one styling..

mertyildiran: @gadotroee thanks! The JSON style guide of choice should be stated in this file https://github.com/up9inc/mizu/blob/develop/CONTRIBUTE.md

Also that file should be renamed to `CONTRIBUTING.md` (not  `CONTRIBUTE.md`)

gadotroee (approved):

PR #268, opened 2021-09-10
Make `ScrollableFeed` virtualized by replacing `react-scrollable-feed` with `react-scrollable-feed-virtualized`

mertyildiran:
Replaces [react-scrollable-feed](https://www.npmjs.com/package/react-scrollable-feed) with [react-scrollable-feed-virtualized](https://www.npmjs.com/package/react-scrollable-feed-virtualized) to fix the performance issue related to rendering a huge number of list items.


mertyildiran:
I don't know why `npm install` removed 22,792 lines from `package-lock.json` but the React app works fine.

lirazyehezkel (commented):
It looks like a great solution for the perf issue - great job!
I run the UI and I see some issues though:

1. There should be a "fetch more entries" button when we are in pause mode and the scrollbar is attached to the bottom and I do not see it on the screen (you can look at the EntriesList.tsx file in row 120). We want this button to be able to retrieve new entries that came during the pause - in the current state that the button does not appear, this is not possible.

2. When we scroll to the top, a request is called to retrieve old entries. In this situation, when another 50 entries are added from above, a behavior was defined that the scrollbar descends back to the last entry we watched. It does not seem to be happening now - so as soon as I scroll to the top, if I want to pull out 50 more I will have to lower the scrollbar a bit and then lift it back

3. The fetched old entries include an empty protocol object for some reason

4. I saw that you removed the restriction to the 1000 entries we hold in FE. I'm wondering if we still need to limit some number - I'm a little scared to blow up the browser's memory since that number can grow to infinity. What do you think?

Let me know if you can handle these bugs and feel free to reach out for questions or any help

image

https://user-images.githubusercontent.com/61656597/133047734-5e86303f-3155-415f-aba3-7bcd570fac0e.mov

lirazyehezkel (dismissed):

PR #271, opened 2021-09-13
[TRA-3659] Fix rules

seltonfiuza:
Working!


tap/api/api.go
```diff
@@ -132,11 +133,16 @@ type MizuEntry struct {
...
+	RulesMatched []map[string]interface{} `json:"rulesMatched,omitempty"`
+	Entry        MizuEntry            `json:"entry"`
+}
+
 type MizuEntryWrapper struct {
 	Protocol       Protocol  `json:"protocol"`
 	Representation string    `json:"representation"`
 	BodySize       int64     `json:"bodySize"`
-	Data           MizuEntry `json:"data"`
+	Data           FullEntryWithPolicy `json:"data"`
```
gadotroee, line 145:
```suggestion
	Data MizuEntry `json:"data"`
	RulesMatched []map[string]interface{} `json:"rulesMatched,omitempty"`	
```

gadotroee: I think it will be better to put the rules near the entry and not wrap it in single item because it creates thenested object. so in frontend you have it like `data.entry.entry` 
wdyt?

agent/pkg/controllers/entries_controller.go
```diff
@@ -140,11 +140,26 @@ func GetEntry(c *gin.Context) {
...
+		var pair tapApi.RequestResponsePair
+		json.Unmarshal([]byte(entryData.Entry), &pair)
+		harEntry, _ := utils.NewEntry(&pair)
+		_, rulesMatched := models.RunValidationRulesState(*harEntry, entryData.Service)
+		inrec, _ := json.Marshal(rulesMatched)
+		var rules []map[string]interface{}
+		json.Unmarshal(inrec, &rules)
+		fewp.RulesMatched = rules
+	}
+
```
gadotroee, line 157:
```suggestion
        var rules []map[string]interface{}
	if entryData.ProtocolName == "http" {
		var pair tapApi.RequestResponsePair
		json.Unmarshal([]byte(entryData.Entry), &pair)
		harEntry, _ := utils.NewEntry(&pair)
		_, rulesMatched := models.RunValidationRulesState(*harEntry, entryData.Service)
		inrec, _ := json.Marshal(rulesMatched)
		json.Unmarshal(inrec, &rules)
	}
	
```

agent/pkg/controllers/entries_controller.go
```diff
@@ -140,11 +140,26 @@ func GetEntry(c *gin.Context) {
...
+		json.Unmarshal(inrec, &rules)
+		fewp.RulesMatched = rules
+	}
+	
 	c.JSON(http.StatusOK, tapApi.MizuEntryWrapper{
 		Protocol:       protocol,
 		Representation: string(representation),
 		BodySize:       bodySize,
-		Data:           entryData,
+		Data:           fewp,
```
gadotroee, line 162:
```suggestion
		Data:           entryData,
		Rules:          rules
		  
```

gadotroee (commented):
I added small comments that can be improvements IMO.
Also we think that we need to add acceptance tests to this feature (there is documentation  about how to write them and we will share it soon so you will have example on how to do it)

with that said - it might be possible to merge and add the tests after (if we see it takes too much time and want to integrate / test it in develop soon)

mertyildiran:
I've made a subjective change here https://github.com/up9inc/mizu/commit/e32bf7caa7325180752aacba5f4344a2c1ecaeb4

I believe it fits better to the design language. You can revert it if you want.

tap/api/api.go
```diff
@@ -132,11 +132,17 @@ type MizuEntry struct {
 	EstimatedSizeBytes      int    `json:"-" gorm:"column:estimatedSizeBytes"`
 }
 
+type FullEntryWithPolicy struct {
```
gadotroee, line 135: I think this type is not used, am I missing something?
@seltonfiuza  @mertyildiran  ?

agent/pkg/models/models.go
```diff
@@ -105,33 +97,8 @@ type ExtendedCreator struct {
 	Source *string `json:"_source"`
 }
 
-type FullEntryWithPolicy struct {
-	RulesMatched []rules.RulesMatched `json:"rulesMatched,omitempty"`
-	Entry        har.Entry            `json:"entry"`
-	Service      string               `json:"service"`
+func RunValidationRulesState(harEntry har.Entry, service string) (tapApi.ApplicableRules, []rules.RulesMatched) {
+	numberOfRules, resultPolicyToSend := rules.MatchRequestPolicy(harEntry, service)
```
gadotroee, line 101: here there is a weird thing.. 
the numnerOfRules is a parameter defined when it is returned from MatchRequestPolicy function and then reassigned in the seconds function? (and I see the second parameter to the PassedValidationRules function is redundant (because it is not used inside)..
again - maybe I miss something but I think there is small fix needed here...

seltonfiuza: Yes you are right, this is from a really old implementation but that it was present until now, removed

tap/api/api.go
seltonfiuza: Yes you are right, this is from a really old implementation but that it was present until now, removed

gadotroee: Nice

gadotroee (approved):
Looks good IMO.

PR #272, opened 2021-09-13
Move `stats_tracker.go` into the extension API and increment `MatchedPairs` from inside the `Emit` method

mertyildiran:
Fixes the logged `matchedPairs` value.

Also replaces multiple `sync.Mutex`(es) with low-level atomic memory primitives.


gadotroee:
I think it is a great change. 
much less code (with all mutexes) and it might also improve some performance.

gadotroee (approved):

PR #280, opened 2021-09-18
Bring back the sensitive data filtering feature

mertyildiran:


gadotroee (commented):
Looks very nice (not that big change).
Just to be sure I understand correctly `sensitive_data_cleaner.go` is a copied functionality (that probably changed a little bit because of the new format) of the logic we had before, right?

tap/passive_tapper.go
```diff
@@ -96,6 +96,8 @@ var ownIps []string             // global
 var hostMode bool               // global
 var extensions []*api.Extension // global
 
+var filteringOptions *api.TrafficFilteringOptions
```
gadotroee, line 99: add //global comment (to be like others..)

mertyildiran:
> Looks very nice (not that big change).
> Just to be sure I understand correctly `sensitive_data_cleaner.go` is a copied functionality (that probably changed a little bit because of the new format) of the logic we had before, right?

@gadotroee yes, it's providing the same functionality of https://github.com/up9inc/mizu/blob/17fa163ee310d94c3c6efea0509b014312290842/agent/pkg/sensitiveDataFiltering/messageSensitiveDataCleaner.go

tap/passive_tapper.go
mertyildiran: @gadotroee added with https://github.com/up9inc/mizu/commit/73ed5b9a7776cfe87f77c9fe86159e79e57f0339

gadotroee: 👍

gadotroee (approved):
Looking good

PR #282, opened 2021-09-18
[refactor/TRA-3693] type:latency to slo and latency field to response-time

seltonfiuza:


shared/models.go
```diff
@@ -99,14 +99,16 @@ type RulesMatched struct {
...
+		log.Printf("Error: %s. ", r.Name)
+		log.Fatalf("Only json, header and slo types are supported on rule definition.\n")
 	}
-	if strings.ToLower(r.Type) == "latency" {
-		if r.Latency == 0 {
-			fmt.Printf("\nRule with name %s will be ignored. Err: when type=latency, the field Latency should be specified and have a value >= 1\n\n", r.Name)
+	if strings.ToLower(r.Type) == "slo" {
+		if r.ResponseTime == 0 {
+			log.Printf("Error: %s. ", r.Name)
+			log.Fatalf("When type=slo, the field response-time should be specified and have a value >= 1\n\n")
```
gadotroee, line 111: I guess this is not something ready for merge (Or I don't understand why do we have here log.Fatal...)

shared/models.go
```diff
@@ -89,7 +89,7 @@ type RulePolicy struct {
 	Method  string `yaml:"method"`
 	Key     string `yaml:"key"`
 	Value   string `yaml:"value"`
-	Latency int64  `yaml:"latency"`
+	ResponseTime int64  `yaml:"response-time"`
```
gadotroee, line 92: formatting, this struct: 
type RulePolicy struct {
	Type    string `yaml:"type"`
	Service string `yaml:"service"`
	Path    string `yaml:"path"`
	Method  string `yaml:"method"`
	Key     string `yaml:"key"`
	Value   string `yaml:"value"`
	ResponseTime int64  `yaml:"response-time"`
	Name    string `yaml:"name"`
}

should be formatted like this:

```
type RulePolicy struct {
	Type         string `yaml:"type"`
	Service      string `yaml:"service"`
	Path         string `yaml:"path"`
	Method       string `yaml:"method"`
	Key          string `yaml:"key"`
	Value        string `yaml:"value"`
	ResponseTime int64  `yaml:"response-time"`
	Name         string `yaml:"name"`
}
```

(types and annotations of the properties start from the same place)

gadotroee (changes requested):
not sure about the log.Fatal and error we are printing

shared/models.go
seltonfiuza: Now i'm printing that the rule will be ignored

shared/models.go
seltonfiuza: Sorry about that, i'm formatting manually, forgot about it.

shared/models.go
```diff
@@ -99,14 +99,17 @@ type RulesMatched struct {
...
-		fmt.Printf("\nRule with name %s will be ignored. Err: only json, header and latency types are supported on rule definition.\n", r.Name)
+		log.Printf("Error: %s. ", r.Name)
+		log.Printf("Only json, header and slo types are supported on rule definition. This rule will be ignored\n")
+		found = false
 	}
-	if strings.ToLower(r.Type) == "latency" {
-		if r.Latency == 0 {
-			fmt.Printf("\nRule with name %s will be ignored. Err: when type=latency, the field Latency should be specified and have a value >= 1\n\n", r.Name)
+	if strings.ToLower(r.Type) == "slo" {
+		if r.ResponseTime == 0 {
```
IgorGov, line 110: in the log you are saying:
"response-time should be specified and have a value >= 1"
but here you are checking "ResponseTime== 0" why ?

shared/models.go
```diff
@@ -148,10 +147,25 @@ func DecodeEnforcePolicy(path string) (RulesPolicy, error) {
...
+				enforcePolicy.Rules[k] = rule
+				k++
+			}
 		}
+		enforcePolicy.Rules = enforcePolicy.Rules[:k]
 	}
 	return enforcePolicy, nil
 }
+
+func contains(s []int, num int) bool {
```
IgorGov, line 163: we have already "contains" function in cli/mizu/sliceUtils. can we move it to shared and make it more generic?

shared/models.go
seltonfiuza: it is <=0 now.
I think that we need to consider only positive values for that, if the user pass 0 or negative the rule will be ignored

gadotroee (approved):

PR #286, opened 2021-09-19
Fix some errors related to conversion to HAR

mertyildiran:


agent/pkg/utils/har.go
```diff
@@ -203,9 +202,11 @@ func NewResponse(response *api.GenericMessage) (harResponse *har.Response, err e
 
 	status := int(resDetails["status"].(float64))
 	if strings.HasPrefix(mimeType.(string), "application/grpc") {
-		status, err = strconv.Atoi(_status)
+		if _status != "" {
```
gadotroee, line 205: why it is happen in GRPC and not before when we converting to int? (status := int(resDetails["status"].(float64)))

gadotroee (commented):
small comment (question)

agent/pkg/utils/har.go
mertyildiran: @gadotroee the variable name is `_status` and not `status`. It's different, it comes from `:status` header.

mertyildiran: @gadotroee OK fixed the wrong variable name in `BuildHeaders` function.

agent/pkg/utils/har.go
```diff
@@ -205,7 +204,7 @@ func NewResponse(response *api.GenericMessage) (harResponse *har.Response, err e
 	if strings.HasPrefix(mimeType.(string), "application/grpc") {
 		status, err = strconv.Atoi(_status)
 		if err != nil {
-			tap.SilentError("convert-response-status-for-har", "Failed converting status to int %s (%v,%+v)", err, err, err)
+			rlog.Debugf("Failed converting status to int %s (%v,%+v)", err, err, err)
```
gadotroee, line 207: change to error?

mertyildiran: @gadotroee I didn't understand.

mertyildiran: @gadotroee oh OK `Errorf` done.

gadotroee (approved):

PR #298, opened 2021-09-22
Omit the `RESPONSE` tab and `elapsedTime` if the `response` is empty

mertyildiran:
This PR omits the the `RESPONSE` tab and `elapsedTime` if the `response` is empty.

### DIFF
--- a/ui/src/components/UI/Tabs.tsx
+++ b/ui/src/components/UI/Tabs.tsx
@@ -17,6 +17,7 @@ interface Props {
     tabs: Tab[],
     currentTab: string,
     color: string,
+    response: any,
     onChange: (string) => void,
     leftAligned?: boolean,
     dark?: boolean,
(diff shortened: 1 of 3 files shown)


IgorGov:
@An1mus can you please review this one as @lirazyehezkel is on a vacation

gadotroee (approved):
lgtm

ui/src/components/UI/Tabs.tsx
```diff
@@ -72,10 +73,13 @@ const useTabsStyles = makeStyles((theme) => ({
 }));
 
 
-const Tabs: React.FC<Props> = ({classes={}, tabs, currentTab, color, onChange, leftAligned, dark}) => {
+const Tabs: React.FC<Props> = ({classes={}, tabs, currentTab, color, response, onChange, leftAligned, dark}) => {
     const _classes = {...useTabsStyles(), ...classes};
     return <div className={`${_classes.root} ${leftAligned ? _classes.tabsAlignLeft : ''}`}>
         {tabs.filter((tab) => !tab.hidden).map(({tab, disabled, disabledMessage, highlight, badge}, index) => {
+            if (!response && index === 1) {
```
gadotroee, line 80: maybe it can be changed to be hidden when there is no response? (I think it will be shown now with no content and I don't see reason for it)
 
@mertyildiran

mertyildiran: @gadotroee that filter is iterating through this array:

https://github.com/up9inc/mizu/blob/ee985eee37e1a06ef30fd3ff4a51673d0596d8ce/ui/src/components/EntryDetailed/EntryViewer.tsx#L37-L47

`index === 1` indicate the `RESPONSE` tab and if the `response` is `false` we don't render the tab with this check. So the result looks like this:

Screenshot from 2021-09-22 13-56-00

What do you suggest other than `return <></>;`? As far as I know, we need to return something in React, maybe I'm wrong.

gadotroee: I understand it is filtering the RESPONSE but I also see that we have filtering by the hidden parameter.
so maybe it is better to change the hidden in this case (!response) and then it can be removed

mertyildiran: @gadotroee oh yeah OK, there is `tab.hidden`. Let me use that instead...

gadotroee (changes requested):
use the hidden param

gadotroee (approved):
I thought i already approved this one and see it didn't send my review

PR #302, opened 2021-09-23
Don't omit the `Latency` field in `BaseEntryDetails`

mertyildiran:
Since the `Latency` field was being omitted, the boolean expression below was becoming `false`:

https://github.com/up9inc/mizu/blob/43f44af0dd2f376bf508551b36bb800787a160c4/ui/src/components/EntryListItem/EntryListItem.tsx#L71-L77

The **rules.yaml** file below reproduces it:

```yaml
rules:
- name: latency-test
  type: latency
  latency: 1
  service: ".*"
  path: "/*"
```

Before:

Screenshot from 2021-09-23 14-21-26

After:

Screenshot from 2021-09-23 14-25-51


gadotroee (approved):
looking good

PR #303, opened 2021-09-23
Omit the `RULES` tab if the policy rules feature is inactive

mertyildiran:
This PR omits the `RULES` tab if the policy rules feature is inactive.

`--traffic-validation-file` option is **not** supplied:

Screenshot from 2021-09-24 14-42-16

`--traffic-validation-file` option is supplied:

Screenshot from 2021-09-24 14-42-52

Screenshot from 2021-09-24 14-42-57

Screenshot from 2021-09-24 20-13-03

Screenshot from 2021-09-24 20-14-59

### DIFF
--- a/ui/src/components/EntryDetailed.tsx
+++ b/ui/src/components/EntryDetailed.tsx
@@ -72,7 +72,7 @@ export const EntryDetailed: React.FC<EntryDetailedProps> = ({entryData}) => {
         />
         {entryData.data && <EntrySummary data={entryData.data}/>}
         <>
-            {entryData.data && <EntryViewer representation={entryData.representation} rulesMatched={entryData.rulesMatched} elapsedTime={entryData.data.elapsedTime} color={entryData.protocol.backgroundColor}/>}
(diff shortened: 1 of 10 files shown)


ui/src/components/EntryDetailed.tsx
```diff
@@ -62,6 +62,7 @@ const EntrySummary: React.FC<any> = ({data}) => {
 };
 
 export const EntryDetailed: React.FC<EntryDetailedProps> = ({entryData}) => {
+    console.log("entryData.rulesMatched:", entryData.rulesMatched)
```
gadotroee, line 65: I guess this is a debug log that should be removed?
@mertyildiran

gadotroee (commented):
after removing the debug log I think this is a good change

ui/src/components/EntryDetailed.tsx
mertyildiran: @gadotroee I'm sorry, I forgot to remove that.

gadotroee (approved):
Good change.

PR #304, opened 2021-09-24
Start the tapper after the API server is ready and watch, stream events to web UI through WebSocket

mertyildiran:
**NOTE:** Toast messages are removed from this PR upon request. See the comments after https://github.com/up9inc/mizu/pull/304#issuecomment-927650086

---

This PR;

- Starts the tapper after the API server is ready.
- Prints an error in the CLI if the API server couldn't be deployed because of insufficient resources:

```text
Tapping pods in namespaces "sock-shop"
+carts-5db79fbddf-qw28s
+carts-db-6c6c68b747-psx54
+catalogue-5f4cb4f68b-9g9ks
+catalogue-db-96f6f6b4c-2vm7k
+front-end-5c89db9f57-5q6ss
+orders-8458b7f5db-bfff9
+orders-db-659949975f-dj765
+payment-f58b8c445-6bxkl
+queue-master-84bbb789b7-v6wnq
+rabbitmq-5bcbb547d7-4cbrs
+session-db-7cf97f8d4f-z7jn6
+shipping-548f696b44-mvx6s
+user-756b89d69c-tt7hc
+user-db-6df7444fc-72vrg
Cannot deploy the API server. Reason: "0/1 nodes are available: 1 Insufficient cpu."
Error creating resources: Post "https://192.168.99.103:8443/api/v1/namespaces/mizu/services": context canceled

Removing mizu resources
```
- Prints a similar message for the tapper in case of the same insufficient resources condition:
```text
...
+user-756b89d69c-tt7hc
+user-db-6df7444fc-72vrg
Mizu is available at http://localhost:8899/mizu
Cannot deploy the tapper. Reason: "0/1 nodes are available: 1 Insufficient cpu."

Removing mizu resources
```

- Removes the 25 seconds hard-coded timeout of API server deployment. (the timeout of Mizu Agent image pull and deployment as the API server)

- Prints an error in the CLI if an incorrect agent image is supplied:

```text
...
+user-756b89d69c-nsghq
+user-db-6df7444fc-wddmp
Cannot deploy the API server. (ErrImagePull) Reason: "rpc error: code = Unknown desc = Error response from daemon: pull access denied for mertyildiran/mizuagent-123, repository does not exist or may require 'docker login': denied: requested access to the resource is denied"

Removing mizu resources
```

- Prints an error in the CLI if there is a network error while pulling the image:

```text
...
+user-756b89d69c-nsghq
+user-db-6df7444fc-wddmp
Cannot deploy the API server. (ErrImagePull) Reason: "rpc error: code = Unknown desc = Error response from daemon: Get https://registry-1.docker.io/v2/: dial tcp: lookup registry-1.docker.io: Temporary failure in name resolution"

Removing mizu resources
```

- Rest of the events that come from the tapper watch, are displayed as toast messages in the web UI:

Screenshot from 2021-09-25 22-10-05

Screenshot from 2021-09-25 22-10-09

Screenshot from 2021-09-25 22-18-15

Screenshot from 2021-09-25 22-18-18

Screenshot from 2021-09-25 23-24-51


IgorGov:
@mertyildiran super informative description, from the pictures it look good. One comment from my side when we say "tapper is terminated, increase pod resource" can we print where we should increase the resources?
In general i think we need to print some tapper "identifier" so the user won't see 2 messages of "tapper is running" and think we have some duplicate UI bug

mertyildiran:
> @mertyildiran super informative description, from the pictures it look good. One comment from my side when we say "tapper is terminated, increase pod resource" can we print where we should increase the resources?

@IgorGov what will be the exact text?

> In general i think we need to print some tapper "identifier" so the user won't see 2 messages of "tapper is running" and think we have some duplicate UI bug

How do I have multiple tappers deployed? I couldn't achieve that and I also couldn't understand how the code supports multiple tappers?

cli/cmd/tapRunner.go
```diff
@@ -598,6 +611,96 @@ func watchApiServerPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
 	}
 }
 
+func watchTapperPod(ctx context.Context, kubernetesProvider *kubernetes.Provider, cancel context.CancelFunc) {
+	podExactRegex := regexp.MustCompile(fmt.Sprintf("^%s.*", mizu.TapperDaemonSetName))
+	added, modified, removed, errorChan := kubernetes.FilteredWatch(ctx, kubernetesProvider, []string{config.Config.MizuResourcesNamespace}, podExactRegex)
+	var prevPodPhase core.PodPhase
+	for {
```
IgorGov, line 618: i think we should do here same as we did in watchApiServerPod, if not all tappers are up and running after 25 seconds we should cancel and close the application.
WDYT @gadotroee ?

IgorGov: I think we should print "Mizu is available at..." after all Mizu components are up and running which are APIServer and all tappers

ui/src/components/TrafficPage.tsx
```diff
@@ -108,9 +111,31 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
         }
     }
 
+    const openWebSocketCli = () => {
+        wsCli.current = new WebSocket("ws://localhost:8898/wsCli");
```
IgorGov, line 115: what will happen here when user running "mizu view"? (it doesn't create the tappers nor create the socket)

IgorGov:
> > @mertyildiran super informative description, from the pictures it look good. One comment from my side when we say "tapper is terminated, increase pod resource" can we print where we should increase the resources?
> 
> @IgorGov what will be the exact text?
> 
> > In general i think we need to print some tapper "identifier" so the user won't see 2 messages of "tapper is running" and think we have some duplicate UI bug
> 
> How do I have multiple tappers deployed? I couldn't achieve that and I also couldn't understand how the code supports multiple tappers?

1. The exact text can be machine name (it is more of a question not a comment), can we do it?
2. In a cluster where the application pods are on several nodes we will deploy several tappers. Tapper per node which have a pod which we would like to tap

ui/src/components/TrafficPage.tsx
mertyildiran: @IgorGov it just prints an error into console and it does not crash the React app.

mertyildiran:
> 1. The exact text can be machine name (it is more of a question not a comment), can we do it?

What is the machine here? The machine that deployed Mizu into the cluster? (which runs the Mizu CLI)

> 2. In a cluster where the application pods are on several nodes we will deploy several tappers. Tapper per node which have a pod which we would like to tap

Do we have such capability already? or is it planned? If we have that then how can I reproduce such a case?

cli/cmd/tapRunner.go
mertyildiran: @IgorGov once the API server pod is ready, the app is ready so it's opened in browser. Then the events related to tapper pod is streamed to the user via WebSocket in the form of toast messages.

If there are insufficient resources for the tapper pod then the deployment is canceled. If the tapper pod is terminated then the user is informed through toast messages. Such that they can restart Mizu with more resources.

The only two case I see here that might cause the termination of the tapper pod are:
- OOMKilled.
- A runtime error.

Other than these we can assume that it's guaranteed to be deployed if there are enough resources in the cluster.

mertyildiran: I mean termination case can only happen after the deployment so it will be deployed if there are enough resources.

gadotroee: the issue is that tappers are updates after initialization
the can be (and probably will be) restarted more than one time if there is change in the pods that matching the regex (because the nodes that we need to deploy the daemon set on can be changed).

I didn't take a deeper look on the PR so I will try to do it and see if I have comments

mertyildiran:
I've pushed this commit https://github.com/up9inc/mizu/pull/304/commits/62244f6f5d98a7f6cd6b519e441cc943c27d4cbe that covers the `ErrImagePull` case and removed the 25 seconds timeout with https://github.com/up9inc/mizu/pull/304/commits/7f7777b01050db43194093a18efb2068c8f2cb2f since it's no longer needed.

mertyildiran:
> In general i think we need to print some tapper "identifier" so the user won't see 2 messages of "tapper is running" and think we have some duplicate UI bug

> 2. In a cluster where the application pods are on several nodes we will deploy several tappers. Tapper per node which have a pod which we would like to tap

@IgorGov I've included the pod name to toast messages with https://github.com/up9inc/mizu/pull/304/commits/92baf2fad8d0e79a3a7adbbc3a00c8a4ac3cb64f

Now the messages look like this:

Screenshot from 2021-09-26 21-59-50

Screenshot from 2021-09-26 21-59-52

cli/cmd/tapRunner.go
mertyildiran: @gadotroee https://github.com/up9inc/mizu/pull/304/commits/4d3f0f7ad0959ed55cf7c0aadf7cae19189a3bd2 fixes the issues related to tapper updates. (removal and recreation)

IgorGov:
> the machine here? The machine that deployed Mizu into the cluster? (

A tapper is deployed to a k8s node (as daemonset), if the this node doesn't have enough resources, we will notify the user that tapper wan't able to boot up on node 'X' because of insufficient resources.

gadotroee:
We (me and Igor) talked about this change and have different comments that are bringing us to this conclusion:
*we want to remove the notification in the UI* - two reasons for this: 
1. currently we have 3 different places of notifications / information - (pod list at the top / https traffic detection in the bottom and now the toasts in the right bottom), this is not the best UX and will be confused to user. so we need to decide on better UX to those 3 details.
2. I personally don't like that we are starting the socket server in CLI - I think that we could use the agent<-> frontend socket communication to inform on this change (it will be used as well when mizu will not be interactive (work without CLI)

So the Action Items: 
1. remove the changes in the UI folder and the socket handler in CLI 
2. instead of Sending the information to UI via socket - Just write it in CLI (so he whole infra for this information which is very good change will be there but we will not send it to UI until we are sure how we want to do it.

If we do those 2 things we can still know about those events so it will be helpful but just not printed in the UI.

@mertyildiran  Is that make sense?

mertyildiran:
@gadotroee I thought about the alternatives below (as you've mentioned) before opening this PR:

1. Display the tapper events as CLI logs
2. Make the WebSocket pipeline CLI -> API Server -> Web UI (display toast messages)

But since it's stated that the information should be displayed inside the Web UI in the issue description, I've eliminated the first option. The second option complicates the already complicated API Server WebSocket code, therefore I went with CLI -> Web UI WebSocket. (we can have a third WebSocket in API Server to reduce complication)

I agree that CLI -> Web UI WebSocket is not the best option but I thought it was good enough since the deployment related issues is a concern only for the system administrators who will deploy the Mizu into the Kubernetes cluster and I assumed that most people run Mizu CLI locally (instead of a virtual machine for example).

If we go with the first option (CLI logs) then in case of tapper events like **OOMKilled and restart** happens too fast that while the user looking at the Web UI might not realize that such an event occurred if he/she is not looking at the CLI logs at the same time.

So we can go with either of the options but I believe **CLI logs** option has certain drawbacks experience-wise.

gadotroee:
> If we go with the first option (CLI logs) then in case of tapper events like **OOMKilled and restart** happens too fast that while the user looking at the Web UI might not realize that such an event occurred if he/she is not looking at the CLI logs at the same time.

First, great that you thought about those two option and I agree with all that you said. 
I understand now that reason was because the ticket description and we do want to add this information in UI but we are still not sure where to put it to make the general experience as best as we can (single place for all kind of notifications for example)

I think that for now, we are Ok that it will be "only" in CLI (I prefer it will be in stdout and not only logs, meaning it should be Info and not debug) and we will use it when we decide how to display it in UI (maybe using the socket between API server to FE and maybe adding this CLI <-> FE communication.

Is it sounds something that you can do in this PR (it will actually makes it smaller and we will review it again?)

mertyildiran:
@gadotroee I've rearranged the CLI websocket connection to have CLI -> API Server -> Web UI pipeline with https://github.com/up9inc/mizu/pull/304/commits/2b20632c6d067715aadb35d4a82812b9fe8aa036

Please take a look at it again.

mertyildiran:
If the toast messages are for some reason not desired then these toast related parts can be removed in the web UI:

https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/src/components/TrafficPage.tsx#L109-L118

https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/src/components/TrafficPage.tsx#L234-L244

https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/src/components/TrafficPage.tsx#L13-L14

https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/package.json#L26

and these data structures can be changed according to desired web UI functionality:

https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/shared/models.go#L48-L57

gadotroee:
> If the toast messages are, for some reason, not desired then these toast related parts can be removed in the web UI:
> 
> https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/src/components/TrafficPage.tsx#L109-L118
> 
> https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/src/components/TrafficPage.tsx#L234-L244
> 
> https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/src/components/TrafficPage.tsx#L13-L14
> 
> https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/ui/package.json#L26
> 
> and these data structures can be changed according to desired web UI functionality:
> 
> https://github.com/up9inc/mizu/blob/2b20632c6d067715aadb35d4a82812b9fe8aa036/shared/models.go#L48-L57

I see the change, like I said personally i didn't like the socket in CLI and this is look better but when I talked with @IgorGov  about it we agree that this is not the only problem (he said mainly about the UX - 3 different places for notifications that can confuse users).

I think for now as I said in the pervious comment - we prefer having the improvement of the infra (track failures of pod and write about them and not start tappers until api server is started) but without the UI logic.

So I think we need to remove the parts you mention here and just have in back of our mind that code for this is already exist (in git history) and can be added easily.

Hope it make sense to you too.. 
we are in holidays (again) until Wednesday so I will try to look before but maybe I will not get to it so just know that it might take a bit more time

mertyildiran:
@gadotroee as you've requested;
- I've made it log into the CLI with https://github.com/up9inc/mizu/pull/304/commits/3eae6cc6e1c96c4f008aa6a549ff413a7ae5ca80
- Removed the `react-toastify` related code with https://github.com/up9inc/mizu/pull/304/commits/13d597b2a4958aa4e7ef91297090074d94040e10

The CLI is not a WebSocket server as of https://github.com/up9inc/mizu/pull/304/commits/2b20632c6d067715aadb35d4a82812b9fe8aa036 It's just sending the events to the API server through WebSocket (client). The desired web UI behavior can be implemented later on using this `case` statement:

https://github.com/up9inc/mizu/blob/13d597b2a4958aa4e7ef91297090074d94040e10/ui/src/components/TrafficPage.tsx#L106-L107

gadotroee:
Wow it was quick. As I said I'm not available as usual but will try to get to it when I can.

ui/src/components/TrafficPage.tsx
```diff
@@ -64,6 +64,7 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
     const [disableScrollList, setDisableScrollList] = useState(false);
 
     const ws = useRef(null);
+    const wsCli = useRef(null);
```
gadotroee, line 67: this can be removed.

mertyildiran: @gadotroee oops, I'm sorry. I forgot to remove that line. Removed with https://github.com/up9inc/mizu/pull/304/commits/9428fa55ceab70e63b5883133899210d51cb69e0

agent/pkg/utils/socket_client.go
```diff
@@ -2,37 +2,12 @@ package utils
...
-		if err != nil {
-			rlog.Warnf("Failed connecting to websocket: %s, attempt: %v/%v, err: %s, (%v,%+v)", address, try, DEFAULT_SOCKET_RETRIES, err, err, err)
-			try++
-		} else {
-			break
-		}
-		time.Sleep(DEFAULT_SOCKET_RETRY_SLEEP_TIME)
-	}
-
+func ConnectToSocketServer(address string) (conn *websocket.Conn, err error) {
```
IgorGov, line 7: after we removed all function content, it looks like it just a wrapper function for "websocket.DefaultDialer.Dial" with no additional value. Do we still need this wrapper?

mertyildiran: @IgorGov removed the function with https://github.com/up9inc/mizu/pull/304/commits/91d938f79782936a00fb64b5a55ae4392d78b8c3

cli/cmd/common.go
```diff
@@ -50,6 +51,7 @@ func waitForFinish(ctx context.Context, cancel context.CancelFunc) {
 }
 
 func openBrowser(url string) {
+	go socket.OpenWebsocket(url)
```
IgorGov, line 54: why do we need websocket from cli?

cli/cmd/tapRunner.go
```diff
@@ -560,6 +556,21 @@ func watchApiServerPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
 			}
 
 			logger.Log.Debugf("Watching API Server pod loop, modified: %v", modifiedPod.Status.Phase)
+
+			if modifiedPod.Status.Phase == core.PodPending {
```
IgorGov, line 560: i am not sure what i feel regarding removing the timer: "if passed 25 sec -> exit"
In order to remove it we need to support all different scenarios that can happened to the pod and define how it will respond.
Currently (in this PR) we are checking only "Pending" and "Running" event, but there are other event like "failed", "unknown", "Succeeded":
go/pkg/mod/k8s.io/api@v0.21.2/core/v1/types.go:2509
So i feel much more safe to bring back the 25 seconds timer until we will know all cases here handled properly.
@gadotroee WDYT?

cli/cmd/tapRunner.go
```diff
@@ -586,13 +604,92 @@ func watchApiServerPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
+				text = fmt.Sprintf("%s %s", text, strings.ToLower(string(podStatus.Phase)))
+			}
+			socket.Send(messageType, autoClose, text, modifiedPod.ObjectMeta.Name, appendMetaname)
+		case _, ok := <-errorChan:
+			if !ok {
+				errorChan = nil
+				continue
+			}
+
+			logger.Log.Debugf("[ERROR] Tapper creation, watching %v namespace", config.Config.MizuResourcesNamespace)
```
IgorGov, line 688: can we print the error here?

cli/cmd/tapRunner.go
```diff
@@ -586,13 +604,92 @@ func watchApiServerPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
+	autoClose := uint(5000)
+	for {
+		select {
+		case addedPod, ok := <-added:
+			if !ok {
+				added = nil
+				continue
+			}
+
+			socket.Send("info", autoClose, "Tapper is created", addedPod.ObjectMeta.Name, appendMetaname)
```
IgorGov, line 631: lets remove the "report to socket" and log.debug this info

cli/cmd/tapRunner.go
```diff
@@ -586,13 +604,92 @@ func watchApiServerPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
+		case removedPod, ok := <-removed:
+			if !ok {
+				removed = nil
+				continue
+			}
+
+			socket.Send("warning", autoClose, "Tapper is removed", removedPod.ObjectMeta.Name, appendMetaname)
```
IgorGov, line 638: lets remove the "report to socket" and log.debug this info

cli/cmd/tapRunner.go
IgorGov: one more comment, It is good approach and we should have in mizu, but it is not related to this PR.
Let's separate it to different PR, in order to be focused, work effectively and merge sooner this PR.

cli/cmd/tapRunner.go
```diff
@@ -586,13 +604,92 @@ func watchApiServerPod(ctx context.Context, kubernetesProvider *kubernetes.Provi
...
+		case modifiedPod, ok := <-modified:
+			if !ok {
+				modified = nil
+				continue
+			}
+
+			if modifiedPod.Status.Phase == core.PodPending && modifiedPod.Status.Conditions[0].Type == core.PodScheduled && modifiedPod.Status.Conditions[0].Status != core.ConditionTrue {
+				msg := fmt.Sprintf("Cannot deploy the tapper. Reason: \"%s\"", modifiedPod.Status.Conditions[0].Message)
+				socket.Send("error", autoClose, msg, modifiedPod.ObjectMeta.Name, appendMetaname)
```
IgorGov, line 647: only in case of an error we should print it to user (logger.log.info) all other cases lets print to file only logger.log.debug

cli/socket/socket.go
```diff
@@ -0,0 +1,70 @@
...
+			logger.Log.Errorf("WebSocket JSON Marshal error:", err)
+		}
+		err = conn.WriteMessage(websocket.TextMessage, data)
+		if err != nil {
+			logger.Log.Errorf("WebSocket write error:", err)
+		}
+	}
+}
+
+func Send(_type string, autoClose uint, text string, metaname string, appendMetaname bool) {
```
IgorGov, line 49: lets move the log message to where it used without the socket, it feels like it is redundant here

cli/socket/socket.go
```diff
@@ -0,0 +1,70 @@
...
+			WebSocketMessageMetadata: &shared.WebSocketMessageMetadata{
+				MessageType: shared.WebSocketMessageTypeToast,
+			},
+			Data: msg,
+		}
+		data, err := json.Marshal(wsMsg)
+		if err != nil {
+			logger.Log.Errorf("WebSocket JSON Marshal error:", err)
+		}
+		err = conn.WriteMessage(websocket.TextMessage, data)
```
IgorGov, line 42: reporting to api server we will do in separate pr, when we will use it in front end

IgorGov:
Lets make this pr smaller, smaller changes are merged faster and hard to miss anything.
On big PR we can miss some bugs or unexpected behaviors, hard to review and and take much time to merge.
Can we make this PR just an "infra" PR for tapper booting issue, 
For instance code we can move to other PRs:
- API server watch 
- new websocket - reporting to API server

cli/cmd/common.go
mertyildiran: @IgorGov I just didn't revert the code related to streaming the events from CLI -> API Server -> Web UI such that the desired web UI behavior can be implemented later on. Related https://github.com/up9inc/mizu/pull/304#issuecomment-927976401

mertyildiran:
Closing upon @IgorGov's request.

PR #306, opened 2021-09-27
TRA-3679 Mizu rules font

lirazyehezkel:


An1mus (approved):

mertyildiran:
FYI https://github.com/up9inc/mizu/pull/271#issuecomment-920951298 No one disagreed then :slightly_smiling_face:

Italic indicates an optional information is displayed there.

mertyildiran:
Oh it's bold in the Figma design. But according to the Figma design, the number inside the parentheses has even greater `font-weight`.

lirazyehezkel:
@mertyildiran sorry, missed your comment.
anyway, Neim told me that Alon asked for him to change the font styling to be as we have in the Figma so I removed the italic for now and added the bold
we can change it again in the future if we will get another feedback

PR #334, opened 2021-10-09
Fix the interface conversion errors in Redis

mertyildiran:
Fixes a bunch of interface conversion errors in Redis dissector.

Screenshot from 2021-10-09 07-31-12

Screenshot from 2021-10-09 07-31-42

### DIFF
--- a/tap/extensions/redis/read.go
+++ b/tap/extensions/redis/read.go
@@ -296,12 +296,28 @@ func (p *RedisProtocol) Read() (packet *RedisPacket, err error) {
 	switch x.(type) {
 	case []interface{}:
 		array := x.([]interface{})
-		packet.Command = RedisCommand(strings.ToUpper(string(array[0].([]uint8))))
-		if len(array) > 1 {
-			packet.Key = string(array[1].([]uint8))
-		}
-		if len(array) > 2 {
-			packet.Value = string(array[2].([]uint8))
+		switch array[0].(type) {
+		case []uint8:
+			packet.Command = RedisCommand(strings.ToUpper(string(array[0].([]uint8))))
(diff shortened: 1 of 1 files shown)


tap/extensions/redis/read.go
```diff
@@ -296,12 +296,28 @@ func (p *RedisProtocol) Read() (packet *RedisPacket, err error) {
...
+				packet.Value = fmt.Sprintf("[%s", packet.Value)
+				for _, item := range array[3:] {
+					packet.Value = fmt.Sprintf("%s, %s", packet.Value, item.([]uint8))
+				}
+				packet.Value = strings.TrimSuffix(packet.Value, ", ")
+				packet.Value = fmt.Sprintf("%s]", packet.Value)
+			}
+		default:
+			msg := fmt.Sprintf("Unrecognized element in Redis array: %v\n", reflect.TypeOf(array[0]))
+			log.Printf(msg)
```
IgorGov, line 318: Can we use rlog here, as we do in all other places?

mertyildiran: @IgorGov no, we can't. Because `github.com/romana/rlog` is not in the extension's dependencies ([go.mod](https://github.com/up9inc/mizu/blob/eb4a5413768dcc2cbaf17301b6e75e8d0cf8d479/tap/extensions/redis/go.mod)). Also there are 37 other `log.Printf` occurrences in the codebase. Should I replace all of the `log.Printf` with `rlog.Debugf` with this PR?

IgorGov: not in this PR

IgorGov (approved):

PR #344, opened 2021-10-12
changed sync entries to start on startup

RoyUP9:


agent/main.go
```diff
@@ -275,3 +283,15 @@ func pipeTapChannelToSocket(connection *websocket.Conn, messageDataChannel <-cha
...
 	}
 }
+
+func getSyncEntriesConfig() *shared.SyncEntriesConfig {
+	var syncEntriesConfig = &shared.SyncEntriesConfig{}
+
+	syncEntriesConfigJson := os.Getenv(shared.SyncEntriesConfigEnvVar)
+	err := json.Unmarshal([]byte(syncEntriesConfigJson), syncEntriesConfig)
+	if err != nil {
+		panic(fmt.Sprintf("env var %s's value of %s is invalid! json must match the shared.SyncEntriesConfig struct, err: %v", shared.SyncEntriesConfigEnvVar, syncEntriesConfigJson, err))
```
IgorGov, line 293: it feels that in this case we should "log.exception" - which prints the exception to log and panic

agent/pkg/up9/main.go
```diff
@@ -133,7 +111,68 @@ func GetAnalyzeInfo() *shared.AnalyzeStatus {
 	}
 }
 
-func SyncEntriesImpl(token string, model string, envPrefix string, uploadIntervalSec int, guestMode bool) {
+func SyncEntries(syncEntriesConfig *shared.SyncEntriesConfig) error {
+	rlog.Infof("Sync entries - started\n")
+
+	if GetAnalyzeInfo().IsAnalyzing {
```
IgorGov, line 117: i think we can remove it already

PR #350, opened 2021-10-14
Replace all `rlog` occurrences with the shared logger

mertyildiran:


nimrod-up9 (dismissed):

cli/uiUtils/confirmation.go
```diff
@@ -15,7 +16,7 @@ func AskForConfirmation(s string) bool {
 
 	response, err := reader.ReadString('\n')
 	if err != nil {
-		log.Fatal(err)
+		logger.Log.Errorf(err.Error())
```
RoyUP9, line 19: I think it needs to be changed to fatal, and maybe add informative message like "Error while reading confirmation string, err: %v"

agent/pkg/utils/truncating_logger.go
```diff
@@ -24,21 +25,21 @@ func (truncatingLogger *TruncatingLogger) Info(_ context.Context, message string
 	if truncatingLogger.LogLevel < logger.Info {
 		return
 	}
-	rlog.Errorf("gorm info: %.150s", message)
+	loggerShared.Log.Errorf("gorm info: %.150s", message)
```
RoyUP9, line 28: do we need this file after the log refactor? is it used?

shared/logger/logger.go
```diff
@@ -28,3 +28,14 @@ func InitLogger(logPath string) {
...
 	logging.SetBackend(backend1Leveled, backend2Formatter)
 }
+
+func InitLoggerStderrOnly() {
+	consoleLog := logging.NewLogBackend(os.Stderr, "", 0)
+
+	backend1Formatter := logging.NewBackendFormatter(consoleLog, format)
+
+	backend1Leveled := logging.AddModuleLevel(consoleLog)
+	backend1Leveled.SetLevel(logging.INFO, "")
```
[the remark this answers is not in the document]
mertyildiran: @RoyUP9 oh yeah, I'm sorry. Changed it to `DEBUG` with https://github.com/up9inc/mizu/pull/350/commits/6eb3bc3e190903c450faff6837e6a9ead0d44a87

By looking at this line: 

https://github.com/up9inc/mizu/blob/b30b62ef77d83af6a5f51cb658861f29b8b5920d/shared/logger/logger.go#L27

I thought that, this is the desired debug level, sorry.

agent/pkg/utils/truncating_logger.go
IgorGov: hmm as the main author I would like to address this question to @RamiBerm :)
feels like a hot potato game

RamiBerm: This is still needed, its used to override the default sqlite logger.
The sqlite logger tends to log entire SQL queries for warnings/errors which can be huge for large entries, this class ensures the log lines are truncated at a certain point

RoyUP9 (approved):

PR #351, opened 2021-10-14
Ignore an eslint error

mertyildiran:
Introduced by [this change](https://github.com/up9inc/mizu/pull/346/files#diff-3550a8c8813bc37aef982dcb8f241c1756429b859b5ffc0ab9190cc2e3fdb522R33-R48).

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -35,7 +35,7 @@ const App = () => {
             try {
                 const recentTLSLinks = await api.getRecentTLSLinks();
                 if (recentTLSLinks?.length > 0) {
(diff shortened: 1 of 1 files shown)


ui/src/App.tsx
```diff
@@ -45,7 +45,8 @@ const App = () => {
             }
 
         })();
-    },[]);
+        // eslint-disable-next-line
```
lirazyehezkel, line 48: Try to change:
setAddressesWithTLS(new Set([...addressesWithTLS, ...recentTLSLinks]));
to:
setAddressesWithTLS(new Set(recentTLSLinks));
I think it may solve the problem

PR #356, opened 2021-10-14
errors map

dvdlevanon:


tap/errors_map.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/romana/rlog"
+)
+
+type errorsMap struct {
+	errorsMap      map[string]uint
+	outputLevel    int
+	nErrors        int
```
nimrod-up9, line 13: Did you change this from uint to int on purpose?

PR #365, opened 2021-10-17
Add readiness and liveness probes to API server

mertyildiran:
Related to https://github.com/up9inc/mizu/pull/309

Prevents WebSocket connection errors from tappers to the API server by adding readiness and liveness probes to API server.

### DIFF
--- a/cli/kubernetes/provider.go
+++ b/cli/kubernetes/provider.go
@@ -193,10 +193,7 @@ func (provider *Provider) CreateMizuApiServerPod(ctx context.Context, opts *ApiS
 		command = append(command, "--namespace", opts.Namespace)
 	}
 
-	port := intstr.IntOrString{
-		Type:   intstr.Int,
-		IntVal: 8899,
-	}
+	port := intstr.FromInt(8899)
 
 	pod := &core.Pod{
 		ObjectMeta: metav1.ObjectMeta{
--- a/cli/kubernetes/provider.go
+++ b/cli/kubernetes/provider.go
@@ -193,6 +193,11 @@ func (provider *Provider) CreateMizuApiServerPod(ctx context.Context, opts *ApiS
 		command = append(command, "--namespace", opts.Namespace)
 	}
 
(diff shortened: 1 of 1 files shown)


cli/kubernetes/provider.go
```diff
@@ -193,6 +193,11 @@ func (provider *Provider) CreateMizuApiServerPod(ctx context.Context, opts *ApiS
 		command = append(command, "--namespace", opts.Namespace)
 	}
 
+	port := intstr.IntOrString{
+		Type:   intstr.Int,
+		IntVal: 8899,
```
gadotroee, line 198: why not using the "port"

mertyildiran: @gadotroee what do you mean? Is the port number passed to this function as a variable? If so which one?

gadotroee: Oh, my bad, I saw a port down and didnt understand it was this param.
but I know that port can be changed by user input so it might be wrong to pass 8899 hard-coded here.
I will take a look where it is and reference it in a moment

mertyildiran: @gadotroee it's also hard-coded in here as well

https://github.com/up9inc/mizu/blob/c38a601e662c19a8216f8f7c26c485d2e2a234ff/agent/pkg/utils/utils.go#L39

gadotroee: @mertyildiran 

yes I checked again and saw it but I don't think it is good that we have 2 places that writes "8899" hard-coded because we might change one of them and don'w know it is exists in another places.
do we have way to pass the same value (make it const somewhere) [ I know it is shared between different packages and modules so maybe it is worth the effort but it can be specified in mizu/shared and then taken from there to those 2 places that you mentioned

gadotroee: I'm okay with merging this change (which solves an issue) and maybe in separated PR (working in small PRs) fixing the hard-cided 8899, 
wdyt?

gadotroee (approved):

PR #368, opened 2021-10-17
added upsert workspace before dumping traffic

RoyUP9:


agent/pkg/up9/main.go
```diff
@@ -145,6 +150,30 @@ func SyncEntries(syncEntriesConfig *shared.SyncEntriesConfig) error {
...
+			authHeader: {token},
+		},
+	}
+
+	response, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed request to upsert model, err: %v", err)
+	}
+
+	if response.StatusCode != 201 && response.StatusCode != 409 {
```
gadotroee, line 170: maybe add comment to explain it is ok if we created it or it is already exists (which happens with 409)

gadotroee (dismissed):

PR #369, opened 2021-10-17
Replace all `rlog` occurrences with the shared logger in `tap`

mertyildiran:


mertyildiran:
> No needed to init the logger? Maybe i missed it

@RoyUP9 it's initialized with this line:

https://github.com/up9inc/mizu/blob/c59b9b70da324aa57a11edad380d3d40fffe0f4e/agent/main.go#L43

PR #371, opened 2021-10-17
Fix the `go.mod` of `acceptanceTests`

mertyildiran:


gadotroee (approved):
I see that this time you enable it in develop and then reverted it so it should work

PR #373, opened 2021-10-18
Set the default log level for Agent to `INFO` and raise it to `DEBUG` if `dump-logs=true` is provided

mertyildiran:
This PR set the default log level for Agent to `INFO`.

If `dump-logs=true` is provided then raises it to `DEBUG`. Therefore `dump-logs` serves as "debug mode enabled".

In `DEBUG` there is too much logging going on so it can slow down the application a little bit. `INFO` by default is better.

Also removes the `Trace` method and replaces its calls with `Debug` in `passive_tapper.go`.

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -41,11 +41,7 @@ var extensions []*tapApi.Extension             // global
 var extensionsMap map[string]*tapApi.Extension // global
 
 func main() {
-	logLevel := logging.INFO
-	debugMode := os.Getenv(shared.DebugModeEnvVar) == "1"
-	if debugMode {
-		logLevel = logging.DEBUG
-	}
+	logLevel := determineLogLevel()
 	logger.InitLoggerStderrOnly(logLevel)
 	flag.Parse()
 	loadExtensions()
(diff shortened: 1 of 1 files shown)


agent/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/go-playground/validator/v10 v10.5.0
 	github.com/google/martian v2.1.0+incompatible
 	github.com/gorilla/websocket v1.4.2
+	github.com/op/go-logging v0.0.0-20160315200505-970db520ece7 // indirect
```
RoyUP9, line 15: indirect

agent/main.go
```diff
@@ -40,7 +41,12 @@ var extensions []*tapApi.Extension             // global
 var extensionsMap map[string]*tapApi.Extension // global
 
 func main() {
-	logger.InitLoggerStderrOnly()
+	logLevel := logging.INFO
```
RoyUP9, line 44: maybe extract to func

agent/go.mod
mertyildiran: @RoyUP9 https://github.com/up9inc/mizu/commit/e554e40f4a16f07bf7b1f2ee4a7d2544a5a5a19d removes this line.

agent/main.go
RoyUP9: i don't think that the functions that return the logging level are necessary

mertyildiran: @RoyUP9 done with https://github.com/up9inc/mizu/commit/f502187f81c2533f29726eaa0860493521842670 Please review it once more.

PR #374, opened 2021-10-19
Remove `github.com/romana/rlog` dependency completely

mertyildiran:


tap/api/api.go
```diff
@@ -7,13 +7,13 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"log"
```
RoyUP9, line 10: why are we using log and not the logger in shared?

tap/go.mod
```diff
@@ -3,15 +3,14 @@ module github.com/up9inc/mizu/tap
...
 	github.com/google/gopacket v1.1.19
-	github.com/romana/rlog v0.0.0-20171115192701-f018bc92e7d7
+	github.com/up9inc/mizu/shared v0.0.0
 	github.com/up9inc/mizu/tap/api v0.0.0
-	golang.org/x/net v0.0.0-20210224082022-3d97a244fca7
-	golang.org/x/sys v0.0.0-20210225134936-a50acf3fe073
-	golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d
-	golang.org/x/text v0.3.5
-	golang.org/x/tools v0.0.0-20210106214847-113979e3529a
+	golang.org/x/net v0.0.0-20210224082022-3d97a244fca7 // indirect
```
RoyUP9, line 10: indirect?

tap/api/api.go
mertyildiran: @RoyUP9 logging in protocol extensions is little bit tricky. Because there could be a lot of false negatives in protocol extensions. Since they blindly try to parse the TCP stream in their own protocol specification. In fact, logging should be minimized in the protocol extensions.

Introducing more dependencies to protocol extensions is also not great. (`go-logging`) Because protocol extensions meant to be small shared objects that do their jobs and nothing more. They are compiled separately and adding more dependencies increases their binary size. Hence the load time increases and the runtime slows down.

`github.com/up9inc/mizu/shared` is bloated by irrelevant code. Once something changes in that module, all extensions should be updated as well and since the code in `github.com/up9inc/mizu/shared` constantly, it's better to not depend on that module.

We'll eventually remove almost all logging from the protocol extensions but it's better to remove it gradually.

tap/go.mod
mertyildiran: @RoyUP9 why do you type "indirect?" whenever you see `// indirect`? Could you please share some context?

If the driver program and a Go plugin has a shared dependency, they should have the exact same version of that dependency otherwise the plugin cannot be loaded. (throws a segmentation fault) Standard library is outside of this since it comes with the language's itself.

tap/api/api.go
gadotroee: @mertyildiran 
We want to have those logs because it doesn't look right to comment out a lot of logs.
Also, there is no real reason to take the shared logger into extensions (anyway it is really one repo and they are taking it locally using the "replace"). 
If you can do it we will merge it and maybe think if there should be a better solution.

gadotroee: @mertyildiran 

Ok, we prefer it like this but just need to remove another 2 logs in api.go (line 254 and 266), those are also not relevant like others so we will loose it totally and not keeping some of them

tap/go.mod
RoyUP9: what i meant is that i'm not sure if we want the indirect to be there or not (because it wasn't there before the change).
to be honest, i don't really know when it should be there and when it shouldn't.

PR #416, opened 2021-10-31
TRA-3868 move tapped pod watch and tapper updating to shared

RamiBerm:


cli/cmd/tapRunner.go
```diff
@@ -151,14 +133,88 @@ func RunMizuTap() {
...
+			case managerErr := <-tapperSyncer.ErrorOut:
+				logger.Log.Errorf(uiUtils.Error, getErrorDisplayTextForK8sTapManagerError(managerErr))
+				cancel()
+			case tappedPodChanges := <-tapperSyncer.TapPodChangesOut:
+				if err := apiserver.Provider.ReportTappedPods(tapperSyncer.CurrentlyTappedPods); err != nil {
+					logger.Log.Debugf("[Error] failed update tapped pods %v", err)
+				}
+				displayTapPodChangesEvent(tappedPodChanges)
+			case <-ctx.Done():
+				return
```
IgorGov, line 186: can you add a log here as well?

shared/kubernetes/k8sTapManager.go
```diff
@@ -0,0 +1,217 @@
...
+	TapperResources          shared.Resources
+	ImagePullPolicy          core.PullPolicy
+	DumpLogs                 bool
+	IgnoredUserAgents        []string
+	MizuApiFilteringOptions  api.TrafficFilteringOptions
+	MizuServiceAccountExists bool
+}
+
+func CreateAndStartMizuTapperSyncer(ctx context.Context, kubernetesProvider *Provider, config TapperSyncerConfig, shouldUpdateTappers bool) (*MizuTapperSyncer, error) {
+	manager := &MizuTapperSyncer{
```
IgorGov, line 48: syncer

shared/kubernetes/k8sTapManager.go
```diff
@@ -0,0 +1,217 @@
...
+		context:             ctx,
+		CurrentlyTappedPods: make([]core.Pod, 0),
+		config:              config,
+		kubernetesProvider:  kubernetesProvider,
+		TapPodChangesOut:    make(chan TappedPodChangeEvent, 100),
+		ErrorOut:            make(chan K8sTapManagerError, 100),
+		shouldUpdateTappers: shouldUpdateTappers,
```
IgorGov, line 55: in cli you can run k8s get pods machining regex, print to user 
you can use list of currently tapped pods

shared/kubernetes/k8sTapManager.go
```diff
@@ -0,0 +1,217 @@
...
+		return nil, err
+	}
+
+	if shouldUpdateTappers {
+		if err := manager.updateMizuTappers(); err != nil {
+			return nil, err
+		}
+	}
+
+	go goUtils.HandleExcWrapper(manager.watchPodsForTapping)
```
IgorGov, line 68: the objective of this function is printing to mizu cli log when there is panic in go routine, it is not really needed in server pods

shared/kubernetes/errors.go
```diff
@@ -0,0 +1,26 @@
...
+// K8sTapManagerError implements the Error interface.
+func (e *K8sTapManagerError) Error() string {
+	return e.OriginalError.Error()
+}
+
+type ClusterBehindProxyError struct{}
+
+// ClusterBehindProxyError implements the Error interface.
+func (e *ClusterBehindProxyError) Error() string {
+	return "Cluster is behind proxy"
```
IgorGov, line 25: maybe something like "we suspect the cluster behind proxy like lens" (we don't have 100% confident this validation method)

shared/kubernetes/k8sTapManager.go
```diff
@@ -0,0 +1,217 @@
+package kubernetes
```
IgorGov, line 1: should the file name itself be "syncer" instead of "manager"?

shared/kubernetes/errors.go
RamiBerm: This string is not visible to users, its just there because I need to implement `Error() string` for this struct to count as an error

IgorGov (approved):

PR #429, opened 2021-11-03
TRA-3842 daemon acceptance tests

RamiBerm:


acceptanceTests/tap_test.go
```diff
@@ -973,3 +973,197 @@ func TestTapDumpLogs(t *testing.T) {
...
+				}
+			})
+
+			if err := tapCmd.Run(); err != nil {
+				t.Errorf("error occured while running the tap command, err: %v", err)
+				return
+			}
+
+			t.Logf("running command: %v", viewCmd.String())
+			if err := viewCmd.Start(); err != nil {
```
RoyUP9, line 1017: do we wanna run the view command before the tappers are ready?

acceptanceTests/tap_test.go
```diff
@@ -973,3 +973,197 @@ func TestTapDumpLogs(t *testing.T) {
...
+				break
+			}
+		}
+
+		if !podFound {
+			t.Errorf("unexpected result - expected pod not found, pod namespace: %v, pod name: %v", expectedPod.Namespace, expectedPod.Name)
+			return
+		}
+	}
+}
```
RoyUP9, line 1169: add newline

acceptanceTests/tap_test.go
```diff
@@ -973,3 +973,197 @@ func TestTapDumpLogs(t *testing.T) {
...
+	var namespacesCmd []string
+	for _, expectedPod := range expectedPods {
+		namespacesCmd = append(namespacesCmd, "-n", expectedPod.Namespace)
+	}
+	tapCmdArgs = append(tapCmdArgs, namespacesCmd...)
+
+	tapCmd := exec.Command(cliPath, tapCmdArgs...)
+	t.Logf("running command: %v", tapCmd.String())
+
+	viewCmd := exec.Command(cliPath, getDefaultViewCommandArgs()...)
```
RoyUP9, line 1103: same comment here about the view command

acceptanceTests/tap_test.go
```diff
@@ -973,3 +973,197 @@ func TestTapDumpLogs(t *testing.T) {
...
+			t.Cleanup(func() {
+				if err := runMizuClean(); err != nil {
+					t.Logf("error running mizu clean: %v", err)
+				}
+				if err := cleanupCommand(viewCmd); err != nil {
+					t.Logf("failed to cleanup view command, err: %v", err)
+				}
+			})
+
+			if err := tapCmd.Run(); err != nil {
```
nimrod-up9, line 1011:
```go
t.Logf("running command: %v", tapCmd.String())
```

or delete 1016.

acceptanceTests/tap_test.go
RamiBerm: we need view to be able to connect to the api server, you're right about the tappers tho, this might check entries too soon

acceptanceTests/tap_test.go
```diff
@@ -973,3 +973,197 @@ func TestTapDumpLogs(t *testing.T) {
 		return
 	}
 }
+
+func TestTapDaemon(t *testing.T) {
```
nimrod-up9, line 977: Give it a more detailed name.
This function doesn't test all of daemon mode capabilities and edge cases.
It only tests that traffic is recorded. For example, this test doesn't compare the recorded traffic to the transmitted messages, and it doesn't check that pods are filtered correctly.
This test is good in itself. I think that the names should be accurate so that we are clear on what the tests cover.
Maybe `TestDaemonSeeTraffic` or `TestDaemonListEntries`?

acceptanceTests/tap_test.go
```diff
@@ -973,3 +973,197 @@ func TestTapDumpLogs(t *testing.T) {
...
+			}
+			if err := retriesExecute(shortRetriesCount, entriesCheckFunc); err != nil {
+				t.Errorf("%v", err)
+				return
+			}
+		})
+	}
+}
+
+func TestTapDaemonMultipleNamespaces(t *testing.T) {
```
nimrod-up9, line 1073: Also needs a more detailed name.
From just looking at the name of this function, I thought that it tests the same thing as `TestTapDaemon` only in the specified namespaces, but it doesn't.
Maybe `TestDaemonSeePods`?

nimrod-up9: Add a negative test (in this function or a different one) that tests that a pod in another namespace is not seen. (`unexpectedPods`)

acceptanceTests/tap_test.go
RamiBerm: its necessary so the test can connect to the api

acceptanceTests/tap_test.go
RamiBerm: Was just following naming convention, this is a daemon version of the test `TestTap`, didnt realize this would be misleading

nimrod-up9: I think it is 🤷‍♂️ 
You don't have to fix all of the tests, just the new ones.

RoyUP9 (approved):

nimrod-up9 (approved):

PR #431, opened 2021-11-04
Fix clean again

RamiBerm:


cli/cmd/tapRunner.go
```diff
@@ -41,10 +41,11 @@ type tapState struct {
 }
 
 var state tapState
-var apiProvider = apiserver.NewProvider(GetApiServerUrl(), apiserver.DefaultRetries, apiserver.DefaultTimeout)
+var apiProvider *apiserver.Provider
```
RoyUP9, line 45: Do we still need to declare it here?

RoyUP9 (approved):

PR #434, opened 2021-11-05
Remove the `ERROR` level logs in `loadOAS` and let the caller handle them on `INFO` level

mertyildiran:


gadotroee (approved):
I'm thinking it is worth to have log in this "valid" situation that will say something like:
"Contract validation file does not exist, this feature will not be used" in Info severity. but I understand the error is mistake so I'm approving and if you think this log can help you can push a change and I will re-approve.

PR #436, opened 2021-11-05
Replace privileged container with `CAP_NET_RAW`, `CAP_NET_ADMIN` Linux capabilities (tapper)

mertyildiran:
This PR replaces privileged container configuration of tapper pod with `CAP_NET_RAW`, `CAP_NET_ADMIN` Linux capabilities.

I'm not sure if it satisfies all possible cases that the privileged container was satisfying. But as far as I can see it's able to capture the whole traffic just like a privileged container.

`NET_RAW` permits raw access to an interface for capturing directly off the wire (by default enabled in Kubernetes)
`NET_ADMIN` allows us to set an interface to promiscuous mode (required by the `-promisc` flag in the tapper)
See `CAP_NET_` prefixed Linux capabilities for more https://man7.org/linux/man-pages/man7/capabilities.7.html

_I'm sorry for the additional diff. [The VS Code Go extension](https://marketplace.visualstudio.com/items?itemName=golang.go) is automatically fixing the coding style, according to the [Effective Go](https://golang.org/doc/effective_go)._

### DIFF
--- a/shared/kubernetes/provider.go
+++ b/shared/kubernetes/provider.go
@@ -6,11 +6,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"net/url"
+	"path/filepath"
+	"regexp"
+
 	"github.com/up9inc/mizu/shared"
 	"github.com/up9inc/mizu/shared/logger"
 	"github.com/up9inc/mizu/shared/semver"
 	"github.com/up9inc/mizu/tap/api"
-	"io"
 	v1 "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
@@ -31,9 +35,6 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 	watchtools "k8s.io/client-go/tools/watch"
-	"net/url"
-	"path/filepath"
-	"regexp"
 )
 
 type Provider struct {
@@ -239,8 +240,8 @@ func (provider *Provider) GetMizuApiServerPodObject(opts *ApiServerOptions, moun
 			},
 		})
 		volumeMounts = append(volumeMounts, core.VolumeMount{
-			Name:             volumeClaimName,
-			MountPath:        shared.DataDirPath,
+			Name:      volumeClaimName,
+			MountPath: shared.DataDirPath,
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


gadotroee:
> This PR replaces privileged container configuration of tapper pod with `CAP_NET_RAW`, `CAP_NET_ADMIN` Linux capabilities.
> 
> I'm not sure if it satisfies all possible cases that the privileged container was satisfying. But as far as I can see it's able to capture the whole traffic just like a privileged container.
> 
> `NET_RAW` permits raw access to an interface for capturing directly off the wire (by default enabled in Kubernetes) `NET_ADMIN` allows us to set an interface to promiscuous mode (required by the `-promisc` flag in the tapper) See `CAP_NET_` prefixed Linux capabilities for more https://man7.org/linux/man-pages/man7/capabilities.7.html
> 
> _I'm sorry for the additional diff. [The VS Code Go extension](https://marketplace.visualstudio.com/items?itemName=golang.go) is automatically fixing the coding style, according to the [Effective Go](https://golang.org/doc/effective_go)._

about the diff, it is ok because there is option in Github to hide whitespace changes.
image

then the whole diff is the small change you made.

image


about the change itself - it can be a good idea - I think we will talk about it tomorrow and see if this can be added as change.

mertyildiran:
`privileged: false` seems like not enough to prevent escalations. So when they restrict, they usually do;

```yaml
  requiredDropCapabilities:
    - ALL
```
too as it seems. Therefore relying on Linux capabilities is also not a guaranteed solution either.

```yaml
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: restricted
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default,runtime/default'
    apparmor.security.beta.kubernetes.io/allowedProfileNames: 'runtime/default'
    apparmor.security.beta.kubernetes.io/defaultProfileName:  'runtime/default'
  namespace: sock-shop
spec:
  privileged: false
  # Required to prevent escalations to root.
  allowPrivilegeEscalation: false
  requiredDropCapabilities:
    - ALL
  # Allow core volume types.
  volumes:
    - 'configMap'
    - 'emptyDir'
    - 'projected'
    - 'secret'
    - 'downwardAPI'
    # Assume that ephemeral CSI drivers & persistentVolumes set up by the cluster admin are safe to use.
    - 'csi'
    - 'persistentVolumeClaim'
    - 'ephemeral'
  hostNetwork: false
  hostIPC: false
  hostPID: false
  runAsUser:
    # Require the container to run without root privileges.
    rule: 'MustRunAsNonRoot'
  seLinux:
    # This policy assumes the nodes are using AppArmor rather than SELinux.
    rule: 'RunAsAny'
  supplementalGroups:
    rule: 'MustRunAs'
    ranges:
      # Forbid adding the root group.
      - min: 1
        max: 65535
  fsGroup:
    rule: 'MustRunAs'
    ranges:
      # Forbid adding the root group.
      - min: 1
        max: 65535
  readOnlyRootFilesystem: false
```

PR #279, opened 2021-09-17
Migrate from SQLite to Basenine and introduce a new filtering syntax

mertyildiran:
Migrates from SQLite to [Basenine](https://github.com/up9inc/basenine) and introduces [a new filtering syntax](https://github.com/up9inc/basenine/wiki/BFL-Syntax-Reference).

Screenshot from 2021-10-29 06-16-13

Screenshot from 2021-10-29 06-17-02

Screenshot from 2021-10-29 06-17-50

Screenshot from 2021-10-29 06-18-45

Screenshot from 2021-10-29 06-19-41

Screenshot from 2021-10-29 06-20-32

Screenshot from 2021-10-29 06-20-56

Screenshot from 2021-10-29 06-21-23

Screenshot from 2021-10-29 06-24-08

Screenshot from 2021-10-29 06-25-17

Screenshot from 2021-10-29 06-52-40

Screenshot from 2021-10-29 07-17-47

Screenshot from 2021-10-29 07-18-29

Screenshot from 2021-10-29 07-21-17

PR #443, opened 2021-11-09
Fix the acceptance tests after the merger of #279

mertyildiran:
Fixes the failing acceptance tests after the merger of https://github.com/up9inc/mizu/pull/279


cli/cmd/viewRunner.go
```diff
@@ -58,7 +58,9 @@ func runMizuView() {
 
 	logger.Log.Infof("Mizu is available at %s\n", url)
 
-	uiUtils.OpenBrowser(url)
+	if !config.Config.Tap.HeadlessMode {
```
gadotroee, line 61: I think it should not be here.
(we are taking tap field in the view command) 
and when user runs the view command it make sense to always open browser (IMO).

gadotroee (commented):
small comment on the headless param on the view command

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
 	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/assert"
 )
 
+// waitTimeout waits for the waitgroup for the specified max timeout.
+// Returns true if waiting timed out.
+func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
+	c := make(chan struct{})
```
RoyUP9, line 24: c is a non informative name

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
+	}
+}
+
+// checkDBHasEntries checks whether there are any entries in the database
+// before the given timestamp. Returns a slice of non-empty entries if it succeeds.
+func checkDBHasEntries(t *testing.T, timestamp int64, limit int) (entries []map[string]interface{}) {
+	query := fmt.Sprintf("timestamp < %d and limit(%d)", timestamp, limit)
+	webSocketUrl := getWebSocketUrl(defaultApiServerPort)
+
+	c, _, err := websocket.DefaultDialer.Dial(webSocketUrl, nil)
```
RoyUP9, line 43: c is a non informative name

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
+	go handleWSConnection(&wg)
+	wg.Add(1)
+
+	waitTimeout(&wg, 1*time.Second)
+
+	if len(entries) == 0 {
+		t.Error("unexpected entries result - Expected more than 0 entries")
+	}
+
+	return
```
RoyUP9, line 78: no need for return here

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
+	c, _, err := websocket.DefaultDialer.Dial(webSocketUrl, nil)
+	assert.Nil(t, err)
+	defer c.Close()
+
+	handleWSConnection := func(wg *sync.WaitGroup) {
+		defer wg.Done()
+		for {
+			_, message, err := c.ReadMessage()
+			if err != nil {
+				return
```
RoyUP9, line 52: don't we need to fail the test if there was an error here?

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
+}
+
+// checkDBHasEntries checks whether there are any entries in the database
+// before the given timestamp. Returns a slice of non-empty entries if it succeeds.
+func checkDBHasEntries(t *testing.T, timestamp int64, limit int) (entries []map[string]interface{}) {
+	query := fmt.Sprintf("timestamp < %d and limit(%d)", timestamp, limit)
+	webSocketUrl := getWebSocketUrl(defaultApiServerPort)
+
+	c, _, err := websocket.DefaultDialer.Dial(webSocketUrl, nil)
+	assert.Nil(t, err)
```
RoyUP9, line 44: the convention we are using to assert the test is like so:
```suggestion
	if err != nil {
		t.Errorf("%v", err)
		return
	}
```

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
+	handleWSConnection := func(wg *sync.WaitGroup) {
+		defer wg.Done()
+		for {
+			_, message, err := c.ReadMessage()
+			if err != nil {
+				return
+			}
+
+			var data map[string]interface{}
+			err = json.Unmarshal([]byte(message), &data)
```
RoyUP9, line 56: same assert convention here
```suggestion
			if err = json.Unmarshal([]byte(message), &data); err != nil {
				t.Errorf("%v", err)
				return
			}
```

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
+			assert.Nil(t, err)
+
+			if data["messageType"] == "entry" {
+				entries = append(entries, data)
+			}
+		}
+	}
+
+	err = c.WriteMessage(websocket.TextMessage, []byte(query))
+	assert.Nil(t, err)
```
RoyUP9, line 66: same assert convention here
```suggestion
	if err != nil {
		t.Errorf("%v", err)
		return
	}
```

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
+	case <-c:
+		return false // completed normally
+	case <-time.After(timeout):
+		return true // timed out
+	}
+}
+
+// checkDBHasEntries checks whether there are any entries in the database
+// before the given timestamp. Returns a slice of non-empty entries if it succeeds.
+func checkDBHasEntries(t *testing.T, timestamp int64, limit int) (entries []map[string]interface{}) {
```
RoyUP9, line 39: this function can be moved to test utils

acceptanceTests/tap_test.go
```diff
@@ -10,10 +10,74 @@ import (
...
 	"testing"
 	"time"
+
+	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/assert"
 )
 
+// waitTimeout waits for the waitgroup for the specified max timeout.
+// Returns true if waiting timed out.
+func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
```
RoyUP9, line 23: this function can be moved to test utils

acceptanceTests/tap_test.go
mertyildiran: It's a `missing return` compile-time error if we remove the `return` statement there.

acceptanceTests/tap_test.go
mertyildiran: The `err` is not `nil` in case of WebSocket connection is closed which is `defer`ed above. It's not a criteria for a test failure I believe.

acceptanceTests/tap_test.go
RoyUP9: so why do you need to have a return value when you don't return the entries at all?

mertyildiran: `entries []map[string]interface{}` is defined in the method signature:

https://github.com/up9inc/mizu/blob/3e2102e262089f85b69c93b5ea25980b1f7e6824/acceptanceTests/testsUtils.go#L339

Populated on this line:

https://github.com/up9inc/mizu/blob/3e2102e262089f85b69c93b5ea25980b1f7e6824/acceptanceTests/testsUtils.go#L365

Checked on this line:

https://github.com/up9inc/mizu/blob/3e2102e262089f85b69c93b5ea25980b1f7e6824/acceptanceTests/testsUtils.go#L382

Returned on this line:

https://github.com/up9inc/mizu/blob/3e2102e262089f85b69c93b5ea25980b1f7e6824/acceptanceTests/testsUtils.go#L386

I don't understand what's the problem here?

RoyUP9: I see, the name of the function confused me. because you can't understand from the name of the function that it returns the entries (from the name i understood that it just checks if the db has entries)

mertyildiran: Is it resolved?

RoyUP9: IMO you should change the name of the function to getDBEntries and move the if len(entries) == 0 check to the tests/other function.

mertyildiran: Done with https://github.com/up9inc/mizu/pull/443/commits/9ff9be9d6804c8bf05a68f977aac17d5bf7e11f5

acceptanceTests/tap_test.go
```diff
@@ -66,21 +66,12 @@ func TestTap(t *testing.T) {
...
-				}
-
-				entries := requestResult.([]interface{})
-				if len(entries) == 0 {
-					return fmt.Errorf("unexpected entries result - Expected more than 0 entries")
-				}
-
-				entry := entries[0].(map[string]interface{})
+				entries := getDBEntries(t, timestamp, entriesCount, 1*time.Second)
+				assertEntriesAtLeast(t, entries, 1)
```
RoyUP9, line 70: you need to stop the function if len(entries) == 0, so you won't continue if you know that you don't have entries

mertyildiran: Doesn't that cause code duplication? How about using `t.Fatalf` instead of `t.Errorf`?

mertyildiran: I guess not because it's tried several times. I'm not sure why there are `retriesExecute` everywhere but anyway. Let me make it return instead.

RoyUP9: I wanted to switch from t.Errorf to t.Fatalf for a long time, i'll do it next spring so let's leave it like that for now.
retriesExecute are called because it can take a few sec for the entries to get inserted into the db, and we don't want to fail the test if they didn't exist on the first check.

mertyildiran: Done with https://github.com/up9inc/mizu/commit/4879346f8fe06c68a8e5b477564d84e6ade8898f

RoyUP9 (approved):

PR #449, opened 2021-11-10
Send the message into this WebSocket connection instead of all

mertyildiran:
Introduced by https://github.com/up9inc/mizu/pull/279

The bug was not affecting any stable releases.


gadotroee:
Just to be sure I understand, this causes messages about errors in specific user to arrive to other users too)
And we still have the logic to send to all clients messages that related to all users (like tappedPods status) right?

gadotroee (approved):

mertyildiran:
> @mertyildiran Just to be sure I understand, this causes messages about errors in specific user to arrive to other users too) And we still have the logic to send to all clients messages that related to all users (like tappedPods status) right?

Yes, exactly.

PR #452, opened 2021-11-11
Optimize UI entry feed performance

mertyildiran:
Fixes the performance issues in the UI by mostly updating the React code.

It makes the entries render only once.

Update the list of entries in case of metadata update instead of entry update in the WebSocket which was making the entries updated too aggressively. The metadata update frequency is determined by the database and it scales with the number of records.

It also upgrades `react-scrollable-feed-virtualized` version from `1.4.3` to `1.4.8`. Which is optimized separately. Also fixes the old issues that were related to scrollbar being snapped to the bottom too aggressively.

Upgrades the Basenine from `v0.2.6` to `v0.2.8` which brings metadata related improvements for performance and accuracy. Also fixes the desynced "displayed" vs. "queried".

Updates the `http` macro to prevent collisions with `grpc` macro.

Prevents query update on serial clicks to the same clickable UI element.

Lastly, it fixes the issue in the filter input textarea by setting the query text before deciding the background color to prevent lags while typing.


gadotroee:
[baseenine-data.zip](https://github.com/up9inc/mizu/files/7533336/baseenine-data.zip

https://user-images.githubusercontent.com/55343099/141671814-fc644f02-ba4b-406c-8275-81664cb4ec8c.mov

)

I tried to use this version and check the input text box a little bit, it is much better. 
I'm trying to understand the difference between displayed and queried - I shared video that show how I get data that say that 1 is missing (maybe there is no bug here but when I didn't have query I saw one missing and when I did "http" and "redis" I also saw 1 is not displayed so I tried to do the opposite filter and got 0 so in my opinion it is bug in the numbers that something is not counted correctly..

In the beginning of the video I do see "all entries" so the numbers somehow correct. 

to make the debug easier (if it helps) I added here the database folder so we can take a look on it too

mertyildiran:
@gadotroee the issues that you have reported are fixed now.

ui/src/helpers/api.js
```diff
@@ -47,7 +50,9 @@ export default class Api {
     validateQuery = async (query) => {
         const form = new FormData();
         form.append('query', query)
-        const response = await this.client.post(`/query/validate`, form);
+        const response = await this.client.post(`/query/validate`, form, {
+            cancelToken: source.token
```
gadotroee, line 54: why do you need this?
I don't see usage in the source (like source.cancel())

ui/src/components/TrafficPage.tsx
```diff
@@ -50,50 +51,55 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
...
 
     const [tappingStatus, setTappingStatus] = useState(null);
 
-    const [disableScrollList, setDisableScrollList] = useState(false);
+    const [isSnappedToBottom, setIsSnappedToBottom] = useState(true);
 
-    const [query, setQueryDefault] = useState("");
+    const [query, setQuery] = useState("");
     const [queryBackgroundColor, setQueryBackgroundColor] = useState("#f5f5f5");
+    const [addition, updateQuery] = useState("");
```
gadotroee, line 65: the convention is to use
```const [param, setParam] = useState("");```
so I think it should be renamed

File: ui/src/components/TrafficPage.tsx [hunk 3]
```diff
@@ -50,50 +51,55 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
...
+    useEffect(() => {
         if (query) {
-            setQuery(`${query} and ${addition}`)
+            setQuery(`${query} and ${addition}`);
         } else {
-            setQuery(addition)
+            setQuery(addition);
         }
-    }
+        // eslint-disable-next-line
```
gadotroee, line 93: can we remove this comment? or is this a good reason to ignore the lint

ui/src/helpers/api.js
mertyildiran: @gadotroee I tried a patch like this:
```diff
diff --git a/ui/src/helpers/api.js b/ui/src/helpers/api.js
index abe936a..0caa8ab 100644
--- a/ui/src/helpers/api.js
+++ b/ui/src/helpers/api.js
@@ -48,6 +48,8 @@ export default class Api {
     }
 
     validateQuery = async (query) => {
+        source.cancel();
+
         const form = new FormData();
         form.append('query', query)
         const response = await this.client.post(`/query/validate`, form, {
```

But it prevents HTTP requests altogether. Right now with `cancelToken`, `axios` cancels the previous requests so we only deal with the latest HTTP request.

ui/src/components/EntryListItem/EntryListItem.tsx
```diff
@@ -38,12 +38,14 @@ interface Rules {
...
     setFocusedEntryId: (id: string) => void;
-    isSelected?: boolean;
     style: object;
     updateQuery: any;
 }
 
-export const EntryItem: React.FC<EntryProps> = ({entry, setFocusedEntryId, isSelected, style, updateQuery}) => {
+export const EntryItem: React.FC<EntryProps> = ({entry, setFocusedEntryId, style, updateQuery}) => {
+
+    const [isSelected, setIsSelected] = useState(false);
```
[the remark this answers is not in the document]
mertyildiran: @gadotroee we don't want to re-render 10k+ `EntryItem` component each time user clicked to one of them. Therefore the state should be tracked internally.

ui/src/components/TrafficPage.tsx
mertyildiran: @gadotroee I know the convention.

That state serves a special purpose. It replaces the legacy `updateQuery` method that's used everywhere:

https://github.com/up9inc/mizu/blob/426953c3500a05825eaf72a1878c9d4412348ed2/ui/src/components/TrafficPage.tsx#L87-L94

Therefore we don't want to change its name.

ui/src/components/TrafficPage.tsx
mertyildiran: @gadotroee that's a linter error that expects developer to include `query` in the list of states that provided on:

https://github.com/up9inc/mizu/blob/426953c3500a05825eaf72a1878c9d4412348ed2/ui/src/components/TrafficPage.tsx#L94

But if we add `query` to that list, the behavior changes so we need to ignore that ESLint check.

ui/src/components/TrafficPage.tsx
```diff
@@ -108,15 +114,18 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
...
-                    }
+                    const entry = message.data;
+                    if (!focusedEntryId) setFocusedEntryId(entry.id.toString());
+                    setEntriesBuffer([
+                        ...entriesBuffer,
+                        <EntryItem
+                            key={entry.id}
+                            entry={entry}
+                            setFocusedEntryId={setFocusedEntryId}
+                            style={{}}
```
gadotroee, line 125: can we remove this empty "style"

mertyildiran: @gadotroee I tried a patch like this:
```diff
diff --git a/ui/src/components/EntryListItem/EntryListItem.tsx b/ui/src/components/EntryListItem/EntryListItem.tsx
index e9346d6..1ff59f6 100644
--- a/ui/src/components/EntryListItem/EntryListItem.tsx
+++ b/ui/src/components/EntryListItem/EntryListItem.tsx
@@ -38,7 +38,7 @@ interface Rules {
 interface EntryProps {
     entry: Entry;
     setFocusedEntryId: (id: string) => void;
-    style: object;
+    style?: object;
     updateQuery: any;
 }
 
diff --git a/ui/src/components/TrafficPage.tsx b/ui/src/components/TrafficPage.tsx
index ae24851..55ec883 100644
--- a/ui/src/components/TrafficPage.tsx
+++ b/ui/src/components/TrafficPage.tsx
@@ -122,7 +122,6 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
                             key={entry.id}
                             entry={entry}
                             setFocusedEntryId={setFocusedEntryId}
-                            style={{}}
                             updateQuery={updateQuery}
                         />
                     ]);
```

But since that object is populated in here:

https://github.com/mertyildiran/react-scrollable-feed/blob/01c8dd7c031aa438edc58fa6dda24d9c8d9ff0ce/src/index.tsx#L271-L272

It throws this error:

```
Uncaught TypeError: Cannot set properties of undefined (setting 'top')
    at t.render (index.tsx:271)
    at $o (react-dom.production.min.js:187)
    at jo (react-dom.production.min.js:186)
    at Vs (react-dom.production.min.js:269)
    at Il (react-dom.production.min.js:250)
    at Nl (react-dom.production.min.js:250)
    at _l (react-dom.production.min.js:250)
    at hl (react-dom.production.min.js:243)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:18)
```

If you look at the diff, it was like that before as well:

https://github.com/up9inc/mizu/pull/452/files#diff-54e1660fae70e6338ceb35b4a59835b8271c56969997f71bdfe01fea7cbb6c28L37

ui/src/helpers/api.js
gadotroee: I didn't see any canceled requests when i use it, 
did you see it is really working? and if not maybe this code can be removed

ui/src/components/TrafficPage.tsx
gadotroee: Ok, I will mark this as resolved

ui/src/components/TrafficPage.tsx
gadotroee: why it can be like before? without state and only function that passed?

```
    const updateQuery = (addition) => {
        if (query) {
            setQuery(`${query} and ${addition}`)
        } else {
            setQuery(addition)
        }
 }
 ```

File: ui/src/helpers/api.js (hunk 1 above)
Comment (mertyildiran) [reply ui/src/helpers/api.js]: @gadotroee OK, I realized my mistake, I'm sorry. Fixed with https://github.com/up9inc/mizu/commit/faaf1c37006d1450cde376cb43fb262d444055e9

File: ui/src/components/TrafficPage.tsx (hunk 3 above)
Comment (mertyildiran) [reply ui/src/components/TrafficPage.tsx:96]: @gadotroee it's changed with https://github.com/up9inc/mizu/commit/803907de93a0671aba373dc01226bccfae81bad1

Without `useEffect`, the state of `query` sticks to its original value (at the time they passed to the components). To be able to get the latest state of `query` I turned it into `useEffect`.

File: ui/src/helpers/api.js [hunk 6]
```diff
@@ -48,11 +49,27 @@ export default class Api {
...
+        this.source = CancelToken.source();
+
         const form = new FormData();
         form.append('query', query)
         const response = await this.client.post(`/query/validate`, form, {
-            cancelToken: source.token
+            cancelToken: this.source.token
+        }).catch(function (thrown) {
+            if (axios.isCancel(thrown)) {
+                console.log('Validate canceled');
```
gadotroee, line 63: I think this can be removed (it is probably good for debugging but not when we are going to merge)

ui/src/helpers/api.js
```diff
@@ -45,9 +49,27 @@ export default class Api {
...
         form.append('query', query)
-        const response = await this.client.post(`/query/validate`, form);
+        const response = await this.client.post(`/query/validate`, form, {
+            cancelToken: this.source.token
+        }).catch(function (thrown) {
+            if (axios.isCancel(thrown)) {
+                console.log('Validate canceled');
+            } else {
+                console.log('Validate error', thrown.message);
+            }
```
gadotroee, line 66:
```suggestion
            if (!axios.isCancel(thrown)) {
                console.error('Validate error', thrown.message);
            }
```

gadotroee: @mertyildiran  what do you think about this?
I don't want to potentially spam when everything is good

mertyildiran: @gadotroee yeah that makes sense. Done with https://github.com/up9inc/mizu/commit/3016cd5bd3fa6dbad08fc239bd3706dc6e9b8206

gadotroee (approved):

PR #456, opened 2021-11-11
changed logger debug mode to log level

RoyUP9:


shared/consts.go
```diff
@@ -13,7 +13,7 @@ const (
 	ConfigFileName                   = "mizu-config.json"
 	GoGCEnvVar                       = "GOGC"
 	DefaultApiServerPort             = 8899
-	DebugModeEnvVar                  = "MIZU_DEBUG"
+	LogLevelEnvVar                   = "MIZU_LOG_LEVEL"
```
IgorGov, line 16: LOG_LEVEL

cli/config/configStruct.go
```diff
@@ -32,6 +33,15 @@ type ConfigStruct struct {
 	KubeConfigPathStr      string                      `yaml:"kube-config-path"`
 	ConfigFilePath         string                      `yaml:"config-path,omitempty" readonly:""`
 	HeadlessMode           bool                        `yaml:"headless" default:"false"`
+	LogLevelStr            string                      `yaml:"server-log-level,omitempty" default:"INFO" readonly:""`
```
IgorGov, line 36: log-level

cli/cmd/root.go
```diff
@@ -23,6 +23,11 @@ Further info is available at https://github.com/up9inc/mizu`,
 		if err := config.InitConfig(cmd); err != nil {
 			logger.Log.Fatal(err)
 		}
+
+		if err := config.Config.Validate(); err != nil {
```
gadotroee, line 27: why not validate when Init?

RoyUP9: because you have to do it after the config is initiated

IgorGov (approved):

PR #459, opened 2021-11-14
Add support to auto discover envoy processes

dvdlevanon:
Service mesh that configured with mtls, prevent mizu from seeing traffic because its encrypted.

The majority of service mesh solutions relaying on an Envoy proxy, that run in a sidecar along side every container in the cluster.

This PR, add the ability to automatically detect Envoy process, if the `--mtls` flag sent to tapper.

It detect envoy process, then inspect their INSTANCE_IP environment variable. If it match one of the cluster ips the tapper get (via the filter authorities), it listen to their private network namespace.

The traffic in the private network namespace is not encrypted, and by this it mitigate  the mtls problem.

The PR force filter authorities to be passed only once instead of multiple times, It does it by moving `tap.SetFilterAuthorities` to `TapOpts` parameter of `StartPassiveTapper`. The motivation for it is because the behavior of the tapper changed, adding and removing Authorities requires extra work. At this point we don't need this ability, hence we prevent it by removing `tap.SetFilterAuthorities`.


tap/source/envoy_discoverer.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const envoyBinary = "/envoy"
+
+var numberRegex = regexp.MustCompile("[0-9]+")
+
+func discoverRelevantEnvoyPids(procfs string, clusterIps []string) ([]string, error) {
+	result := make([]string, 0)
+
+	pids, err := ioutil.ReadDir(procfs)
```
nimrod-up9, line 20: What privileges do we need to access procfs?

tap/source/envoy_discoverer.go
```diff
@@ -0,0 +1,112 @@
...
+
+func readEnvironmentVariable(file string, name string) (string, error) {
+	bytes, err := ioutil.ReadFile(file)
+
+	if err != nil {
+		logger.Log.Warningf("Error reading environment file %v - %v", file, err)
+		return "", err
+	}
+
+	envs := strings.Split(string(bytes), string([]byte{0}))
```
nimrod-up9, line 95: Do we need to take the file encoding into account?

tap/source/envoy_discoverer.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if err != nil {
+		// Debug on purpose - it may happen due to many reasons and we only care
+		//	for it during troubleshooting
+		//
+		logger.Log.Debugf("Unable to read link %v - %v\n", execLink, err)
+		return false
+	}
+
+	if !strings.HasSuffix(exec, envoyBinary) {
```
nimrod-up9, line 60: I think this might be the most fragile part of envoy detection.
It is worth documenting exactly which envoys we support for sure.

agent/main.go
```diff
@@ -94,17 +94,17 @@ func main() {
 			panic("API server address must be provided with --api-server-address when using --tap")
 		}
 
+		hostMode := os.Getenv(shared.HostModeEnvVar) == "1"
+		tapOpts := &tap.TapOpts{HostMode: hostMode}
 		tapTargets := getTapTargets()
 		if tapTargets != nil {
-			tap.SetFilterAuthorities(tapTargets)
-			logger.Log.Infof("Filtering for the following authorities: %v", tap.GetFilterIPs())
+			tapOpts.FilterAuthorities = tapTargets
```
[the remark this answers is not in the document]
dvdlevanon: From the PR description: @nimrod-up9 
The PR force filter authorities to be passed only once instead of multiple times, It does it by moving tap.SetFilterAuthorities to TapOpts parameter of StartPassiveTapper. The motivation for it is because the behavior of the tapper changed, adding and removing Authorities requires extra work. At this point we don't need this ability, hence we prevent it by removing tap.SetFilterAuthorities.

---

Regarding the movement of the condition (the second comment), the tapper package should protect itself from nil (panic). It can't count on the caller to set the right defaults for it, that's why I put the protection in the entrypoint to tapper (StartPassiveTapper)

tap/passive_tapper.go
```diff
@@ -146,17 +150,17 @@ func initializePacketSources() (*source.PacketSourceManager, error) {
 		BpfFilter:   bpffilter,
 	}
 
-	return source.NewPacketSourceManager(*procfs, *pids, *fname, *iface, behaviour)
+	return source.NewPacketSourceManager(*procfs, *pids, *fname, *iface, *mtls, opts.FilterAuthorities, behaviour)
```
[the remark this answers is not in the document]
dvdlevanon: Currently `opts` contain parameters that the user can't configure by itself, the struct passed to `StartPassiveTapper` programmatically. As opposite the command line flags are outside the `opts` struct.

I'm all in to merge them all together and creating a unified TapperConfig. But its better to do it in its own dedicated refactor branch

tap/settings.go
```diff
@@ -18,24 +18,6 @@ const (
...
-
-func SetFilterAuthorities(ipAddresses []string) {
-	gSettings.filterAuthorities = ipAddresses
-}
-
-func GetFilterIPs() []string {
-	addresses := make([]string, len(gSettings.filterAuthorities))
-	copy(addresses, gSettings.filterAuthorities)
-	return addresses
-}
```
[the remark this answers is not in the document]
dvdlevanon: I thought up9 is version locked + it already break by tapping changes. @nimrod-up9 Am I wrong?

tap/source/envoy_discoverer.go
dvdlevanon: With `--privileged` container we can access all.
With non-privileged we can access only what the user of the tapper container can, and that's depend on various parameters.

At this point we can start with this, if it doesn't have permission nothing happen but an error in the log.

Overtime we may find out edge cases where we can't access it, and fix it on demand. In any case, its a common thing for monitoring containers to map `/proc` of host to container

tap/source/envoy_discoverer.go
dvdlevanon: We are looking for english and number letters only, so I don't think we should take it into account

tap/source/envoy_discoverer.go
dvdlevanon: Agree, one of my immediate tasks is to put Envoy.md in docs

tap/source/envoy_discoverer.go
mertyildiran: @nimrod-up9 I think the privileges provided through

https://github.com/up9inc/mizu/blob/4c04ae39c975845e409a074b8a9cf6663518a4da/shared/kubernetes/provider.go#L715-L721

which is added by https://github.com/up9inc/mizu/pull/418 and it's `readOnlyRootFilesystem`. That seems OK.

tap/source/envoy_discoverer.go
```diff
@@ -0,0 +1,112 @@
...
+		return false
+	}
+
+	environmentFile := fmt.Sprintf("%v/%v/environ", procfs, pid)
+	clusterIp, err := readEnvironmentVariable(environmentFile, "INSTANCE_IP")
```
mertyildiran, line 65: Use [hashicorp/go-envparse](https://github.com/hashicorp/go-envparse). `readEnvironmentVariable` is not a valid parser and therefore error-prone.

tap/source/envoy_discoverer.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, value := range clusterIps {
+		if value == clusterIp {
+			return true
+		}
+	}
+
+	return false
+}
+
+func readEnvironmentVariable(file string, name string) (string, error) {
```
mertyildiran, line 87: Remove `readEnvironmentVariable` method.

mertyildiran (changes requested):

tap/source/envoy_discoverer.go
dvdlevanon: The `/proc/PID/environ` file is not split by lines, but null terminated strings. I don't think it would fit.

dvdlevanon: Moreover, this is not a regular file, no comments or empty lines allowed in there. Its written by the kernel and its format is simpler than a regular environment file

mertyildiran: @dvdlevanon `strings /proc/*/environ` splits the null terminated strings. Such that after it's splitted, https://github.com/hashicorp/go-envparse/blob/master/envparse.go#L102 can be used (though it's not exported). But anyway I guess the value cannot be so complicated therefore `strings.Split` does not fail.

tap/settings.go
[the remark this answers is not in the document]
nimrod-up9: You are right, it is version locked.
I'm just think that it is easier if we don't break functionality where we can.
On second thought you can ignore this comment. We will handle the problem when it arises.

nimrod-up9 (approved):

nimrod-up9 (approved):

PR #461, opened 2021-11-14
changing the mizu UI image

alongir:


gadotroee:
Anyway the image is not the current UI (after @mertyildiran DB change) so it will be in separate PR that will fix the image to be the new one

PR #464, opened 2021-11-15
Return `404` instead of `500` if the entry could not be found and display a toast message

mertyildiran:
`404` happens if the entry is removed because of database size limit (200MB by default). So we display a toast message to inform the user about inexistence of the entry.

Screenshot from 2021-11-15 22-39-40

Screenshot from 2021-11-15 22-40-20

### DIFF
--- a/agent/pkg/controllers/entries_controller.go
+++ b/agent/pkg/controllers/entries_controller.go
@@ -25,7 +25,12 @@ func Error(c *gin.Context, err error) bool {
 	if err != nil {
 		logger.Log.Errorf("Error getting entry: %v", err)
 		c.Error(err)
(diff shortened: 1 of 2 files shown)


gadotroee:
I thought about it with little help from the team (@lirazyehezkel  and @nimrod-up9) and I have suggestions, let me know what do you think about them:

1. Ideally - User need to know that entries are not available (deleted due to storage limit) and we thought it might be possible to update FE with list of entries that we are deleting (in the web socket connection, same way as we know about new entries) - then we can potentially change the entry in the left side to be greyed out or not clickable and explain on hover or something that you can see this one.

2. if it is not possible and requires more rendering (which might be the reason not to do it) - we can show the error in the right section and not as toast - so I would think of something like the attached image 


. 

Let me know what do you think about it and see if we can decide on implementation.

gadotroee (approved):

PR #473, opened 2021-11-16
TRA-3903 fix daemon mode in permission restricted configs

RamiBerm:
…re files...


cli/cmd/tapRunner.go
```diff
@@ -387,6 +373,25 @@ func createMizuApiServerDeployment(ctx context.Context, kubernetesProvider *kube
...
+			logger.Log.Warningf(uiUtils.Yellow, "An error has occured while creating a persistent volume claim for mizu, this means mizu data will be lost on mizu-api-server pod restart")
+			logger.Log.Debugf("error creating persistent volume claim: %v", err)
+		} else {
+			return true
+		}
+	} else {
+		logger.Log.Warningf(uiUtils.Yellow, "Could not find default volume provider in this cluster, this will mean that mizu's data will be lost on pod restart")
+	}
+	return false
+}
```
nimrod-up9, line 393: Simplify.
Less nesting and "quit early" are easier to follow.

```go
func TryToCreatePersistentVolumeClaim(ctx context.Context, kubernetesProvider *kubernetes.Provider) bool {
    if isDefaultStorageClassAvailable, err := kubernetesProvider.IsDefaultStorageProviderAvailable(ctx); err != nil {
        logger.Log.Debugf(...)
        return false
    } else if ~isDefaultStorageClassAvailable {
        logger.Log.Warningf(...)
        return false
    }

    if _, err = kubernetesProvider.CreatePersistentVolumeClaim(...); err != nil {
        logger.Log.Warningf(...)
        logger.Log.Debugf(...)
        return false
    }

return true
```

examples/roles/permissions-ns-daemon.yaml
```diff
@@ -32,9 +32,9 @@ rules:
 - apiGroups: ["rbac.authorization.k8s.io"]
   resources: ["rolebindings"]
   verbs: ["get", "create", "delete"]
-- apiGroups: ["apps", "extensions"]
+- apiGroups: ["apps", "extensions", ""]
   resources: ["pods"]
-  verbs: ["get", "list", "watch"]
+  verbs: ["get", "list", "watch", "create"]
```
nimrod-up9, line 37: Remove create if it is only needed for non-daemon.

gadotroee (approved):

PR #477, opened 2021-11-16
Sync entries in batches just as before (using `uploadIntervalSec` parameter)

mertyildiran:
`uploadIntervalSec` parameter became unused with https://github.com/up9inc/mizu/pull/279 and this PR fixes it to its original state.


agent/pkg/up9/main.go
```diff
@@ -227,6 +227,10 @@ func syncEntriesImpl(token string, model string, envPrefix string, uploadInterva
 		connection.Close()
 	}()
 
+	lastTimeSynced := time.Now().Add(time.Duration(-uploadIntervalSec) * time.Second)
```
RoyUP9, line 230: Can be changed to default time

```suggestion
	lastTimeSynced := time.Time{}
```

mertyildiran: Done with https://github.com/up9inc/mizu/commit/174abf89022c99e0c47673d1d6bdd510b00118c1

RoyUP9 (approved):

PR #481, opened 2021-11-17
Prevent the crash on client-side in case of `text` being undefined in `FancyTextDisplay`

mertyildiran:
Introduced by https://github.com/up9inc/mizu/pull/478 not present in a stable release.

Prevents:

Screenshot from 2021-11-17 17-49-49

### DIFF
--- a/ui/src/components/UI/FancyTextDisplay.tsx
+++ b/ui/src/components/UI/FancyTextDisplay.tsx
@@ -17,6 +17,7 @@ interface Props {
 const FancyTextDisplay: React.FC<Props> = ({text, className, isPossibleToCopy = true, applyTextEllipsis = true, flipped = false, useTooltip= false, displayIconOnMouseOver = false, buttonOnly = false}) => {
     const [showCopiedNotification, setCopied] = useState(false);
     const [showTooltip, setShowTooltip] = useState(false);
+    if (text == null) text = '';
 
     const onCopy = () => {
         setCopied(true)
--- a/ui/src/components/UI/FancyTextDisplay.tsx
+++ b/ui/src/components/UI/FancyTextDisplay.tsx
@@ -17,7 +17,7 @@ interface Props {
 const FancyTextDisplay: React.FC<Props> = ({text, className, isPossibleToCopy = true, applyTextEllipsis = true, flipped = false, useTooltip= false, displayIconOnMouseOver = false, buttonOnly = false}) => {
     const [showCopiedNotification, setCopied] = useState(false);
     const [showTooltip, setShowTooltip] = useState(false);
-    if (text == null) text = '';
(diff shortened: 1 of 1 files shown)


ui/src/components/UI/FancyTextDisplay.tsx
```diff
@@ -17,6 +17,7 @@ interface Props {
 const FancyTextDisplay: React.FC<Props> = ({text, className, isPossibleToCopy = true, applyTextEllipsis = true, flipped = false, useTooltip= false, displayIconOnMouseOver = false, buttonOnly = false}) => {
     const [showCopiedNotification, setCopied] = useState(false);
     const [showTooltip, setShowTooltip] = useState(false);
+    if (text == null) text = '';
```
gadotroee, line 20: @mertyildiran 

It can't be ```text = text || ''``` because it will also change it in case of "0" or "false"?

mertyildiran: @gadotroee if we're talking about `"0"` or `"false"` string literals, then non-empty string literals are evaluated to `true` as far as I know in JavaScript. But if talk about `0` and `false` integer and boolean literals, I guess the server returns only a string literal or `nil` but I'm not sure. Should we handle all possible cases?

mertyildiran: @gadotroee how about now? Changed it to `text = String(text);`:

Screenshot from 2021-11-17 18-07-38

[`String()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String) is pretty much fail-safe.

mertyildiran: [The typing there](https://github.com/up9inc/mizu/blob/fix/fancy-text-display/ui/src/components/UI/FancyTextDisplay.tsx#L7) `string` or `number` but `String()` pretty much guarantees to be a string without a failure.

gadotroee: I thinks it might be better. 
taking another look and will approve

gadotroee: I see that String(null) returns "null" which is probably not the right thing.. 
but I'm not sure in which cases it is happening

mertyildiran: @gadotroee do you see a `null` in the UI or you just looked up `String(null)` returns `"null"` and do you think it's not correct? If it's the latter, then why not displaying `null` if it's a `null`? Although I saw only `undefined` (in the screenshot **Encoding** field) and that seems also OK. Since the encoding cannot be determined hence undefined.

gadotroee (approved):

PR #487, opened 2021-11-19
Disable IPv4 defragmentation and support IPv6

mertyildiran:
Removes the extra negation on `nodefrag` flag's value. (introduced by https://github.com/up9inc/mizu/pull/396)

In case of IPv6, layer type becomes `LayerTypeIPv6` hence;

https://github.com/up9inc/mizu/blob/7e0bb766fb66bb5f93e0fa06d30e439769209222/tap/source/tcp_packet_source.go#L125

returns `nil`. Therefore the packet is ignored.

It broke [**Local Machine Development Workflow**](https://github.com/up9inc/mizu/wiki/Development-Workflows#local-machine-development-workflow) because of breaking IPv6 support.

It disables IPv4 defragmentation and adds a check to ignore layers other than `LayerTypeIPv4`.

### PCAP Testcases

[http.cap](https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=http.cap) fragmented using:

```
IPFragUtil http.cap -o http_fragmented_128_bytes.cap -s 128
```

[http_fragmented_128_bytes.cap](https://drive.google.com/file/d/1p9wFkyzKpsKnaAYfMPBUy9pKWOu80ZsN/view?usp=sharing) fragmented IPv4 HTTP

[v6-http.cap](https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=v6-http.cap) IPv6 HTTP


nimrod-up9 (approved):

shared/kubernetes/provider.go
```diff
@@ -634,7 +635,6 @@ func (provider *Provider) ApplyMizuTapperDaemonSet(ctx context.Context, namespac
 		"-i", "any",
 		"--tap",
 		"--api-server-address", fmt.Sprintf("ws://%s/wsTapper", apiServerPodIp),
-		"--nodefrag",
```
nimrod-up9, line 637: We used to have an issue where defrag causes fatal errors.
That's why we disable it.
On the other hand, as far as I'm aware, IP fragmentation is rare these days.
Where did you come across IP fragmentation?

mertyildiran: The `--nodefrag` was added in https://github.com/up9inc/mizu/pull/224, thinking it was causing issues for other protocols without realizing it was actually an IPv6 issue.

Later on that flag is disabled with [this change](https://github.com/up9inc/mizu/pull/396/files#diff-caef9971ec56b1f572211e0b9bc0f0fe462ec242a6e58e50a294d629b8685609R119) and made it to `0.19.0` stable release. Therefore currently IPv4 defragmentation is enabled in `develop` and `main` branches right now.

Yes, it's correct that IP fragmentation occur only if the TCP packet size is bigger than 1500 MTU by today's standard and it's a rare event.

Enabling IPv4 defragmentation has a slight performance disadvantage. So as you have requested, I've disabled the IPv4 defragmentation with https://github.com/up9inc/mizu/commit/e394d8a96df8b37a16783d1e2f2541b08c55de64

nimrod-up9 (approved):

PR #490, opened 2021-11-20
Add `golangci-lint` to CI

mertyildiran:


IgorGov:
we can close this PR the go linter introduced here:
https://github.com/up9inc/mizu/pull/734

PR #493, opened 2021-11-21
Warn pods not starting

nimrod-up9:
Print warning event related to mizu k8s resources.
In non-daemon print to CLI. In Daemon print to API-Server logs.

https://user-images.githubusercontent.com/59927337/142762097-b7074e09-e1bc-40b0-b01d-54c8915b9353.mov


gadotroee (approved):

cli/cmd/tapRunner.go
```diff
@@ -736,6 +739,89 @@ func watchTapperPod(ctx context.Context, kubernetesProvider *kubernetes.Provider
...
+
+	for {
+		select {
+		case wEvent, ok := <-added:
+			if !ok {
+				added = nil
+				continue
+			}
+
+			event, err := wEvent.ToEvent()
```
RoyUP9, line 758: maybe extract to function, i see the same code used 3 times

cli/cmd/tapRunner.go
```diff
@@ -736,6 +739,89 @@ func watchTapperPod(ctx context.Context, kubernetesProvider *kubernetes.Provider
...
+			if event.Type == core.EventTypeWarning {
+				logger.Log.Warningf(uiUtils.Warning, fmt.Sprintf("Resource %s in state %s - %s", event.Regarding.Name, event.Reason, event.Note))
+			}
+		case err, ok := <-errorChan:
+			if !ok {
+				errorChan = nil
+				continue
+			}
+
+			logger.Log.Errorf("[Error] Error in mizu tapper watch, err: %v", err)
```
RoyUP9, line 815: i think this should be error in mizu events or something like that

shared/kubernetes/watchEvent.go
```diff
@@ -16,3 +18,12 @@ func (we *WatchEvent) ToPod() (*corev1.Pod, error) {
 
 	return pod, nil
 }
+
+func (we *WatchEvent) ToEvent() (*eventsv1.Event, error) {
+	pod, ok := we.Object.(*eventsv1.Event)
```
RoyUP9, line 23: rename pod to event

shared/kubernetes/watchEvent.go
```diff
@@ -16,3 +18,12 @@ func (we *WatchEvent) ToPod() (*corev1.Pod, error) {
...
+	if !ok {
+		return nil, fmt.Errorf("Invalid object type on pod event stream")
```
RoyUP9, line 25: maybe change to something that is not similar to the error we get on ToPod

shared/kubernetes/watchEvent.go
nimrod-up9: Fixed

shared/kubernetes/watchEvent.go
nimrod-up9: done

cli/cmd/tapRunner.go
nimrod-up9: I'm thinking of removing the splitting of the event channel to added, removed, modified and error.
Instead, run a switch case when handling wEvent.
What do you think?

cli/cmd/tapRunner.go
nimrod-up9: done

cli/cmd/tapRunner.go
RoyUP9: i think it's a good idea, but maybe we should do it in separate PR

agent/main.go
```diff
@@ -482,3 +485,86 @@ func startMizuTapperSyncer(ctx context.Context) (*kubernetes.MizuTapperSyncer, e
 
 	return tapperSyncer, nil
 }
+
+func watchMizuEvents(ctx context.Context, kubernetesProvider *kubernetes.Provider, cancel context.CancelFunc) {
```
RoyUP9, line 489: isn't it the exact same function that we used for cli?

RoyUP9 (dismissed):

RoyUP9 (approved):

RoyUP9 (approved):

PR #492, opened 2021-11-21
Make the `gRPC` and `HTTP/2` distinction

mertyildiran:
Screenshot from 2021-11-22 13-22-49

Screenshot from 2021-11-22 13-22-51

Screenshot from 2021-11-22 13-23-00

Screenshot from 2021-11-22 13-23-02

PR #496, opened 2021-11-22
Remove `SetHostname` method in HTTP extension

mertyildiran:
Introduced by https://github.com/up9inc/mizu/pull/492 (not present in a stable release)

### Before:

Screenshot from 2021-11-22 18-28-58

### After:

Screenshot from 2021-11-22 18-53-54

### DIFF
--- a/tap/extensions/http/main.go
+++ b/tap/extensions/http/main.go
@@ -124,15 +124,6 @@ func (d dissecting) Dissect(b *bufio.Reader, isClient bool, tcpID *api.TcpID, co
 	return nil
 }
 
-func SetHostname(address, newHostname string) string {
-	replacedUrl, err := url.Parse(address)
-	if err != nil {
-		return address
-	}
-	replacedUrl.Host = newHostname
-	return replacedUrl.String()
-}
-
 func (d dissecting) Analyze(item *api.OutputChannelItem, resolvedSource string, resolvedDestination string) *api.MizuEntry {
 	var host, authority, path, service string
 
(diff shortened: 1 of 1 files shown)


tap/extensions/http/main.go
```diff
@@ -189,9 +180,9 @@ func (d dissecting) Analyze(item *api.OutputChannelItem, resolvedSource string,
 	reqDetails["queryString"] = mapSliceRebuildAsMap(reqDetails["_queryString"].([]interface{}))
 
 	if resolvedDestination != "" {
-		service = SetHostname(service, resolvedDestination)
+		service = resolvedDestination
```
gadotroee, line 183: @mertyildiran 

Not related to your change explicitly but it feels weird that same field (service) got the value from "resolvedSource" or "resolvedDestination" - those should be opposite values (like if request came from A -> B, the source will be A and the destination will be B, in this simple example request the service should be probably "B" because the request issued to him)..
Maybe I miss something or is it really mistake?

mertyildiran: @gadotroee the logic is like that for ages. Actually saying a traffic log belongs to a service is in accurate. Because that's a binary event that involves two services. But the design was like that since the very beginning.

We try to show a name instead of an IP with that `service` field, if destination is resolved we show destination. If destination is not resolved, we fallback to source. That's why the logic is like that.

gadotroee: I understand you didn't change the behavior in this commit so I'm approving it because this fix is good but I do think that original behavior (maybe a long time ago) was to show the "SERVICE" (which is the target part) if it is not resolved we should (IMO) show the ip because it is more important to be consistent on what we see in this place.

gadotroee (approved):

PR #501, opened 2021-11-22
Revert the select entry behavior into its original state (before #452) TRA-3983

mertyildiran:
https://user-images.githubusercontent.com/2502080/142930879-e99b2cd0-b8d8-4365-b606-b08741597cab.mp4

### Update: Not Racing Anymore

Race condition is fixed with https://github.com/up9inc/mizu/commit/dcfa056c69d95f294f20935277e07bc29dea8470

https://user-images.githubusercontent.com/2502080/143896184-b85f21df-a278-4c1a-a5f8-91e5e4d65531.mp4

### DIFF
--- a/ui/src/components/EntryDetailed.tsx
+++ b/ui/src/components/EntryDetailed.tsx
@@ -69,6 +69,7 @@ const EntrySummary: React.FC<any> = ({data, updateQuery}) => {
     return <EntryItem
         key={entry.id}
         entry={entry}
+        focusedEntryId={null}
         setFocusedEntryId={null}
         style={{}}
         updateQuery={updateQuery}
--- a/ui/src/components/EntryDetailed.tsx
+++ b/ui/src/components/EntryDetailed.tsx
@@ -72,7 +72,6 @@ const EntrySummary: React.FC<any> = ({data, updateQuery}) => {
         setFocusedEntryId={null}
         style={{}}
         updateQuery={updateQuery}
-        forceSelect={false}
         headingMode={true}
     />;
 };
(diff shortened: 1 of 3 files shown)


gadotroee:
From what I see in the attached video, and I will test it myself too looks like it is ok when you do not have high load (when there are many entries in the screen already but smaller number of entries arrive - it looks good, you do have the selection context and it is not seems to be that slow..

mertyildiran:
@gadotroee my bad that I didn't write to hook properly so it raced. Race condition is fixed with https://github.com/up9inc/mizu/commit/dcfa056c69d95f294f20935277e07bc29dea8470

gadotroee:
> @gadotroee my bad that I didn't write to hook properly so it raced. Race condition is fixed with https://github.com/up9inc/mizu/commit/dcfa056c69d95f294f20935277e07bc29dea8470

I will review it probably tomorrow morning, maybe I will have time to do it before.

gadotroee:
I can say that code-wise it looks ok. I want to run it one time to get the feeling of this so I will approve after simple sanity check..

mertyildiran:
@gadotroee I realized that it's racing again but in a different way this time. Such that this line:

https://github.com/up9inc/mizu/blob/dcfa056c69d95f294f20935277e07bc29dea8470/ui/src/components/TrafficPage.tsx#L203

races with this line:

https://github.com/up9inc/mizu/blob/dcfa056c69d95f294f20935277e07bc29dea8470/ui/src/components/TrafficPage.tsx#L127-L138

when the stream is aggressive. Therefore if you click to the items several times while an aggressive stream is going on, a few entries that are streamed through WebSocket are lost and the length of `entriesBuffer` becomes smaller than the total number of streamed items:

Screenshot from 2021-11-30 07-22-20

mertyildiran:
Closing in favor of https://github.com/up9inc/mizu/pull/513

PR #502, opened 2021-11-23
Revert the select entry behavior into its original state BAD PERFORMANCE! (before #452) [TRA-3983 alternative 4]

mertyildiran:
https://user-images.githubusercontent.com/2502080/142975799-2a774e31-d136-4d87-9387-0b8342a1e2d9.mp4


mertyildiran:
The performance is almost equal to the state in `git reset --hard 2582b7a65cb2a6787c243a342275ef03e627de4c`

PR #508, opened 2021-11-25
Handle unexpected socket close and replace the default `rlimit(100)` filter with `leftOff(-1)` filter

mertyildiran:
### Before

https://user-images.githubusercontent.com/2502080/143348245-8bd3e08f-7cdd-4511-9789-93865fd7b0a6.mp4

### After

https://user-images.githubusercontent.com/2502080/143349187-d357ea44-4362-4402-8e57-29f92b1933bf.mp4


ui/src/components/TrafficPage.tsx
```diff
@@ -100,16 +101,30 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
...
             setConnection(ConnectionStatus.Connected);
+            ws.current.send(query);
+        }
+        ws.current.onclose = () => {
+            setConnection(ConnectionStatus.Closed);
+        }
+        ws.current.onerror = (event) => {
+            console.error("WebSocket error:", event);
+            if (query) {
+                openWebSocket(`(${query}) and leftOff(${leftOff})`, true);
```
gadotroee, line 123: do we want to support leftOff that will be added by user? 
because then you can get 
query that has "http and leftOff(-1) and leftOff(600)" which is weird.
but maybe I misunderstand and leftOff is internal parameter that should not be specified by user

ui/src/components/TrafficPage.tsx
```diff
@@ -100,16 +101,30 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
 
     const listEntry = useRef(null);
 
-    const openWebSocket = (query) => {
-        setFocusedEntryId(null);
-        setEntries([]);
-        setEntriesBuffer([]);
+    const openWebSocket = (query: string, dontClear: boolean) => {
+        if (!dontClear) {
```
gadotroee, line 105: maybe another name is better, it was hard for e to understand (especially when the name is negative name).. 
I prefer positive names which makes it easier when reading (no need to think about `!dont*` expressions)
about the name itself - maybe `resetEntriesBuffer` or something like this is better?

ui/src/components/TrafficPage.tsx
mertyildiran: @gadotroee it's aimed to be used internally. There is no accurate information in the UI for user to provide a correct value also. They can do inspect element to get the ids though.

gadotroee (commented):
2 small questions added

ui/src/components/TrafficPage.tsx
mertyildiran: @gadotroee makes sense. Done with https://github.com/up9inc/mizu/commit/19157139d9c901f18755e132f1083f99a8c0fd6b

gadotroee (approved):

PR #510, opened 2021-11-25
Add HTTP2 Over Cleartext (H2C) support

mertyildiran:
Screenshot from 2021-11-25 07-20-57

Screenshot from 2021-11-25 07-21-00

Screenshot from 2021-11-25 07-21-12

Screenshot from 2021-11-25 07-21-18

### DIFF
--- a/tap/extensions/http/handlers.go
+++ b/tap/extensions/http/handlers.go
@@ -24,7 +24,7 @@ func filterAndEmit(item *api.OutputChannelItem, emitter api.Emitter, options *ap
 	emitter.Emit(item)
 }
 
-func handleHTTP2Stream(http2Assembler *Http2Assembler, tcpID *api.TcpID, superTimer *api.SuperTimer, emitter api.Emitter, options *api.TrafficFilteringOptions, isClient bool) error {
+func handleHTTP2Stream(http2Assembler *Http2Assembler, tcpID *api.TcpID, superTimer *api.SuperTimer, emitter api.Emitter, options *api.TrafficFilteringOptions) error {
 	streamID, messageHTTP1, isGrpc, err := http2Assembler.readMessage()
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


tap/extensions/http/matcher.go
```diff
@@ -92,6 +92,6 @@ func splitIdent(ident string) []string {
 }
 
 func genKey(split []string) string {
-	key := fmt.Sprintf("%s:%s->%s:%s,%s", split[0], split[2], split[1], split[3], split[4])
+	key := fmt.Sprintf("%s:%s->%s:%s,%s%s", split[0], split[2], split[1], split[3], split[4], split[5])
```
IgorGov, line 95: no "," needed?
```suggestion
	key := fmt.Sprintf("%s:%s->%s:%s,%s,%s", split[0], split[2], split[1], split[3], split[4], split[5])
```

IgorGov (approved):

tap/extensions/http/handlers.go
```diff
@@ -34,12 +35,13 @@ func handleHTTP2Stream(http2Assembler *Http2Assembler, tcpID *api.TcpID, superTi
 	switch messageHTTP1 := messageHTTP1.(type) {
```
IgorGov, line 35: messageHTTP1?

tap/extensions/http/matcher.go
mertyildiran: A delimiter is not really required since this is a unique identifier.

tap/extensions/http/handlers.go
mertyildiran: It's named `messageHTTP1` because [`http2_assembler.go` returns HTTP 1.x data structure](https://github.com/up9inc/mizu/blob/4007b842573832ebe5f3ba1fe988f7302900fe54/tap/extensions/http/http2_assembler.go#L160) from `net/http` package. (`http.Request` and `http.Response`) It returns HTTP 1.x data structure because everything related to HTTP designed around that data structure in HTTP extension or other parts of the software related to HAR conversion etc.

PR #513, opened 2021-11-30
Fix the selected entry behavior by propagating the `focusedEntryId` through WebSocket (before #452) TRA-3983

mertyildiran:
Introduces a slight delay in case of network bottleneck:

https://user-images.githubusercontent.com/2502080/144033464-a9944bbd-b051-4720-86fd-63c3a8827bee.mp4

https://user-images.githubusercontent.com/2502080/144034777-59e128ee-a882-4908-8079-815b5a64cfc0.mp4

No racing, no incorrect behavior.

### DIFF
--- a/agent/pkg/api/socket_routes.go
+++ b/agent/pkg/api/socket_routes.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"mizuserver/pkg/models"
 	"net/http"
+	"strconv"
 	"sync"
 	"time"
 
@@ -94,6 +95,8 @@ func websocketHandler(w http.ResponseWriter, r *http.Request, eventHandlers Even
 	startTimeBytes, _ := models.CreateWebsocketStartTimeMessage(startTime)
 	SendToSocket(socketId, startTimeBytes)
 
+	queryRecieved := false
+
 	for {
 		_, msg, err := ws.ReadMessage()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


agent/pkg/api/socket_routes.go
```diff
@@ -94,72 +95,84 @@ func websocketHandler(w http.ResponseWriter, r *http.Request, eventHandlers Even
...
-			if err != nil {
-				toastBytes, _ := models.CreateWebsocketToastMessage(&models.ToastMessage{
-					Type:      "error",
-					AutoClose: 5000,
-					Text:      fmt.Sprintf("Syntax error: %s", err.Error()),
-				})
-				SendToSocket(socketId, toastBytes)
-				break
-			}
+		if !queryRecieved {
```
gadotroee, line 107: @mertyildiran 
I'm probably miss something - and I'm not sure I understand the "queryRecieved" here.
can you help me understand what is the idea here?

gadotroee: If I understand correctly, this for loop is happening when opening connection (this happen every time filter is applied), and the loop will start send information to the browser via the web socket connection.
so this change allowing the client (UI) to send message to the server when entry is changed and (
```

        if (ws.current.readyState === WebSocket.OPEN) {
            ws.current.send(focusedEntryId);
        }
```

and if I understand correctly the server will just return message with the same data?

mertyildiran: @gadotroee kind of yes.

The first ever message from client-to-server is the query like `request.path == "..."` The subsequent messages can only be the `focusedEntryId` values and yes we simply mirror the value back to the client. It's required to resolve the racing issues (finally I understand this time). So;

This from this line:

https://github.com/up9inc/mizu/blob/a62fba4cb63fdb29873633f765115c1ff777b014/ui/src/components/TrafficPage.tsx#L205-L207

value is passed to this block:

https://github.com/up9inc/mizu/blob/a62fba4cb63fdb29873633f765115c1ff777b014/ui/src/components/TrafficPage.tsx#L170-L177

through WebSocket. To prevent race condition that's caused by this block:

https://github.com/up9inc/mizu/blob/a62fba4cb63fdb29873633f765115c1ff777b014/ui/src/components/TrafficPage.tsx#L127-L138

gadotroee: @mertyildiran 
Ok, I understand it now and it should work - I Just think if it is good practice to assume that all messages after query passed are only messages of focusedEntryId (it might be Ok because before we didn't use the client to server connection after making the query and it was used only for server to client messaged (different types including data / metadata (tapped pods) / maybe another type that  I forget about).

Also I think it will be different after the pagination change.. 
so I will approve this to get "working" develop.

gadotroee (approved):

PR #514, opened 2021-11-30
Replace `privileged` with specific CAPABILITIES requests 

dvdlevanon:
Drop `--privileged` permissions and most capabilities from Mizu Pod, use certain Linux Capabilities instead.

It would help users with strict PodSecurityPolicy (TRA-3991, TRA-3965)

The capabilities needed for regular tapping is:
`NET_RAW` - for libpcap

Additional capabilities needed for istio tapping:
- `SYS_ADMIN` - for reading /proc/PID/net/ns
- `SYS_PTRACE` - for setting netns to other process
- `DAC_OVERRIDE` - for reading /proc/PID/environ


shared/kubernetes/provider.go
```diff
@@ -651,7 +651,17 @@ func (provider *Provider) ApplyMizuTapperDaemonSet(ctx context.Context, namespac
 	agentContainer.WithName(tapperPodName)
 	agentContainer.WithImage(podImage)
 	agentContainer.WithImagePullPolicy(imagePullPolicy)
-	agentContainer.WithSecurityContext(applyconfcore.SecurityContext().WithPrivileged(true))
+
+	caps := applyconfcore.Capabilities().WithDrop("ALL").WithAdd("NET_RAW")
```
mertyildiran, line 655: I think we need to add `NET_ADMIN` as well because of the `-promisc` flag.

> `NET_RAW` permits raw access to an interface for capturing directly off the wire (by default enabled in Kubernetes)
> `NET_ADMIN` allows us to set an interface to promiscuous mode (required by the `-promisc` flag in the tapper)
> See `CAP_NET_` prefixed Linux capabilities for more https://man7.org/linux/man-pages/man7/capabilities.7.html

Related PR https://github.com/up9inc/mizu/pull/436

shared/kubernetes/provider.go
```diff
@@ -651,7 +651,17 @@ func (provider *Provider) ApplyMizuTapperDaemonSet(ctx context.Context, namespac
...
+	
+	if istio {
+		caps = caps.WithAdd("SYS_ADMIN") // for reading /proc/PID/net/ns
+		caps = caps.WithAdd("SYS_PTRACE") // for setting netns to other process
+		caps = caps.WithAdd("DAC_OVERRIDE") // for reading /proc/PID/environ
```
mertyildiran, line 660: Do we really need `SYS_ADMIN` and `DAC_OVERRIDE` since they come from a mounted directory?

dvdlevanon: Yes, tried without them

shared/kubernetes/provider.go
dvdlevanon: It's interesting, I wonder how come it worked without it, maybe it has something to do with kernel versions (tested with `5.15.4`).

I'll add the NET_ADMIN because:
- Istio need it as well
- To be on the safe side

mertyildiran: @dvdlevanon great!

shared/kubernetes/provider.go
mertyildiran: @dvdlevanon OK.

IgorGov (approved):

PR #515, opened 2021-11-30
Bring back `GetEntries` HTTP endpoint

mertyildiran:
### Example request:

```bash
$ curl -X GET \
  'http://localhost:8899/entries?leftOff=20&direction=-1&query=&limit=5&timeoutMs=5000' \
  -H 'cache-control: no-cache' \
  -H 'postman-token: 87704b8a-aa7f-6ff4-8a01-863a2b00b901' -L
```

### Example response:

```json
{
    "data": [
        {
            "contractStatus": 0,
            "destinationIp": "172.17.0.9",
            "destinationPort": "8079",
            "id": 20,
            "latency": 1,
            "method": "GET",
            "protocol": {
                "abbr": "HTTP",
                "backgroundColor": "#205cf5",
                "fontSize": 12,
                "foregroundColor": "#ffffff",
                "longName": "Hypertext Transfer Protocol -- HTTP/1.1",
                "macro": "http",
                "name": "http",
                "ports": [
                    "80",
                    "443",
                    "8080"
                ],
                "priority": 0,
                "referenceLink": "https://datatracker.ietf.org/doc/html/rfc2616",
                "version": "1.1"
            },
            "requestSenderIp": "172.17.0.1",
            "rules": {},
            "service": "front-end.sock-shop",
            "sourceIp": "172.17.0.1",
            "sourcePort": "36878",
            "statusCode": 200,
            "summary": "/",
            "timestamp": 1638305660498,
            "url": "front-end.sock-shop/"
        },
        {
            "contractStatus": 0,
            "destinationIp": "172.17.0.9",
            "destinationPort": "8079",
            "id": 19,
            "latency": 1,
            "method": "GET",
            "protocol": {
                "abbr": "HTTP",
                "backgroundColor": "#205cf5",
                "fontSize": 12,
                "foregroundColor": "#ffffff",
                "longName": "Hypertext Transfer Protocol -- HTTP/1.1",
                "macro": "http",
                "name": "http",
                "ports": [
                    "80",
                    "443",
                    "8080"
                ],
                "priority": 0,
                "referenceLink": "https://datatracker.ietf.org/doc/html/rfc2616",
                "version": "1.1"
            },
            "requestSenderIp": "172.17.0.1",
            "rules": {},
            "service": "front-end.sock-shop",
            "sourceIp": "172.17.0.1",
            "sourcePort": "36876",
            "statusCode": 200,
            "summary": "/",
            "timestamp": 1638305660498,
            "url": "front-end.sock-shop/"
        },
        {
            "contractStatus": 0,
            "destinationIp": "172.17.0.6",
            "destinationPort": "80",
            "id": 18,
            "latency": 0,
            "method": "GET",
            "protocol": {
                "abbr": "HTTP",
                "backgroundColor": "#205cf5",
                "fontSize": 12,
                "foregroundColor": "#ffffff",
                "longName": "Hypertext Transfer Protocol -- HTTP/1.1",
                "macro": "http",
                "name": "http",
                "ports": [
                    "80",
                    "443",
                    "8080"
                ],
                "priority": 0,
                "referenceLink": "https://datatracker.ietf.org/doc/html/rfc2616",
                "version": "1.1"
            },
            "requestSenderIp": "172.17.0.1",
            "rules": {},
            "service": "catalogue.sock-shop",
            "sourceIp": "172.17.0.1",
            "sourcePort": "54240",
            "statusCode": 200,
            "summary": "/health",
            "timestamp": 1638305660396,
            "url": "catalogue.sock-shop/health"
        },
        {
            "contractStatus": 0,
            "destinationIp": "172.17.0.6",
            "destinationPort": "80",
            "id": 17,
            "latency": 0,
            "method": "GET",
            "protocol": {
                "abbr": "HTTP",
                "backgroundColor": "#205cf5",
                "fontSize": 12,
                "foregroundColor": "#ffffff",
                "longName": "Hypertext Transfer Protocol -- HTTP/1.1",
                "macro": "http",
                "name": "http",
                "ports": [
                    "80",
                    "443",
                    "8080"
                ],
                "priority": 0,
                "referenceLink": "https://datatracker.ietf.org/doc/html/rfc2616",
                "version": "1.1"
            },
            "requestSenderIp": "172.17.0.1",
            "rules": {},
            "service": "catalogue.sock-shop",
            "sourceIp": "172.17.0.1",
            "sourcePort": "54242",
            "statusCode": 200,
            "summary": "/health",
            "timestamp": 1638305660396,
            "url": "catalogue.sock-shop/health"
        },
        {
            "contractStatus": 0,
            "destinationIp": "172.17.0.15",
            "destinationPort": "80",
            "id": 16,
            "latency": 0,
            "method": "GET",
            "protocol": {
                "abbr": "HTTP",
                "backgroundColor": "#205cf5",
                "fontSize": 12,
                "foregroundColor": "#ffffff",
                "longName": "Hypertext Transfer Protocol -- HTTP/1.1",
                "macro": "http",
                "name": "http",
                "ports": [
                    "80",
                    "443",
                    "8080"
                ],
                "priority": 0,
                "referenceLink": "https://datatracker.ietf.org/doc/html/rfc2616",
                "version": "1.1"
            },
            "requestSenderIp": "172.17.0.1",
            "rules": {},
            "service": "user.sock-shop",
            "sourceIp": "172.17.0.1",
            "sourcePort": "37478",
            "statusCode": 200,
            "summary": "/health",
            "timestamp": 1638305658690,
            "url": "user.sock-shop/health"
        }
    ],
    "meta": {
        "current": 16,
        "total": 565,
        "numberOfWritten": 4,
        "leftOff": 16
    }
}
```


agent/pkg/models/models.go
```diff
@@ -16,6 +16,14 @@ func GetEntry(r *tapApi.MizuEntry, v tapApi.DataUnmarshaler) error {
 	return v.UnmarshalData(r)
 }
 
+type EntriesRequest struct {
+	LeftOff   int    `form:"leftOff" validate:"required,min=0,max=2147483647"`
```
gadotroee, line 20: why the lefoOff is required?

agent/pkg/models/models.go
```diff
@@ -16,6 +16,14 @@ func GetEntry(r *tapApi.MizuEntry, v tapApi.DataUnmarshaler) error {
 	return v.UnmarshalData(r)
 }
 
+type EntriesRequest struct {
```
gadotroee, line 19: I'm thinking if we need to specify the "max" in the validation.. 
it will probably give better errorMessage but if the validate package is working well with specify only min, it might be enought. (and in case of bigger number we will probably got exception when trying to assign this number to the struct

agent/pkg/models/models.go
```diff
@@ -16,6 +16,14 @@ func GetEntry(r *tapApi.MizuEntry, v tapApi.DataUnmarshaler) error {
...
+	LeftOff   int    `form:"leftOff" validate:"required,min=0,max=2147483647"`
+	Direction int    `form:"direction" validate:"required,oneof='1' '-1'"`
+	Query     string `form:"query"`
+	Limit     int    `form:"limit" validate:"required,min=1,max=2147483647"`
+	TimeoutMs int    `form:"timeoutMs" validate:"required,min=1,max=2147483647"`
```
gadotroee, line 24: I think the timeout can be optional too. in wasn't added we will put default value (something like couple of seconds)

agent/pkg/controllers/entries_controller.go
```diff
@@ -36,6 +38,58 @@ func Error(c *gin.Context, err error) bool {
...
+	}
+
+	var metadata *basenine.Metadata
+	err = json.Unmarshal(meta, &metadata)
+	if err != nil {
+		logger.Log.Debugf("Error recieving metadata: %v", err.Error())
+	}
+
+	result["data"] = dataSlice
+	result["meta"] = metadata
```
gadotroee, line 88: I'm not sure the meta will be important but it is not harming to add metadata instead of iterating over the response 
 - Nice change.
 
The reason I wrote about it is that "GetEntries" not really returning array of entries.
Anyway this is a good change

agent/pkg/routes/entries_routes.go
```diff
@@ -10,5 +10,6 @@ import (
 func EntriesRoutes(ginApp *gin.Engine) {
 	routeGroup := ginApp.Group("/entries")
 
+	routeGroup.GET("/", controllers.GetEntries)  // get entries (base/thin entries)
```
gadotroee, line 13:
```suggestion
	routeGroup.GET("/", controllers.GetEntries)  // get entries( base/thin entries) and metadata
```

agent/pkg/controllers/entries_controller.go
gadotroee: Maybe we can just create struct instead of `[]interface{}` and `map[string]interface{}`

gadotroee (commented):
@mertyildiran 
Added few small comments - they are not blockers but I would like to hear your opinions

agent/pkg/models/models.go
mertyildiran: `leftOff` is the index the querying and fetching will start. That's why it's required. I forgot that it can take a negative value as well so I set min to `-1` with https://github.com/up9inc/mizu/commit/3d88ae84f41a1c8060bac0395f39de6f9c6f7760

agent/pkg/models/models.go
mertyildiran: Done with https://github.com/up9inc/mizu/commit/9e6919a01c1e9e7cdb253636a26560d275b49ea6

agent/pkg/models/models.go
mertyildiran: Done with https://github.com/up9inc/mizu/commit/c137437ee8326ad7acc71ee897459c4b794d5db0

agent/pkg/routes/entries_routes.go
mertyildiran: Done with https://github.com/up9inc/mizu/commit/2bbcba4b4017e190402768ce144bf7a8855afb16

agent/pkg/controllers/entries_controller.go
mertyildiran: Never return a JSON array. Instead return a JSON object that contains a JSON array. See;
- https://softwareengineering.stackexchange.com/a/327632/236831
- https://stackoverflow.com/questions/3503102/what-are-top-level-json-arrays-and-why-are-they-a-security-risk

gadotroee: @mertyildiran 
Maybe I didn't explain myself, I think that changing to object that holds "meta" and "result" (which is the really json array is OK. 
I just think that we can create struct 
```type EntriesResponse struct {
            Data map[string]interfact{}  /// if we can specify exact type here it is even better
            Meta *basenine.Metadata
```
and then return this, so when someone wants to make change or read this function it will be easier to understand what is this function returning.. 

Again, it is not blocker and I'm ok with current code - just think it might be better and want to know if there is any reason not to do it like this..

mertyildiran: @gadotroee I think https://github.com/up9inc/mizu/commit/2c263db7b9c7a306ca3ab968917e379c6f40f27a meets your request.

gadotroee (approved):

PR #517, opened 2021-12-01
Disable telemetry by env var MIZU_DISABLE_TELEMTRY

IgorGov:


cli/telemetry/telemetry.go
```diff
@@ -62,6 +63,9 @@ func ReportAPICalls(apiProvider *apiserver.Provider) {
 }
 
 func shouldRunTelemetry() bool {
+	if _, present := os.LookupEnv("MIZU_DISABLE_TELEMTRY"); present {
```
RoyUP9, line 66: Maybe change to const

IgorGov: I can do it, but can't see any advantage to it

RoyUP9 (approved):

PR #518, opened 2021-12-02
TRA-4017 Bring back `getOldEntries` method using fetch API and always start streaming from now

mertyildiran:
Makes it fetch the old records based on the given display filter when the scrollbar hits the top. It also adds a button to do the same when the scrollbar is not visible.

Makes it always start the stream from now with oppose to the current state in `develop` which starts from the very beginning of the capture.

### Caution

- It will probably break the acceptance tests since they were adapted to the pure WebSocket workflow with https://github.com/up9inc/mizu/pull/443 after the merger of https://github.com/up9inc/mizu/pull/279
- It introduces a - *requested but problematic* - behavior of filtering the old captured records. `getOldEntries` (through an HTTP request) does not guarantee to return a matching record in a given 3 seconds hard-coded timeout. The first ever matching record **can be seconds or even minutes away** from the current head in terms of filtering duration since we do [**lazy evaluation**](https://en.wikipedia.org/wiki/Lazy_evaluation). This makes the filtering past **practically impossible**.
- It introduces another - *requested but problematic* - behavior of WebSocket connection being **closed** upon the break of scrollbar snap to the bottom. The **connection can only be restored** by clicking to "snap to bottom" button or "play" button at top.
- It unfortunately removes the query progress information from the footer since it's **not possible to reliably track the query progress** because of the mixture of WebSocket and HTTP calls.
- Reintroduces the front-end total entries fetched through WebSocket limit that's removed in [here](https://github.com/up9inc/mizu/pull/268/files#diff-7170980be6ad275bd973945378d09e7705f54e48edd42e2daac359d7d9793c15L89-L92) but this time sets it to `10000` since the feed is virtualized now. From the experiments, it's observed that even `200000+` entries on front-end does not hinder the performance on an average machine but the limit is reintroduced because it's requested.
- Makes it lose the UI performance gain that's provided by https://github.com/up9inc/mizu/pull/452 in favor of complying with the defined behavior in `TRA-4017`.
- Completely reverts https://github.com/up9inc/mizu/pull/513

### Note

Please consider comparing `develop` and `feat/ui-paging-mode-3` branches side-by-side via building their Docker image and running them with a load test to evaluate the pros and cons of this PR.


gadotroee:
I found few bugs that I will try to show here:
1. 
image
the scroll bar is at the bottom and the results are at the top so I don't see them (after I had more than 10K results) and hit "apply" 

2. I can't reproduce it easily but I saw case that I have bigger number in the displayed that "total".. 

3. 
image
the arrow up button is displayed in incorrect conditions (it appears now when I click and nothing is changing and it is not there when I expect it to be 


Those things are from first try - so I will try to do more deeply check and explain with demos (you can ignore this comment for now)

ui/src/components/TrafficPage.tsx
```diff
@@ -134,23 +135,14 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus, onTLS
             switch (message.messageType) {
                 case "entry":
                     const entry = message.data;
-                    var focusThis = false;
-                    if (!focusedEntryId) {
-                        focusThis = true;
-                        setFocusedEntryId(entry.id.toString());
+                    if (!focusedEntryId) setFocusedEntryId(entry.id.toString())
+                    let newEntries = [...entries, entry];
```
lirazyehezkel, line 139: can be const

ui/src/components/EntriesList.tsx
```diff
@@ -1,33 +1,132 @@
...
                     <ScrollableFeedVirtualized ref={scrollableRef} itemHeight={48} marginTop={10} onSnapBroken={onSnapBrokenEvent}>
-                        {false /* TODO: why there is a need for something here (not necessarily false)? */}
-                        {entries}
+                        {false /* It's because the first child is ignored by ScrollableFeedVirtualized */}
+                        {memoizedEntries.map(entry => <EntryItem
+                            key={`entry-${entry.id}`}
+                            entry={entry}
+                            focusedEntryId={focusedEntryId}
+                            setFocusedEntryId={setFocusedEntryId}
+                            style={{}}
```
lirazyehezkel, line 111: let's make the style an optional param

ui/src/components/TrafficPage.tsx
mertyildiran: @lirazyehezkel done with https://github.com/up9inc/mizu/commit/5c05b613ab3b9dda63088fadd9c45e39eb844a3d

ui/src/components/EntriesList.tsx
mertyildiran: @lirazyehezkel since that property set on these lines https://github.com/mertyildiran/react-scrollable-feed/blob/master/src/index.tsx#L292-L293 it needs to be initialized and not optional.

A diff like this:

```diff
diff --git a/ui/src/components/EntriesList.tsx b/ui/src/components/EntriesList.tsx
index e6bfab0..479bb1c 100644
--- a/ui/src/components/EntriesList.tsx
+++ b/ui/src/components/EntriesList.tsx
@@ -108,7 +108,6 @@ export const EntriesList: React.FC<EntriesListProps> = ({entries, setEntries, qu
                             entry={entry}
                             focusedEntryId={focusedEntryId}
                             setFocusedEntryId={setFocusedEntryId}
-                            style={{}}
                             updateQuery={updateQuery}
                             headingMode={false}
                         />)}
diff --git a/ui/src/components/EntryListItem/EntryListItem.tsx b/ui/src/components/EntryListItem/EntryListItem.tsx
index 40d94e9..52543bb 100644
--- a/ui/src/components/EntryListItem/EntryListItem.tsx
+++ b/ui/src/components/EntryListItem/EntryListItem.tsx
@@ -40,9 +40,9 @@ interface EntryProps {
     entry: Entry;
     focusedEntryId: string;
     setFocusedEntryId: (id: string) => void;
-    style: object;
     updateQuery: any;
     headingMode: boolean;
+    style?: object;
 }
 
 export const EntryItem: React.FC<EntryProps> = ({entry, focusedEntryId, setFocusedEntryId, style, updateQuery, headingMode}) => {
```

causes the error below:

Screenshot from 2021-12-08 14-45-07

lirazyehezkel (approved):

mertyildiran:
It didn't break the acceptance tests, cool! :smile:

PR #524, opened 2021-12-09
Include milliseconds information into the timestamps in the UI

mertyildiran:
Screenshot from 2021-12-10 04-00-49


gadotroee (commented):
@mertyildiran 
 so now all times will be UTC, right? (Ideally I think frontend should have some settings to show it in current timezone) but it will add complexity that we don't need now and at least it will be less confusing

mertyildiran:
> @mertyildiran so now all times will be UTC, right? (Ideally I think frontend should have some settings to show it in current timezone) but it will add complexity that we don't need now and at least it will be less confusing

@gadotroee yes, correct it's all UTC now.

gadotroee (approved):
Looking good

PR #525, opened 2021-12-10
Remove unnecessary fields and split `service` into `src.name` and `dst.name`

mertyildiran:
Screenshot from 2021-12-12 19-29-01


ui/src/components/EntryListItem/EntryListItem.tsx
```diff
@@ -146,17 +148,30 @@ export const EntryItem: React.FC<EntryProps> = ({entry, focusedEntryId, setFocus
...
+                        <span
+                            title="Source Name"
+                        >
+                            {entry.src.name ? entry.src.name : entry.src.ip}
+                        </span>
+                    </Queryable>
+                    <div style={{marginRight: "10px"}}>{"->"}</div>
                     <Queryable
-                        query={`service == "${entry.service}"`}
+                        query={`dst.name == "${entry.dst.name}"`}
```
gadotroee, line 166: if I understand correctly when there dest.name does not exists this filter will add "dst.name == undefined " or "dst.name == null" (because when there is no name we are showing ip. but this will not be the filter that will be added... maybe we need to address this difference ?

ui/src/components/EntryListItem/EntryListItem.tsx
```diff
@@ -146,17 +148,30 @@ export const EntryItem: React.FC<EntryProps> = ({entry, focusedEntryId, setFocus
...
+                    <Queryable
+                        query={`src.name == "${entry.src.name}"`}
+                        updateQuery={updateQuery}
+                        displayIconOnMouseOver={true}
+                        style={{marginTop: "-4px", marginRight: "10px"}}
+                    >
+                        <span
+                            title="Source Name"
+                        >
+                            {entry.src.name ? entry.src.name : entry.src.ip}
```
gadotroee, line 161: when entry.src.name is null, undefined  or empty string it will show IP which is ok but then clicking on it will add filter of 
src.name == <value> which is not what user clicked on.. 
same goes to the "dest.name" in line 172

mertyildiran: @gadotroee it can only be `""` but not `null`. Yes in that case the filter is `src.name == ""`. We have already a clickable UI element for the `src.ip` and `dst.ip` so no need to duplicate them. We inform the user that this field is `src.name`.

gadotroee (dismissed):

mertyildiran:
247aef8fe5448dfce77a488c9291b9c233874bf5 upon @IgorGov 's request.

Also replaced the separator `->` text with `SwapHorizIcon` for having a more beautiful visuals.

haiut (approved):

PR #531, opened 2021-12-15
API server stores tappers status

IgorGov:
Server health will have the tapped pod per machine & tapper status per machine, so we will be able to let user know that some of the pods are not tapped
image

Working with watching to k8s events instead of pods, to avoid use case when no pod alert but event is fired (for instance "FailedScheduling"), and logs will get much more reacher info about the mizu API server and tappers
image


shared/models.go
```diff
@@ -75,6 +81,7 @@ type TapStatus struct {
 type PodInfo struct {
 	Namespace string `json:"namespace"`
 	Name      string `json:"name"`
+	NodeName  string `json:"NodeName"`
```
gadotroee, line 84: lower case?

shared/kubernetes/mizuTapperSyncer.go
```diff
@@ -65,9 +69,72 @@ func CreateAndStartMizuTapperSyncer(ctx context.Context, kubernetesProvider *Pro
 	}
 
 	go syncer.watchPodsForTapping()
+	go syncer.watchTapperEvents()
 	return syncer, nil
 }
 
+func (tapperSyncer *MizuTapperSyncer) watchTapperEvents() {
+	mizuResourceRegex := regexp.MustCompile(fmt.Sprintf("^%s.*", TapperPodName))
+	eventWatchHelper := NewEventWatchHelper(tapperSyncer.kubernetesProvider, mizuResourceRegex, "pod")
```
gadotroee, line 78: this should be probably a consts (that maybe exists in the k8s libraray)

cli/mizu/consts.go
```diff
@@ -6,8 +6,8 @@ import (
 )
 
 var (
-	SemVer                                    = "0.0.1"
-	Branch                                    = "develop"
+	SemVer                                    = "0.0.0"
+	Branch                                    = "feature/tapped_pod_stats"
```
gadotroee: mistake?

IgorGov: update

agent/main.go
```diff
@@ -477,6 +474,16 @@ func startMizuTapperSyncer(ctx context.Context, provider *kubernetes.Provider) (
 				api.BroadcastToBrowserClients(serializedTapStatus)
 				providers.TapStatus.Pods = tapStatus.Pods
 				providers.ExpectedTapperAmount = tapPodChangeEvent.ExpectedTapperAmount
+			case tapperStatus, ok := <-tapperSyncer.TapperStatusChangedOut:
+				if !ok {
```
gadotroee, line 478: don't you need to change the channel to nil or something?

cli/cmd/tapRunner.go
```diff
@@ -557,11 +553,9 @@ func waitUntilNamespaceDeleted(ctx context.Context, cancel context.CancelFunc, k
 }
 
 func watchApiServerPod(ctx context.Context, kubernetesProvider *kubernetes.Provider, cancel context.CancelFunc) {
-	podExactRegex := regexp.MustCompile(fmt.Sprintf("^%s$", kubernetes.ApiServerPodName))
-	podWatchHelper := kubernetes.NewPodWatchHelper(kubernetesProvider, podExactRegex)
-	eventChan, errorChan := kubernetes.FilteredWatch(ctx, podWatchHelper, []string{config.Config.MizuResourcesNamespace}, podWatchHelper)
-	isPodReady := false
-	timeAfter := time.After(25 * time.Second)
+	podExactRegex := regexp.MustCompile(fmt.Sprintf("^%s", kubernetes.ApiServerPodName))
```
gadotroee, line 556: why did you remove the "$" form the regex?

gadotroee: maybe it will not work when it is daemon (because it is deployment)

gadotroee (commented):
added comments

gadotroee (approved):

PR #547, opened 2021-12-20
support linkerd - initial commit

dvdlevanon:


tap/source/discoverer_util.go
```diff
@@ -0,0 +1,38 @@
...
+	"io/ioutil"
+	"regexp"
+	"strings"
+
+	"github.com/up9inc/mizu/shared/logger"
+)
+
+var numberRegex = regexp.MustCompile("[0-9]+")
+
+func readEnvironmentVariable(file string, name string) (string, error) {
```
RamiBerm, line 13: `getSingleValueFromEnvironmentVariableFile(filePath string, variableName string)` would be more clean

RamiBerm (changes requested):

RamiBerm (approved):

PR #549, opened 2021-12-22
Create README.md

sararozenfeld:
Release notes version update for V0.21.0

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,23 @@
+Release Notes 
+
+V0.21.0
+Main features
+* New Traffic search & Stream experience 
+  * Rich query language with full-text search capabilities on headers & body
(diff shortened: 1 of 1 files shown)


haiut (approved):

gadotroee (changes requested):
do we know this one is to "main" and will create new version?

docs/README.md
```diff
@@ -0,0 +1,23 @@
+Release Notes 
+
+V0.21.0
```
IgorGov, line 3: the name: should be named "release notes" / "Changelog"

docs/README.md
```diff
@@ -0,0 +1,23 @@
+Release Notes
```
IgorGov, line 1: we should add formating, headers, etc
similar to what we have in mockintosh 
https://github.com/up9inc/mockintosh/blob/main/docs/Changelog.md

PR #553, opened 2021-12-22
Remove TCP related logging to reduce the amount of logs on `DEBUG` level

mertyildiran:
The diff to demonstrate how the amount of logs in tapper daemon set pod, can be reduced in case of debug level is `DEBUG`.


gadotroee (commented):
I think it is a good change - let's do this and maybe add logs that help us.

gadotroee (approved):

PR #556, opened 2021-12-26
TRA-4065 support inflight tap target update

RamiBerm:


tap/passive_tapper.go
```diff
@@ -96,6 +101,20 @@ func StartPassiveTapper(opts *TapOpts, outputItems chan *api.OutputChannelItem,
...
+	initializePacketSources()
+	printNewTapTargets()
+}
+
+func printNewTapTargets() {
+	printStr := ""
+	for _, tapTarget := range tapTargets {
+		printStr += tapTarget.Status.PodIP + " "
+	}
+	logger.Log.Infof("Now tapping: %s", printStr)
```
IgorGov, line 115: should we print a pod name as well, as this is what user see, and will open github issues about..

RamiBerm: very good point

IgorGov (approved):

IgorGov (approved):

PR #572, opened 2021-12-29
Support custom RBAC resources

nimrod-up9:
This PR adds supports for a use case where a user wants to edit Mizu's `ServiceAccount`, `(Cluster)Role` and `(Cluster)RoleBindings`.

Suppose a user wants to edit one of the RBAC resources (`ServiceAccount`, `(Cluster)Role` and `(Cluster)RoleBindings).
In its current state, Mizu will not create RBAC resources if they already exist. This allows users to override the default `ClusterRole` by creating their own `ClusterRole` by the same name before running Mizu.
However, Mizu will delete the user defined `ClusterRole` after the `tap` command is done (or when the `clean` command is issued). This behavior forces the user to create the `ClusterRole` before every run.

This PR adds labels to every k8s resource that Mizu creates.

The labels are recommended in the [k8s docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/).
`app.kubernetes.io/created-by` and `app.kubernetes.io/managed-by`.
`created-by` takes on the values `mizu-cli` and `mizu-agent`. `managed-by` takes the value `mizu`.

When Mizu cleans the cluster (CTRL-C or `mizu clean`) it will delete all RBAC resources which have the label `app.kubernetes.io/managed-by=mizu`. This allows users to retain their custom `ClusterRole` (or other RBAC resources) by not having the `managed-by` label or by setting it to something else (`custom`, `manual`, `my-app` etc.).

`managed-by` and `created-by` when running in daemon mode:
```sh
kubectl get -A namespace,pod,deployment,service,daemonset,configmap,persistentvolumeclaim,serviceaccount,clusterrole,clusterrolebinding,role,rolebinding -o json | jq "[\"KIND\", \"NAME\", \"MANAGED_BY\", \"CREATED_BY\", \"NAMESPACE\"], (.items[] | select(.metadata.name | contains(\"mizu\")) | [.kind, .metadata.name, .metadata.labels[\"app.kubernetes.io/managed-by\"], .metadata.labels[\"app.kubernetes.io/created-by\"], .metadata.namespace]) | @tsv" -r | column -t
```
image

Note that the new logic only applies to RBAC resources. Removal of other kinds of resources is not affected by this PR. In non-restricted mode `mizu` `Namespace` and all of its inhabitants are always removed regardless of their labels. In restricted mode `Deployments`, `Roles`, `Pods` etc. are always removed regardless of their labels.


nimrod-up9:
Thoughts:
- Are Labels more suitable for this use case than Annotations?
- I will add in this PR or in a separate one a message when a resource was skipped during the clean phase because it doesn't have the annotation.

IgorGov:
So the user experience here is,
Run ./Mizu tap -> edit service role, remove annotation -> CTRL+C (Mizu doesn't clean service role) 
1. Correct? 
2. will it remove the Mizu namespace?
3. Why using annotation and not labels? 
4. Can make it even more generic, on clean get all resource with label "managed-by:Mizu" -> delete them (if not possible then per resource)

shared/kubernetes/consts.go
```diff
@@ -17,3 +17,9 @@ const (
 	PersistentVolumeClaimName  = MizuResourcesPrefix + "volume-claim"
 	MinKubernetesServerVersion = "1.16.0"
 )
+
+const (
+	MizuLabelPrefix         = "mizu.io/"
+	LabelMizuManagement     = MizuLabelPrefix + "management-method"
```
IgorGov, line 23: an idea we should consider, as we planing to support mizu Ent. install with helm, we should use the same label:
"managed-by".
the value can be something like "mizu-cli", when mizu cli created to resources, and responsible of cleaning them

IgorGov (approved):

PR #574, opened 2021-12-30
Upgrade Basenine version from `0.2.19` to `0.2.20` and add the support of indexed queries

mertyildiran:


mertyildiran:
Closing because the upgrade is introducing instability.

PR #581, opened 2022-01-03
Added update config route for install mode

RoyUP9:


agent/pkg/routes/standalone_routes.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/gin-gonic/gin"
+	"mizuserver/pkg/controllers"
+)
+
+func StandaloneRoutes(ginApp *gin.Engine) {
+	routeGroup := ginApp.Group("/standalone")
+
+	routeGroup.POST("/updateConfig", controllers.UpdateConfig)
```
IgorGov, line 11: maybe we should call it updateTapConfig?
next we will have many different configurations, users, workspace etc..

shared/kubernetes/mizuTapperSyncer.go
```diff
@@ -145,6 +148,10 @@ func (tapperSyncer *MizuTapperSyncer) watchTapperEvents() {
 					event.Reason,
 					event.Note))
 
+			if event.Reason == "Killing" {
```
IgorGov, line 151: why should we have special treat for killing?
it will continue in the next if...
maybe  add additional info to debug log, like the event reason?

agent/pkg/models/models.go
```diff
@@ -16,6 +16,11 @@ func GetEntry(r *tapApi.MizuEntry, v tapApi.DataUnmarshaler) error {
 	return v.UnmarshalData(r)
 }
 
+type StandaloneConfig struct {
+	TargetNamespaces []string `json:"targetNamespaces"`
+	PodRegex         string   `json:"podRegex"`
```
IgorGov, line 21: should be removed

agent/pkg/models/models.go
```diff
@@ -16,6 +16,11 @@ func GetEntry(r *tapApi.MizuEntry, v tapApi.DataUnmarshaler) error {
 	return v.UnmarshalData(r)
 }
 
+type StandaloneConfig struct {
+	TargetNamespaces []string `json:"targetNamespaces"`
```
IgorGov, line 20: this one should be list / map of namespace -> bool (tapped or not..)

shared/kubernetes/mizuTapperSyncer.go
RoyUP9: The problem is that 'killing' reason causes the next if to get an error

agent/main.go
```diff
@@ -256,13 +252,14 @@ func hostApi(socketHarOutputChannel chan<- *tapApi.OutputChannelItem) {
...
 	if err := setUIMode(); err != nil {
-		logger.Log.Panicf("Error setting ui mode, err: %v", err)
+		logger.Log.Errorf("Error setting ui mode, err: %v", err)
 	}
 	app.Use(static.ServeRoot("/", "./site"))
 
 	app.Use(CORSMiddleware()) // This has to be called after the static middleware, does not work if its called before
 
 	api.WebSocketRoutes(app, &eventHandlers, startTime)
+	routes.StandaloneRoutes(app)
```
IgorGov, line 262: hostApi is called for basic mizu as well, am i wrong?
if so basic mizu should have these routes?

agent/pkg/routes/standalone_routes.go
```diff
@@ -0,0 +1,13 @@
+package routes
+
+import (
+	"github.com/gin-gonic/gin"
+	"mizuserver/pkg/controllers"
+)
+
+func StandaloneRoutes(ginApp *gin.Engine) {
+	routeGroup := ginApp.Group("/standalone")
```
IgorGov, line 9: mizu Ent. is basically all standalone mode, maybe it should be "/" instead of "/standalone"

agent/pkg/controllers/standalone_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	cancelTapperSyncer = cancel
+	globalTapConfig = tapConfig
+
+	c.JSON(http.StatusOK, "OK")
+}
+
+func GetTapConfig(c *gin.Context) {
+	if globalTapConfig != nil {
```
IgorGov, line 70: why there is a case where server doesn't sends any response? client will wait till timeout?

agent/pkg/controllers/standalone_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	v1 "k8s.io/api/core/v1"
+	"mizuserver/pkg/config"
+	"mizuserver/pkg/models"
+	"mizuserver/pkg/providers"
+	"net/http"
+	"regexp"
+	"time"
+)
+
+var globalTapConfig *models.StandaloneTapConfig
```
IgorGov, line 19: can you open a task in Jira for making it persistent, so we won't forget it...

agent/pkg/controllers/standalone_controller.go
```diff
@@ -0,0 +1,128 @@
...
+		return
+	}
+
+	if cancelTapperSyncer != nil {
+		cancelTapperSyncer()
+
+		providers.TapStatus = shared.TapStatus{}
+		providers.TappersStatus = make(map[string]shared.TapperStatus)
+
+		broadcastTappedPodsStatus()
```
IgorGov, line 36: why should we broadcast this state?

agent/pkg/controllers/standalone_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	kubernetesProvider, err := kubernetes.NewProviderInCluster()
```
IgorGov, line 41: why do we need to create a new provider each post request?
why shouldn't we create it once and use it?

agent/pkg/controllers/standalone_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		c.JSON(http.StatusBadRequest, err)
+		cancel()
```
IgorGov, line 44: what is being canceled here exactly?

agent/pkg/controllers/standalone_controller.go
RoyUP9: so the ui will be updated to zero pods tapped.
if we do not add this and no pods are being tapped in the new change, it will show the old pods that are not being tapped anymore

agent/pkg/controllers/standalone_controller.go
RoyUP9: goland recommended to add it here, it is angry if i remove it

agent/pkg/controllers/standalone_controller.go
RoyUP9: my next pr will be about this case, didn't want to make a big pr

agent/pkg/routes/standalone_routes.go
RoyUP9: i think that for now we should separate it to a different route group, because if we keep adding routes to "/" it's not good

IgorGov: but you will keep adding it only when agent runs in standalone mode, so it will be transparent to "mizu basic"

agent/pkg/controllers/standalone_controller.go
IgorGov: can't we just return empty list? or some "bad request" status with text "you should set config before get..."

agent/pkg/controllers/standalone_controller.go
IgorGov: got you..

agent/pkg/controllers/standalone_controller.go
IgorGov: as the cancel object is never used before this line, you can create it after it

IgorGov (dismissed):

agent/pkg/routes/standalone_routes.go
RoyUP9: i think that i'll keep it as is for now for a clear separation in the route

IgorGov: Maybe we should change it to config route

IgorGov (approved):

agent/pkg/routes/standalone_routes.go
RoyUP9: so i'll change it to config in the second pr

PR #587, opened 2022-01-04
Added endpoint for getting tapped namespaces

RoyUP9:


agent/pkg/controllers/config_controller.go
```diff
@@ -28,6 +28,15 @@ func PostTapConfig(c *gin.Context) {
 		return
 	}
 
+	if kubernetesProvider == nil {
+		var err error
+		kubernetesProvider, err = kubernetes.NewProviderInCluster()
```
IgorGov, line 33: can we use here a singleton pattern? I am sure there are other places in code we use this provider

agent/pkg/controllers/config_controller.go
```diff
@@ -70,11 +70,54 @@ func PostTapConfig(c *gin.Context) {
 }
 
 func GetTapConfig(c *gin.Context) {
-	if globalTapConfig != nil {
-		c.JSON(http.StatusOK, globalTapConfig)
+	tappedNamespaces := make(map[string]bool)
+
+	if config.Config.IsNsRestrictedMode {
```
IgorGov, line 75: i think we need to complicate our lives with this mode in standalone mode... just remove this support

agent/pkg/controllers/config_controller.go
```diff
@@ -70,11 +70,54 @@ func PostTapConfig(c *gin.Context) {
...
+		tappedNamespaces[config.Config.MizuResourcesNamespace] = tapped
+		tapConfig := models.TapConfig{TappedNamespaces: tappedNamespaces}
+
+		c.JSON(http.StatusOK, tapConfig)
+		return
+	}
+
+	if kubernetesProvider == nil {
+		var err error
+		kubernetesProvider, err = kubernetes.NewProviderInCluster()
```
IgorGov, line 90: should be used as a singleton

agent/pkg/controllers/config_controller.go
```diff
@@ -70,11 +70,54 @@ func PostTapConfig(c *gin.Context) {
...
+	}
+
+	for _, namespace := range namespaces {
+		if namespace.Name == config.Config.MizuResourcesNamespace {
+			continue
+		}
+
+		tapped := false
+		if globalTapConfig != nil {
+			tapped = globalTapConfig.TappedNamespaces[namespace.Name]
```
IgorGov, line 113: what if there is a newly created namespace, it can throw exception...

shared/kubernetes/provider.go
```diff
@@ -392,7 +392,7 @@ func (provider *Provider) CreateMizuRBAC(ctx context.Context, namespace string,
 		Rules: []rbac.PolicyRule{
 			{
 				APIGroups: []string{"", "extensions", "apps"},
-				Resources: []string{"pods", "services", "endpoints"},
+				Resources: []string{"pods", "services", "endpoints", "namespaces"},
```
IgorGov, line 395: it is needed only for the standalone mode

agent/pkg/controllers/config_controller.go
IgorGov: your current implementation is not a thread safe ..

IgorGov: @RoyUP9 i will merge this PR because of the time pressure, but i think we need to implement a singleton here.

IgorGov (approved):

PR #603, opened 2022-01-07
Move Basenine binary into a separate container

mertyildiran:
This PR puts the Basenine binary into a separate container named `basenine` instead of spawning a process directly from inside the `mizu-api-server` container.


gadotroee (commented):
I think this is a great change.
still need to understand that it work well on pod restart (because we have persistent storage it should work IMO)

shared/kubernetes/provider.go
```diff
@@ -278,6 +278,36 @@ func (provider *Provider) GetMizuApiServerPodObject(opts *ApiServerOptions, moun
 				},
 			},
 		},
+		{
+			Name:            "basenine",
+			Image:           "ghcr.io/up9inc/basenine:v0.2.26",
```
gadotroee, line 283: maybe this can be configuration option or at least const in the code so we don't need to search inside the function to see which image do we take

shared/kubernetes/provider.go
```diff
@@ -278,6 +278,36 @@ func (provider *Provider) GetMizuApiServerPodObject(opts *ApiServerOptions, moun
...
+			ImagePullPolicy: opts.ImagePullPolicy,
+			VolumeMounts:    volumeMounts,
+			ReadinessProbe: &core.Probe{
+				FailureThreshold: 3,
+				Handler: core.Handler{
+					TCPSocket: &core.TCPSocketAction{
+						Port: intstr.FromInt(9099),
```
gadotroee, line 290: this is also look like a magic number - and I don't even see that we pass it to the command so I guess this is the default value inside basenine.. maybe it can be better to specify this port in the command and make it variable that we pass to both places (here in the ReadinessProbe and down in the Args for the container command)

shared/kubernetes/provider.go
mertyildiran: @gadotroee the image is version locked. So user shouldn't be able to change this image. I put the information into `consts.go` with https://github.com/up9inc/mizu/commit/b44f304618cd52cbf0eace292366ea79b3da46a4

shared/kubernetes/provider.go
mertyildiran: @gadotroee done with https://github.com/up9inc/mizu/commit/b44f304618cd52cbf0eace292366ea79b3da46a4

gadotroee (approved):
Looking good

gadotroee (approved):

PR #578, opened 2022-01-01
Generate OAS from traffic

undera:

PR #606, opened 2022-01-08
Replace source IP with X-Forwarded-For in http, if exists

nimrod-up9:
Support source IP resolving for HTTP traffic in Istio service mesh.

If `X-Forwarded-For` HTTP request header is present, replace the source address with the right most address in `X-Forwarded-For` (latest in Envoy implementation of this header).

This allows Mizu to resolve source IPs in Istio service meshes, if the `use_remote_address` option is on. Added instructions on how to turn it on.

Drawback:
- The implementation in this PR overwrites the original IP that was intercepted. It displays the IP in XFF as the source IP. This might be confusing for users that expect to the raw traffic, like they would see with tcpdump and wireshark.
- This solution does not handle source IP resolving for non-HTTP traffic or for service meshes which do not use Envoy (like Linkerd).
- Are there setups where appending to `X-Forwarded-For` could mess up the user's system?


ui/src/components/EntryListItem/EntryListItem.tsx
```diff
@@ -214,7 +214,7 @@ export const EntryItem: React.FC<EntryProps> = ({entry, focusedEntryId, setFocus
                         {entry.src.ip}
                     </span>
                 </Queryable>
-                <span className={`${styles.tcpInfo}`} style={{marginTop: "18px"}}>:</span>
+				<span className={`${styles.tcpInfo}`} style={{marginTop: "18px"}}>{entry.src.port ? ":" : ""}</span>
```
mertyildiran, line 217: Why adding a ternary operator here? Is it possible to `entry.src.port` to be a empty string or `null`? As far as I know it's not possible.

nimrod-up9: When `X-Forwarded-For` is not empty, we change src IP to the IP in `X-Forwarded-For`, and the src port to `""` (`X-Forwarded-For` does not contain any info regarding the port).
https://github.com/up9inc/mizu/pull/606/files#diff-b90c839c6bda9ab1004f6293c5c5f37858c1597e9f6bfcbf2c2c34e4296e51f6R45

This way the IP is propagated to `resolveIP` and to UI without any more code changes.
https://github.com/up9inc/mizu/blob/e4aa94858cd2547fab6e28a9461ecfda6aa6d614/agent/pkg/api/main.go#L120

mertyildiran: @nimrod-up9 oh I understand now. I think we should do none of those changes to IP and port. Because that means destroying the data. The forwarded request is a different TCP stream. Why altering the reality? To confuse the person who inspects the traffic?

nimrod-up9: I agree ineprincipal, but I had to come up with a quick fix that doen't require a lot of refactoring.

I would love to improve this solution (in a follow up PR). Do you have an idea how we could append additional info in the DB entry? I guess that would require expanding the data model.

nimrod-up9:
I made a mistake. The IP to take from the header shouldebe the leftmost (earliest).

tap/extensions/http/handlers.go
```diff
@@ -21,9 +21,30 @@ func filterAndEmit(item *api.OutputChannelItem, emitter api.Emitter, options *ap
...
+	forwardedFor := request.Header.Get("X-Forwarded-For")
+	if forwardedFor == "" {
+		return
+	}
+
+	ips := strings.Split(forwardedFor, ",")
+	lastIP := strings.TrimSpace(ips[0])
+
+	item.ConnectionInfo.ClientIP = lastIP
+	item.ConnectionInfo.ClientPort = ""
```
RamiBerm, line 45:
```suggestion
        // the client port we see for forwarded traffic is not the original port
	item.ConnectionInfo.ClientPort = ""
```

RamiBerm (changes requested):

tap/extensions/http/handlers.go
RamiBerm: theres probably a better way to word this than what I came up with

ui/src/components/EntryListItem/EntryListItem.tsx
mertyildiran: @nimrod-up9 I'm not sure about the intention of this change. But if the purpose is doing analysis on the traffic (e.g. yet to be introduced features like https://github.com/up9inc/mizu/pull/623 or https://github.com/up9inc/mizu/pull/578) then this `src.ip` and `src.port` replacements shouldn't be done in the dissection layer. Instead it should be done in an upper layer where the TCP stream is displayed as is but the feature treats the TCP stream differently since it's forwarded.

For now, I guess it's OK to do it like this if it's urgent.

dvdlevanon (approved):

mertyildiran (dismissed):

tap/extensions/http/handlers.go
nimrod-up9: Thanks, I will adopt.

RamiBerm (approved):

mertyildiran (approved):

PR #630, opened 2022-01-12
Added post install check

RoyUP9:


cli/cmd/check.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"github.com/creasty/defaults"
+	"github.com/spf13/cobra"
+	"github.com/up9inc/mizu/cli/config/configStructs"
+	"github.com/up9inc/mizu/cli/telemetry"
+)
+
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "Checks that mizu was installed successfully",
```
IgorGov, line 12:
```suggestion
	Short: "Check the Mizu installation for potential problems",
```

cli/cmd/check.go
```diff
@@ -0,0 +1,27 @@
...
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(checkCmd)
+
+	defaultCheckConfig := configStructs.CheckConfig{}
+	defaults.Set(&defaultCheckConfig)
+
+	checkCmd.Flags().StringP(configStructs.ServerUrlCheckName, "u", defaultCheckConfig.ServerUrl, "Provide a custom server url")
```
IgorGov, line 26: why do we need it?

cli/cmd/checkRunner.go
```diff
@@ -0,0 +1,179 @@
...
+func checkKubernetesApi() (*kubernetes.Provider, bool) {
+	logger.Log.Infof("\nkubernetes-api\n--------------------")
+
+	kubernetesProvider, err := getKubernetesProviderForCli()
+	if err != nil {
+		logger.Log.Errorf("%v can't initialize the client, err: %v", fmt.Sprintf(uiUtils.Red, "✗"), err)
+		return nil, false
+	}
+
+	logger.Log.Infof("%v can initialize the client", fmt.Sprintf(uiUtils.Green, "√"))
```
IgorGov, line 46: we should add the check of "√ can query the Kubernetes API" - see if the cluster is accessible to the user

cli/cmd/check.go
RoyUP9: If user didn't use proxy, he can use this url to check the connectivity instead of the proxy

cli/cmd/checkRunner.go
```diff
@@ -0,0 +1,192 @@
...
+		logger.Log.Errorf("%v can't initialize the client, err: %v", fmt.Sprintf(uiUtils.Red, "✗"), err)
+		return nil, nil, false
+	}
+	logger.Log.Infof("%v can initialize the client", fmt.Sprintf(uiUtils.Green, "√"))
+
+	if err := kubernetesProvider.ValidateNotProxy(); err != nil {
+		logger.Log.Errorf("%v mizu doesn't support running through a proxy to k8s server, err: %v", fmt.Sprintf(uiUtils.Red, "✗"), err)
+		return nil, nil, false
+	}
+	logger.Log.Infof("%v not running mizu through a proxy to k8s server", fmt.Sprintf(uiUtils.Green, "√"))
```
IgorGov, line 61: I am not sure user will understand this line.
Maybe we should print it to debug, and not info

RoyUP9: but we somehow need to fail the check after this, because it will not work properly for him

cli/cmd/checkRunner.go
```diff
@@ -0,0 +1,192 @@
...
+		logger.Log.Errorf("%v can't query the Kubernetes API, err: %v", fmt.Sprintf(uiUtils.Red, "✗"), err)
+		return nil, nil, false
+	}
+	logger.Log.Infof("%v can query the Kubernetes API", fmt.Sprintf(uiUtils.Green, "√"))
+
+	return kubernetesProvider, kubernetesVersion, true
+}
+
+func validateCommand(ctx context.Context, kubernetesProvider *kubernetes.Provider) (bool, bool) {
+	logger.Log.Infof("\nvalidate-mizu-command\n--------------------")
```
IgorGov, line 74: Mizu mode

cli/cmd/checkRunner.go
```diff
@@ -0,0 +1,186 @@
...
+		logger.Log.Infof("%v mizu running with tap command", fmt.Sprintf(uiUtils.Green, "√"))
+		return true, false
+	}
+}
+
+func checkKubernetesVersion(kubernetesVersion *semver.SemVersion) bool {
+	logger.Log.Infof("\nkubernetes-version\n--------------------")
+
+	if err := kubernetes.ValidateKubernetesVersion(kubernetesVersion); err != nil {
+		logger.Log.Errorf("%v not running the minimum Kubernetes API version, err: %v", fmt.Sprintf(uiUtils.Red, "✗"), err)
```
IgorGov, line 86: too long version

IgorGov (approved):

PR #648, opened 2022-01-16
Fix: no panic on failure to sync entries to up9.app

IgorGov:


agent/main.go
```diff
@@ -128,7 +128,8 @@ func main() {
 		syncEntriesConfig := getSyncEntriesConfig()
 		if syncEntriesConfig != nil {
 			if err := up9.SyncEntries(syncEntriesConfig); err != nil {
-				panic(fmt.Sprintf("Error syncing entries, err: %v", err))
+				logger.Log.Error("Error syncing entries, err: %v", err)
+				return
```
RoyUP9, line 132: this will cause hostApi function to not be called, so api server will not function.
is this what we want?

IgorGov: you are right

RoyUP9 (approved):

PR #659, opened 2022-01-17
Add ARM64 and cross-compilation support to the agent image

mertyildiran:
Updates the `Dockerfile` and related shell scripts to have multiple microprocessor architecture support through cross-compilation. Mainly related to adding support of ARM64.

Changes the base image from `alpine:3.15` to `busybox:latest`.

It also makes the Mizu agent binary statically compiled.

The image size is reduced to 20.81 MB

The arguments `BUILDARCH` and `TARGETARCH` are BuildKit compatible.

### From AMD64 (x86-64) to AMD64 (x86-64)

```shell
$ docker build . -t up9inc/mizu:develop-amd64 --build-arg BUILDARCH=amd64 --build-arg TARGETARCH=amd64
$ docker push up9inc/mizu:develop-amd64
```

### From ARM64 (AArch64) to ARM64 (AArch64)

```shell
$ docker build . -t up9inc/mizu:develop-arm64v8 --build-arg BUILDARCH=arm64v8 --build-arg TARGETARCH=arm64v8
$ docker push up9inc/mizu:develop-arm64v8
```

### From AMD64 (x86-64) to ARM64 (AArch64)

```shell
$ docker build . -t up9inc/mizu:develop-arm64v8 --build-arg BUILDARCH=amd64 --build-arg TARGETARCH=arm64v8
$ docker push up9inc/mizu:develop-arm64v8
```

### Creating a Docker manifest which lists the OS/ARCH pairs

```shell
$ docker manifest create \
    up9inc/mizu:develop \
    --amend up9inc/mizu:develop-amd64 \
    --amend up9inc/mizu:develop-arm64v8
$ docker manifest push up9inc/mizu:develop
```

which guides Docker to select the correct architecture for the subject machine.

### `up9inc/linux-arm64-musl-go-libpcap`

This is the builder image for achieving x86-64 to AArch64 cross-compilation. Which is included into source.

```Dockerfile
FROM dockcross/linux-arm64-musl:latest AS builder-from-amd64-to-arm64v8

# Install Go
RUN curl https://go.dev/dl/go1.16.13.linux-amd64.tar.gz -Lo ./go.linux-amd64.tar.gz
RUN curl https://go.dev/dl/go1.16.13.linux-amd64.tar.gz.asc -Lo ./go.linux-amd64.tar.gz.asc
RUN curl https://dl.google.com/dl/linux/linux_signing_key.pub  -Lo linux_signing_key.pub
RUN gpg --import linux_signing_key.pub && gpg --verify ./go.linux-amd64.tar.gz.asc ./go.linux-amd64.tar.gz
RUN rm -rf /usr/local/go && tar -C /usr/local -xzf go.linux-amd64.tar.gz
ENV PATH "$PATH:/usr/local/go/bin"

# Compile libpcap from source
RUN curl https://www.tcpdump.org/release/libpcap-1.10.1.tar.gz -Lo ./libpcap.tar.gz
RUN curl https://www.tcpdump.org/release/libpcap-1.10.1.tar.gz.sig -Lo ./libpcap.tar.gz.sig
RUN curl https://www.tcpdump.org/release/signing-key.asc -Lo ./signing-key.asc
RUN gpg --import signing-key.asc && gpg --verify libpcap.tar.gz.sig libpcap.tar.gz
RUN tar -xzf libpcap.tar.gz && mv ./libpcap-* ./libpcap
RUN cd ./libpcap && ./configure --host=arm && make
RUN cp /work/libpcap/libpcap.a /usr/xcc/aarch64-linux-musl-cross/lib/gcc/aarch64-linux-musl/*/
```

## CI

`publish.yml` is adjusted for cross-compilation.

## Testing

On an ARM64 machine with `minikube`;

```shell
$ make cli && ./cli/bin/mizu__ tap -A --set agent-image=docker.io/up9inc/mizu:develop
```

or this is a pre-release which CLI and the agent image is built by CI;

https://github.com/up9inc/mizu/releases/tag/0.22.33


shared/consts.go
```diff
@@ -16,6 +16,6 @@ const (
 	LogLevelEnvVar                   = "LOG_LEVEL"
 	BasenineHost                     = "127.0.0.1"
 	BaseninePort                     = "9099"
-	BasenineImageRepo                = "ghcr.io/up9inc/basenine"
-	BasenineImageTag                 = "v0.3.0"
+	BasenineImageRepo                = "docker.io/up9inc/basenine"
```
gadotroee, line 19: any reason to specify "docker.io", I think this can be just "up9inc/basenine"..

mertyildiran: @gadotroee it's better be `docker.io` though. To clarify it's Docker Hub.

gadotroee: Ok, I thought it might be a best practice

mertyildiran: @gadotroee if a relative path to the default registry has any use case, like a setup that alters the default registry to a mirror. Then we can make it;

```suggestion
	BasenineImageRepo                = "up9inc/basenine"
```

Is there such a use case?

gadotroee: I don't think so, probably better to be specific in this case

gadotroee:
I'm waiting to hear from Alex approval of this solution to be working and then I will be able to approve. the code looks ok but seems that building the arm1 is not part of this (because we need to do it with matrix and build on the correct machine). so I need to understand if this can be merged and adding it later or maybe wait with making the build too in this branch

devops/build-push-featurebranch.sh
```diff
@@ -8,6 +8,8 @@ GIT_BRANCH=$(git branch | grep \* | cut -d ' ' -f2 | tr '[:upper:]' '[:lower:]')
 
 DOCKER_REPO=$REPOSITORY/$SERVER_NAME/$GIT_BRANCH
 SEM_VER=${SEM_VER=0.0.0}
+ARCH=amd64
```
gadotroee, line 11: maybe can be passed as param with default (like SEM_VER)
or completely take from user env (so it will work by default for "arm" users)

mertyildiran: @gadotroee I'm not a user of this script and I don't really understand its purpose. Feel free to modify according to your development workflow.

mertyildiran:
We can probably distribute `gopacket` for AMD64 and ARM64 as a [binary-only package](https://pkg.go.dev/go/build#hdr-Binary_Only_Packages). Hence this will enable us to cross-compile on an AMD64 machine.

mertyildiran:
oh it's not supported anymore :disappointed:
> Go 1.13: Binary only packages are not supported anymore

mertyildiran:
At this point;
- We have a working perfectly fine `debian:buster` based builder image.
- We have a working perfectly fine `debian:buster-slim` based shipped image.
- All with cross-compilation.

Now I'll try to achieve the same thing in Alpine.

mertyildiran:
Alpine does not provide GCC cross-compiler toolchain. There is `/usr/bin/aarch64-alpine-linux-musl-gcc` in [here](https://pkgs.alpinelinux.org/contents?branch=edge&name=gcc&arch=aarch64&repo=main) but it's not available for `amd64`. So we need to compile GCC from source: https://preshing.com/20141119/how-to-build-a-gcc-cross-compiler/

or we can keep it `debian:buster-slim` based. (means that we switch from `alpine:3.15` to `debian:buster-slim` as base)

mertyildiran:
The base of the shipped image is `debian:stable-slim` now.

cli/README.md.TEMPLATE
```diff
@@ -3,22 +3,27 @@ Full changelog for stable release see in [docs](https://github.com/up9inc/mizu/b
...
 ## Download Mizu for your platform
 
-**Mac** (Intel)
+**Mac** (x86-64/Intel)
 ```
 curl -Lo mizu https://github.com/up9inc/mizu/releases/download/_SEM_VER_/mizu_darwin_amd64 && chmod 755 mizu
 ```
 
-**Mac** (Apple M1 silicon)
+**Mac** (AArch64/Apple M1 silicon)
```
IgorGov, line 11:
```suggestion
**Mac** (Arch64/Apple M1 silicon)
```

Makefile
```diff
@@ -71,7 +69,7 @@ push-cli: ## Build and publish CLI.
 	gsutil cp -r ./cli/bin/* gs://${BUCKET_PATH}/
 	gsutil setmeta -r -h "Cache-Control:public, max-age=30" gs://${BUCKET_PATH}/\*
 
-clean: clean-ui clean-agent clean-cli clean-docker clean-extensions ## Clean all build artifacts.
```
IgorGov, line 74: i see you removed the docker debug, so we need 
1. to clean the make file from its reference like: "push-docker-debug"
2. remove build-push-featurebranch-debug.sh

Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM node:16-slim AS site-build
+ARG BUILDARCH=amd64
```
IgorGov, line 1: it duplicated in line 43, do we still need it here as well?

Makefile
mertyildiran: @IgorGov I have actually removed `devops/build-push-featurebranch-debug.sh` as well but I forgot to remove the `push-docker-debug` rule. Done with https://github.com/up9inc/mizu/commit/9b2f01a84237bd7f53d0d893c155ed493819a416

.github/workflows/publish.yml
```diff
@@ -1,4 +1,4 @@
-name: publish
+name: Release
```
IgorGov, line 1: we should rename the file as well

Dockerfile
mertyildiran: @IgorGov `ARG`s are scoped in Dockerfile. For each `FROM` statement they are refreshed. Therefore they need to be redeclared.

.github/workflows/publish.yml
mertyildiran: @IgorGov done with https://github.com/up9inc/mizu/commit/ca55ba976de296c01e35dccd21e01c231de83dd1

.github/workflows/publish.yml
```diff
@@ -11,62 +11,146 @@ concurrency:
...
+          labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            TARGETARCH=${{ matrix.target }}
+            SEM_VER=${{ steps.versioning.outputs.version }}
+            BUILD_TIMESTAMP=${{ steps.version_parameters.outputs.build_timestamp }}
+            GIT_BRANCH=${{ steps.version_parameters.outputs.branch }}
+            COMMIT_HASH=${{ github.sha }}
+
+  gcp-registry:
+    name: Push Docker image to GCR
```
IgorGov: why is it a separate step from "Push Docker image to Docker Hub"?

.github/workflows/publish.yml
```diff
@@ -75,16 +159,115 @@ jobs:
...
+      - name: Determine versioning strategy
+        uses: haya14busa/action-cond@v1
+        id: condval
+        with:
+          cond: ${{ github.ref == 'refs/heads/main' }}
+          if_true: "minor"
+          if_false: "patch"
+
+      - name: Auto increment SemVer action
+        uses: MCKanpolat/auto-semver-action@1.0.5
```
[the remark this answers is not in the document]
mertyildiran: @IgorGov yeah we can have a job that decides the version through this GitHub action and can generate an artifact. But I'm not good at artifacts :slightly_smiling_face:

.github/workflows/publish.yml
mertyildiran: @IgorGov to achieve separation of concerns and a clean workflow.

IgorGov: but you get a duplicate code

haiut (approved):

Dockerfile
IgorGov: got you

PR #667, opened 2022-01-19
TRA-4158 login touch ups

AmitUp9:


ui/src/index.tsx
```diff
@@ -7,6 +7,11 @@ import {ToastContainer} from "react-toastify";
 import 'react-toastify/dist/ReactToastify.css';
 import {RecoilRoot} from "recoil";
 
+if (window['isEnt'] === undefined) {
```
lirazyehezkel, line 10: what is this code?
I think we don't want to push it

PR #681, opened 2022-01-23
Split UI build

lirazyehezkel:


ui/.env.enterprise
```diff
@@ -0,0 +1,3 @@
+REACT_APP_OVERRIDE_WS_URL="ws://localhost:8899/ws"
+REACT_APP_OVERRIDE_API_URL="http://localhost:8899/"
+REACT_APP_OVERRIDE_IS_ENTERPRISE="true"
```
RamiBerm, line 3: awesome

ui/package.json
```diff
@@ -43,11 +43,15 @@
...
   },
+  "devDependencies": {
+    "env-cmd": "^10.0.1"
+  },
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "start": "craco start",
```
RamiBerm, line 50: isnt it supposed to be `./node_modules/.bin/env-cmd -f .env.basic craco start`?

ui/src/AppChooser.tsx
```diff
@@ -0,0 +1,18 @@
+import React, {Suspense} from 'react';
+import LoadingOverlay from "./components/LoadingOverlay";
+
+const AppChooser = () => {
+
+    let MainComponent;
+    if (process.env.REACT_APP_OVERRIDE_IS_ENTERPRISE === "true") {
```
RamiBerm, line 7: should you also check for `window['isEnt']`?

RamiBerm (changes requested):

ui/package.json
lirazyehezkel: so I'm not sure actually because we have the thing with Mert script which copy env.example to .env
I think we can leave it like that for now and maybe change it after if we will be needed

ui/src/AppChooser.tsx
lirazyehezkel: Nope - this split should be a replacement for this
but this is a good point because I can to remove this variable from the HTML file

RamiBerm (approved):

PR #684, opened 2022-01-24
TRA-4205 React router

lirazyehezkel:


ui/src/components/AppSwitchRoutes.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const AppSwitchRoutes = () => {
+
+    const navigate = useNavigate();
+
+    const [isLoading, setIsLoading] = useState(true);
+    const [entPage, setEntPage] = useRecoilState(entPageAtom);
+    const [isFirstLogin, setIsFirstLogin] = useState(false);
+
+    const determinePage =  useCallback(async () => { // TODO: move to state management
```
RamiBerm, line 25: can the `TODO: move to state management` be removed now?

ui/src/components/AppSwitchRoutes.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+
+    if (isLoading) {
+        return <LoadingOverlay/>;
+    }
+
+    return <Routes>
+        <Route path={"/"} element={<SystemViewer isFirstLogin={isFirstLogin} setIsFirstLogin={setIsFirstLogin}/>}>
+            <Route path={"/"} element={<TrafficPage/>} />
+            <Route path={RouterRoutes.SETTINGS} element={<></>} /> {/*todo: set settings component*/}
```
RamiBerm, line 73: why do we need this path configured right now?

RamiBerm (changes requested):

ui/src/components/AppSwitchRoutes.tsx
lirazyehezkel: Leon / Amit should add it right after this merge, so I thought it will be easier if I'll configure it already

ui/src/components/AppSwitchRoutes.tsx
lirazyehezkel: sure, removing

RamiBerm (approved):

PR #688, opened 2022-01-24
TRA-4202 role management

RamiBerm:


agent/kratos/kratos.yml
```diff
@@ -57,7 +57,7 @@ selfservice:
 log:
   level: info
   format: text
-  leak_sensitive_values: true
+  leak_sensitive_values: false
```
IgorGov, line 60: is it related to this change?

agent/pkg/api/socket_routes.go
```diff
@@ -49,7 +48,7 @@ func init() {
 func WebSocketRoutes(app *gin.Engine, eventHandlers EventHandlers, startTime int64) {
 	app.GET("/ws", func(c *gin.Context) {
 		websocketHandler(c.Writer, c.Request, eventHandlers, false, startTime)
-	}, middlewares.RequiresAuth())
+	})
```
IgorGov, line 51: is it related to bug that client doesn't send the token in custom header? 
how did it worked before?

agent/pkg/routes/config_routes.go
```diff
@@ -1,15 +1,16 @@
...
+
+	"github.com/gin-gonic/gin"
 )
 
 func ConfigRoutes(ginApp *gin.Engine) {
 	routeGroup := ginApp.Group("/config")
 	routeGroup.Use(middlewares.RequiresAuth())
 
-	routeGroup.POST("/tapConfig", controllers.PostTapConfig)
+	routeGroup.POST("/tapConfig", middlewares.RequiresAdmin(), controllers.PostTapConfig)
```
IgorGov, line 14: should we rename it to just "tap"? so there will be no duplicate "config" in endpoint:
/config/tapConfig -> /config/tap

IgorGov: another thought we should rename it to tappedNamespace (we have different "tap" endpoints some for tapPods status, some for tapper status..)

agent/kratos/kratos.yml
RamiBerm: no but I noticed this is set to true and we probably dont want it that way

agent/pkg/routes/config_routes.go
RamiBerm: renamed to `/tap`

agent/pkg/providers/user_role_provider.go
```diff
@@ -0,0 +1,180 @@
+package providers
```
IgorGov, line 1: so we won't end up with a package with too many providers, I suggest start splitting them:
"user role provider"
"user provide"

agent/pkg/providers/install_provider.go
```diff
@@ -16,3 +21,27 @@ func IsInstallNeeded() (bool, error) {
 		return !anyUserExists, nil
 	}
 }
+
+func CreateAdminUser(password string, ctx context.Context) (token *string, err error, formErrorMessages map[string][]ory.UiText) {
```
IgorGov, line 25: can we move this file content to user provider?

agent/pkg/providers/user_provider.go
```diff
@@ -66,17 +66,17 @@ func PerformLogin(username string, password string, ctx context.Context) (*strin
 	return result.SessionToken, nil
 }
 
-func VerifyToken(token string, ctx context.Context) (bool, error) {
+func VerifyToken(token string, ctx context.Context) (*ory.Session, error) {
```
IgorGov, line 69: can we rename the package name to "user provider"?

agent/pkg/controllers/install_controller.go
```diff
@@ -16,3 +16,8 @@ func IsSetupNecessary(c *gin.Context) {
 		c.JSON(http.StatusOK, IsInstallNeeded)
 	}
 }
+
+func SetupAdminUser(c *gin.Context) {
+	token, err, formErrorMessages := providers.CreateAdminUser(c.PostForm("password"), c.Request.Context())
```
IgorGov, line 21:
```suggestion
	token, err, formErrorMessages := userProvider.CreateAdmin(c.PostForm("password"), c.Request.Context())
```

IgorGov: see the comment for splitting to dedicated providers below..

IgorGov (approved):

PR #713, opened 2022-01-28
Experimental feature: elastic exporter

IgorGov:


shared/models.go
```diff
@@ -43,6 +43,12 @@ type MizuAgentConfig struct {
 	StandaloneMode         bool          `json:"standaloneMode"`
 	ServiceMap             bool          `json:"serviceMap"`
 	OAS                    bool          `json:"oas"`
+	Elastic                ElasticConfig `json:"elastic"`
+}
+
+type ElasticConfig struct {
+	CloudID string `yaml:"cloud-id,omitempty" default:"" readonly:""`
```
gadotroee, line 50: Better to support elastic urls (it will support cloud and not cloud)

gadotroee (commented):
Added one comment

cli/config/configStruct.go
```diff
@@ -27,7 +27,7 @@ type ConfigStruct struct {
 	Logs                   configStructs.LogsConfig    `yaml:"logs"`
 	Auth                   configStructs.AuthConfig    `yaml:"auth"`
 	Config                 configStructs.ConfigConfig  `yaml:"config,omitempty"`
-	AgentImage             string                      `yaml:"agent-image,omitempty" readonly:""`
```
gadotroee, line 30: why the omitempty and readonly removed

cli/cmd/common.go
```diff
@@ -35,7 +35,7 @@ func startProxyReportErrorIfAny(kubernetesProvider *kubernetes.Provider, ctx con
 		return
 	}
 
-	apiProvider = apiserver.NewProviderWithoutRetries(GetApiServerUrl(), time.Second) // short check for proxy
```
gadotroee, line 38: why do we need 3 retries.. 
I remember that I implement the NewProviderWithoutRetries and there was a reason for it (and now if you are removing this (which is the only place that calls this method - it should be removed)

agent/pkg/elastic/esClient.go
```diff
@@ -0,0 +1,120 @@
...
+	insertedCount int
+}
+
+var instance *client
+var once sync.Once
+
+func GetInstance() *client {
+	once.Do(func() {
+		instance = newClient()
+		logger.Log.Debug("Service Map Initialized")
```
gadotroee, line 28:
```suggestion
		logger.Log.Debug("ElasticSearch client Initialized")
```

agent/pkg/elastic/esClient.go
```diff
@@ -0,0 +1,120 @@
...
+	Response    map[string]interface{} `json:"response"`
+	Summary     string                 `json:"summary"`
+	Method      string                 `json:"method"`
+	Status      int                    `json:"status"`
+	ElapsedTime int64                  `json:"elapsedTime"`
+	Path        string                 `json:"path"`
+}
+
+func (client *client) PushEntry(entry *api.Entry) {
+	if client.es == nil {
```
gadotroee, line 87: I understand that this line will ignore every entry so it might be ok to always initialize.
 but still it feel weird from outside

gadotroee (approved):

PR #714, opened 2022-01-29
Fixing minor bugs and remove unused dependency

IgorGov:


cli/cmd/tapRunner.go
```diff
@@ -405,7 +405,7 @@ func watchApiServerEvents(ctx context.Context, kubernetesProvider *kubernetes.Pr
 			case "FailedScheduling", "Failed":
 				logger.Log.Errorf(uiUtils.Error, fmt.Sprintf("Mizu API Server status: %s - %s", event.Reason, event.Note))
 				cancel()
-				break
```
gadotroee, line 408: why the break here removed? 
don't we want to stop the loop after it "FailedScheduling" or "Failed"

cli/cmd/checkRunner.go
```diff
@@ -205,13 +206,13 @@ func checkInstallResourcesExist(ctx context.Context, kubernetesProvider *kuberne
 	installResourcesExist := checkResourceExist(kubernetes.DaemonRoleName, "role", exist, err)
 
 	exist, err = kubernetesProvider.DoesRoleBindingExist(ctx, config.Config.MizuResourcesNamespace, kubernetes.DaemonRoleBindingName)
-	installResourcesExist = checkResourceExist(kubernetes.DaemonRoleBindingName, "role binding", exist, err)
+	installResourcesExist = installResourcesExist && checkResourceExist(kubernetes.DaemonRoleBindingName, "role binding", exist, err)
```
gadotroee, line 209: I'm little confused from this line,
the idea is to not check the current resource and if the previous is not exists?

gadotroee: so from the moment that something not exists you will not query for more resources?

gadotroee (approved):

PR #725, opened 2022-01-31
Build agent docker image during CI

IgorGov:


Makefile
```diff
@@ -40,17 +40,18 @@ agent: ## Build agent.
 	@ls -l agent/build
 
 agent-debug: ## Build agent for debug.
-	@(echo "building mizu agent for debug.." )
+	@(echo "uilding mizu agent for debug.." )
```
gadotroee, line 43: mistake?

.github/workflows/release.yml
```diff
@@ -82,91 +82,8 @@ jobs:
             GIT_BRANCH=${{ steps.version_parameters.outputs.branch }}
             COMMIT_HASH=${{ github.sha }}
 
-  gcp-registry:
```
gadotroee, line 85: where do we add the GCR?

IgorGov: lets discuss it f2f

Makefile
IgorGov: fixed

RoyUP9 (approved):

PR #733, opened 2022-01-31
TRA-4169 apply query by enter

leon-up9:


ui/src/hooks/useKeyPress.ts
```diff
@@ -10,8 +10,12 @@ const useKeyPress = (eventConfigs, callback, node = null) => {
   // handle what happens on key press
   const handleKeyPress = useCallback(
     (event) => {
+      event.preventDefault();
```
lirazyehezkel, line 13: why do you need this twice?
and are you sure you need this at all? I think maybe stopPropagation should be enough in this case

lirazyehezkel (approved):

PR #734, opened 2022-01-31
Technical depth: Adding Go linter to CI

IgorGov:


gadotroee (approved):

mertyildiran:
I think we forgot to include the modules that are located under;
- `tap/api`
- `tap/extensions/amqp`
- `tap/extensions/http`
- `tap/extensions/kafka`
- `tap/extensions/redis`

mertyildiran:
Also why is it triggered on `pull_request` only and certain branches? I think we can also include the `push`. Since the job is finishing ~3 minutes which is not so long.

mertyildiran:
I think the PR title has a typo. I guess it should be "**Technical debt: Adding Go linter to CI**"

PR #742, opened 2022-02-01
Renaming mizuserver module

IgorGov:


IgorGov:
> looking like a good change (I know it is best practice to work with modules that points to absolute location (github.com...)
it is a preparation for mizu ent

gadotroee (approved):

PR #751, opened 2022-02-02
[TRA-4197] Integration Tests for PCAP files

gustavomassa:


integrationTests/agent.sh
```diff
@@ -0,0 +1,52 @@
...
+
+if [[ $2 != *.cap ]]
+  then echo "PCAP file must be a .cap file!"
+  exit
+fi
+
+PCAP=$PWD/$2
+echo "PCAP file path: $PCAP"
+
+rm -rf entries/ && mkdir -p entries && rm -rf pprof/* && \
```
nimrod-up9, line 34: I assume that you concatenate the commands with `&&` because you want the script to break if one of the commands returns an error.
Instead, add `set -e` at the top of the file, below the hash bang.
It causes bash to exit immediately if a command fails.

integrationTests/agent.sh
```diff
@@ -0,0 +1,52 @@
...
+GOGC=12800 NODE_NAME=dev \
+    ./agent/build/mizuagent \
+        -r $PCAP \
+        --tap \
+        --api-server-address ws://localhost:8899/wsTapper & \
+PID3=$! && \
+read -r -d '' _ </dev/tty
+kill -9 $PID1 && \
+kill -9 $PID2 && \
+kill -9 $PID3
```
nimrod-up9, line 52: Why not SIGTERM and allow a graceful shutdown?

nimrod-up9 (commented):
Added some comments. Haven't finished reviewing all of the files.

integrationTests/agent.sh
gustavomassa: I'm going to remove this script, this is not being used anymore. I started the development using this script, then switched to starting basenine/api-server/tapper directly from the go test code using exec.command.

gadotroee:
similar idea of testing PCAP files already merged in https://github.com/up9inc/mizu/pull/767
so I will close this

PR #754, opened 2022-02-04
Disable DWARF and symbol table generation for CLI and build it statically

mertyildiran:
The size of the binary;

```
Old: 51 MB
-static only: 46 MB
-static + -s -w: 37 MB
```

`-s -w` does not strip the stack trace.

Purpose is avoiding any possible linker errors and reducing the size of the binary.


haiut:
@mertyildiran please keep the "static", but I don't see any need to strip the symbol table (remove -s -w)
symbol table will definitely help us to debug crash if it happens

mertyildiran:
Updated the PR description.

gadotroee (approved):

PR #755, opened 2022-02-05
Moving from Semver to major.patch versioning

IgorGov:
Example of how versioning will look:
https://github.com/IgorGov/mizu/releases
image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,14 +32,15 @@ jobs:
         id: condval
         with:
           cond: ${{ github.ref == 'refs/heads/main' }}
-          if_true: "minor"
-          if_false: "patch"
+          if_true: "stable"
+          if_false: "dev"
 
-      - name: Auto increment SemVer action
-        uses: MCKanpolat/auto-semver-action@1.0.5
+      - name: Auto Increment Ver Action
+        uses: docker://igorgov/auto-inc-ver:v2.0.0
         id: versioning
         with:
-          releaseType: ${{ steps.condval.outputs.value }}
+          mode: ${{ steps.condval.outputs.value }}
+          suffix: 'dev'
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Get version parameters
@@ -109,14 +110,15 @@ jobs:
         id: condval
         with:
           cond: ${{ github.ref == 'refs/heads/main' }}
-          if_true: "minor"
-          if_false: "patch"
+          if_true: "stable"
+          if_false: "dev"
 
-      - name: Auto increment SemVer action
-        uses: MCKanpolat/auto-semver-action@1.0.5
+      - name: Auto Increment Ver Action
+        uses: docker://igorgov/auto-inc-ver:vv2.0.0
         id: versioning
         with:
-          releaseType: ${{ steps.condval.outputs.value }}
+          mode: ${{ steps.condval.outputs.value }}
+          suffix: 'dev'
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Get version parameters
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -242,14 +242,15 @@ jobs:
         id: condval
         with:
           cond: ${{ github.ref == 'refs/heads/main' }}
-          if_true: "minor"
-          if_false: "patch"
+          if_true: "official"
```
nimrod-up9, line 245: Rename to "stable"? The dev versions are also official.

gadotroee:
I converted it to draft because two things need to be chnaged and it is not ready to for merge.
1. semver class https://github.com/up9inc/mizu/blob/develop/shared/semver/semver.go (it will not support those new versions format)
2. all calls to Auto Increment should use the new step.

cli/mizu/version/versionCheck.go
```diff
@@ -12,31 +12,30 @@ import (
 
 	"github.com/up9inc/mizu/cli/apiserver"
 	"github.com/up9inc/mizu/cli/mizu"
+	"github.com/up9inc/mizu/cli/pkg/version"
 	"github.com/up9inc/mizu/shared/logger"
 
 	"github.com/google/go-github/v37/github"
 	"github.com/up9inc/mizu/cli/uiUtils"
-	"github.com/up9inc/mizu/shared/semver"
```
RoyUP9, line 19: is this in any use anymore?

cli/mizu/version/versionCheck.go
```diff
@@ -12,31 +12,30 @@ import (
...
-	apiSemVer, err := apiServerProvider.GetVersion()
+	apiVer, err := apiServerProvider.GetVersion()
 	if err != nil {
 		return false, err
 	}
 
-	if !semver.SemVersion(apiSemVer).IsValid() {
-		logger.Log.Errorf(uiUtils.Red, fmt.Sprintf("api version (%s) is not a valid SemVer", apiSemVer))
-		return false, nil
+	equals, err := version.AreEquals(apiVer, mizu.Ver)
```
RoyUP9, line 28: can be changed to one liner with if and else if

cli/mizu/version/versionCheck.go
IgorGov: yes, k8s versions check

cli/pkg/version/version.go
```diff
@@ -0,0 +1,81 @@
...
+	major, err := strconv.Atoi(match[1])
+	if err != nil {
+		return nil, fmt.Errorf("error parsing major int: %s, err %w", match[1], err)
+	}
+	patch, err := strconv.Atoi(match[2])
+	if err != nil {
+		return nil, fmt.Errorf("error parsing patch int: %s, err %w", match[2], err)
+	}
+
+	if match[3] == "" {
```
RoyUP9, line 30: maybe change to check to len(match) == 3?

cli/pkg/version/version_test.go
```diff
@@ -0,0 +1,81 @@
+package version
+
+import (
+	"testing"
+)
+
+func TestEqualsEquality(t *testing.T) {
+	first, second := "1.0", "1.0"
```
RoyUP9, line 8: you can use test cases and pass it different version checks each time

cli/pkg/version/version.go
IgorGov: it will always will be 4

cli/pkg/version/version_test.go
```diff
@@ -5,77 +5,100 @@ import (
...
-		t.Fatalf("Expected %s > %s", first, second)
+	tests := []struct {
+		Name   string
+		First  string
+		Second string
+	}{
+		{Name: "major", First: "2.0", Second: "1.0"},
+		{Name: "patch", First: "1.1", Second: "1.0"},
+		{Name: "incremental", First: "1.0-dev1", Second: "1.0-dev0"},
+		{Name: "major", First: "1.0", Second: "1.0-dev1"},
```
RoyUP9, line 73: i think you can't have 2 test cases with the same name

cli/pkg/version/version_test.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			if greater, _ := GreaterThen(test.First, test.Second); !greater {
+				t.Fatalf("Expected %s > %s", test.First, test.Second)
+			}
+		})
+	}
+}
+
+func TestGreaterThenLessthen(t *testing.T) {
```
RoyUP9, line 84: Rename to TestGreaterThenLessThen

RoyUP9 (approved):

PR #766, opened 2022-02-07
Moving to go 1.17

IgorGov:
Why so many indirect modules were added?
[Go 1.17 release notes](https://go.dev/doc/go1.17) 

> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.

### DIFF
--- a/tap/extensions/amqp/go.mod
+++ b/tap/extensions/amqp/go.mod
@@ -4,4 +4,6 @@ go 1.17
 
 require github.com/up9inc/mizu/tap/api v0.0.0
 
+require github.com/google/martian v2.1.0+incompatible // indirect
+
 replace github.com/up9inc/mizu/tap/api v0.0.0 => ../../api
--- a/tap/extensions/amqp/go.mod
+++ b/tap/extensions/amqp/go.mod
@@ -1,6 +1,6 @@
 module github.com/up9inc/mizu/tap/extensions/amqp
 
-go 1.16
+go 1.17
 
 require github.com/up9inc/mizu/tap/api v0.0.0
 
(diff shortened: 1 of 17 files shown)


mertyildiran:
I guess this image requires an upgrade as well https://github.com/up9inc/mizu/blob/0365b8971f6bc4743d7bf223ede668f91dd257fe/devops/linux-arm64-musl-go-libpcap/Dockerfile#L4-L5

IgorGov:
> I guess this image requires an upgrade as well

Good catch! Thanks @mertyildiran

mertyildiran:
> > I guess this image requires an upgrade as well
> 
> Good catch! Thanks @mertyildiran

That image also requires a build and push on an ARM64 machine after the merge.

IgorGov:
> That image also requires a build and push on an ARM64 machine after the merge.

done

tap/extensions/amqp/go.mod
```diff
@@ -1,7 +1,9 @@
 module github.com/up9inc/mizu/tap/extensions/amqp
 
-go 1.16
+go 1.17
 
 require github.com/up9inc/mizu/tap/api v0.0.0
 
+require github.com/google/martian v2.1.0+incompatible // indirect
```
gadotroee, line 7: why does the AMQP has reference to martian?

gadotroee: I guess it is because the "tap/api" module has maritan?

IgorGov: yep

gadotroee (approved):

PR #749, opened 2022-02-02
Add a PCAP based testbed

mertyildiran:
Adds a test script (consists of Bash and Python) that downloads a bunch of PCAP files from a Google Cloud bucket, runs the tapper against the PCAP files and generates data file named `suite.json` to lock the state.

Run `make test-run-update` to update `suite.json` and run `make test-run` to validate against that file.

Adds `test-lint` Makefile rule to run linter checks locally with ease.

Improves request-response matching logic of HTTP and Redis.

Fixes some interface conversion errors in AMQP and Kafka that's found through those PCAP files.

Introduces an environment variable named `MIZU_TEST` to alter the certain default behavior of Mizu agent, in the test environment to achieve more consistent behavior.

The testing approach that's introduced by this PR, is planned to be included into the CI in the later PRs.

PR #767, opened 2022-02-07
Add unit tests for HTTP dissector

mertyildiran:
Adds unit tests to the HTTP dissector. The tests download pre-made binary request-response pairs from a Google Cloud bucket. Then tests against expected JSONs which is again downloaded from a Google Cloud bucket.

The tests increases the code coverage of `tap/extensions/http` from 0% to 83.40%.

`make test` runs the tests.

`make test-update` updates the expected JSONs according to the binary request-response pairs. But it does not upload them to Google Cloud. You need to do it manually.

The PR introduces sorting of several inconsistent fields like `Headers`, `QueryString` etc. to make them testable.

It also erases some fields in case of the test enviroment (`MIZU_TEST`) to make it testable.


.github/workflows/test.yml
```diff
@@ -27,19 +27,31 @@ jobs:
...
           sudo apt-get install libpcap-dev
 
+      - name: Install gsutil
+        shell: bash
+        run: |
+          echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+          curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+          sudo apt-get update
+          sudo apt-get install google-cloud-sdk
+
```
gadotroee, line 42:
```suggestion
      - id: 'auth'
        uses: 'google-github-actions/auth@v0'
        with:
          credentials_json: '${{ secrets.GCR_JSON_KEY }}'

      - name: 'Set up Cloud SDK'
        uses: 'google-github-actions/setup-gcloud@v0'
```

gadotroee: the auth part can be removed if the bucket can be access anonymously, but I think those actions are better because we already use them in other workflows..
I'm ok with installing it like this but let me know wdyt.

mertyildiran: @gadotroee oh there is an action for that right? Yeah. But we don't need authentication.

tap/extensions/http/main_test.go
```diff
@@ -0,0 +1,312 @@
...
+	dissector := NewDissector()
+	dissector.Ping()
+}
+
+func TestDissect(t *testing.T) {
+	var testUpdateEnabled bool
+	_, present := os.LookupEnv(testUpdate)
+	if present {
+		testUpdateEnabled = true
+	}
```
gadotroee, line 66:
```suggestion
	_, testUpdateEnabled := os.LookupEnv(testUpdate)
```

gadotroee: I see it in the code and I think it is doing the same thing..

tap/extensions/http/main_test.go
```diff
@@ -0,0 +1,312 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	options := &api.TrafficFilteringOptions{
+		IgnoredUserAgents: []string{},
+	}
+
+	for _, _path := range paths {
+		basePath := _path[:len(_path)-8]
```
gadotroee, line 87: what is this "8"? length of something that I miss?

gadotroee: probably the length of this: "_req.bin" (just to be sure I understand the logic..)

tap/extensions/http/main_test.go
mertyildiran: @gadotroee done with https://github.com/up9inc/mizu/commit/4b6b07bf26df330d014316eaad35b7b83b22067a

tap/extensions/http/main_test.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+}
+
+func TestAnalyze(t *testing.T) {
+	var testUpdateEnabled bool
+	_, present := os.LookupEnv(testUpdate)
+	if present {
+		testUpdateEnabled = true
+	}
```
gadotroee:
```suggestion
	_, testUpdateEnabled := os.LookupEnv(testUpdate)
```

.github/workflows/test.yml
mertyildiran: Anyway I committed it with auth.

tap/extensions/http/main_test.go
mertyildiran: @gadotroee yes it's the length of `_req.bin`.

tap/extensions/http/main_test.go
mertyildiran: @gadotroee this is outdated.

tap/extensions/http/Makefile
```diff
@@ -0,0 +1,16 @@
+skipbin := $$(ls -A --ignore='.??*' bin)
```
gadotroee, line 1: I understand that those options of --ignore are not exists in mac 
(https://github.com/up9inc/mizu/pull/749#discussion_r801828585) 
can we change those to something that will be supported in both mac and linux so developers will be able to use it on their laptop? 
(maybe "find")?

mertyildiran: @gadotroee done with https://github.com/up9inc/mizu/commit/380390d6d05fdfdf220bb4b4e7c3bf919cda1049

gadotroee (approved):

PR #772, opened 2022-02-08
Update Go dependencies to latest and greatest

IgorGov:


IgorGov:
Need to check what should we do with Snyk -> Converted to draft

gadotroee (approved):

Issue #706, opened 2022-01-27
Running mizu fails silently with no logs


peter-dolkens:
**Describe the bug**

**To Reproduce**
1. Run `mizu tap` or any variant of it which does NOT include `mizu-resources-namespace` and `--namespaces`
2. See mizu terminate instantly with no logs or error output
3. Run `mizu tap --set mizu-resources-namespace=my-ns --namespaces=my-ns`
4. Mizu executes as expected
5. Run `mizu tap --set mizu-resources-namespace=mizu --namespaces=mizu`
6. See mizu terminate instantly with no logs or error output

I'm admin on this cluster, and have even gone so far as to create the `mizu` namespace ahead of time after it failed originally.

**Expected behavior**

`mizu tap` command works without `mizu-resources-namespace` and `--namespaces` command

**Logs**

WARNING: No zip logs generated, only CLI logs.

**Screenshots**

n/a

**Desktop (please complete the following information):**
 - OS: Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64
 - Web Browser: n/a

**Additional context**

We do have an OPA policy which enforces a naming scheme on our namespaces unless a certain label is added. As such, I tried adding the `mizu` namespace manually with the `partner: core` label.

Additionally, the mac install instructions don't add mizu to your path, so I manually copied it to `/usr/local/bin/mizu` though the issue was happening even when executing in my home directory with `./mizu` instead of just `mizu`

```
➜  ~ rm -rf .mizu
```

```
➜  ~ mizu tap --set mizu-resources-namespace=mizu --namespaces=mizu --set dump-logs=true
Mizu will store up to 200MB of traffic, old traffic will be cleared once the limit is reached.
Tapping pods in namespaces "mizu"
+bash
Waiting for Mizu Agent to start...
```

```
➜  ~ cat .mizu/mizu_cli.log
[2022-01-27T02:46:07.633+0000] DEBUG ▶ Checking for newer version... ▶ [57108 versionCheck.go:47 CheckNewerVersion]
[2022-01-27T02:46:07.633+0000] DEBUG ▶ Init config finished
 Final config: {
        "Tap": {
                "UploadIntervalSec": 10,
                "PodRegexStr": ".*",
                "GuiPort": 8899,
                "ProxyHost": "127.0.0.1",
                "Namespaces": [
                        "mizu"
                ],
                "Analysis": false,
[...]

PR #776, opened 2022-02-08
Log permission error in resource creation

nimrod-up9:
Fixes https://github.com/up9inc/mizu/issues/706.
`mizu tap` and `mizu install` fail silently if Mizu fails the required permissions to create its k8s resources.


gadotroee (approved):
indeed looks like a mistake, because any error that was not already exists was not written..

PR #777, opened 2022-02-08
Add `test-lint` Makefile rule

mertyildiran:
For local linting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,15 @@ clean-cli:  ## Clean CLI.
 clean-docker:
 	@(echo "DOCKER cleanup - NOT IMPLEMENTED YET " )
 
+test-lint:
+	cd agent && golangci-lint run
+	cd shared && golangci-lint run
+	cd tap && golangci-lint run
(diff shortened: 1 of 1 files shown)


mertyildiran:
Split from https://github.com/up9inc/mizu/pull/749

gadotroee (approved):
I understand you took it from the place you did just to make it available before.

gadotroee:
> I understand you took it from the place you did just to make it available before.

nvm, I see now the message you wrote..

Makefile
```diff
@@ -83,6 +83,15 @@ clean-cli:  ## Clean CLI.
 clean-docker:
 	@(echo "DOCKER cleanup - NOT IMPLEMENTED YET " )
 
+test-lint:
```
IgorGov, line 86: i would call it "go-lint" as it not running FE linter, and not to confuse it with "tests"

mertyildiran: oh we can run the ESLint as well. `test-` is the prefix for the rules that are related to testing. Linting is a test as well.

PR #792, opened 2022-02-10
Print http error response details

nimrod-up9:
When CLI gets a response from agent with status != 200, it prints:

```
[Error] failed update tapper status failed sending to API server the tapper status, response status code 403 ▶ [14476 tapRunner.go:235 func1]
```

The above message only includes the status code. It does not include any details contained in the response body.
This PR adds the response body to the error message.
After this PR, the log shows:

```
[Error] failed update tapper status failed sending to API server the tapped pods got response with status code 403 {"kind":"Status","apiVersion":"v1","metad
ata":{},"status":"Failure","message":"services \"mizu-api-server:80\" is forbidden: User \"user1\" cannot create resource \"services/proxy\" in API group \"\" in the namespace \"mizu\"","reason":"For
bidden","details":{"name":"mizu-api-server:80","kind":"services"},"code":403}
```


cli/apiserver/provider.go
```diff
@@ -145,3 +139,33 @@ func (provider *Provider) GetVersion() (string, error) {
...
+	return provider.checkError(provider.client.Do(req))
+}
+
+func (provider *Provider) checkError(response *http.Response, errInOperation error) (*http.Response, error) {
+	if (errInOperation != nil) {
+		return response, errInOperation
+	// Check only if status != 200 (and not status >= 300). Agent APIs return only 200 on success.
+	} else if response.StatusCode != http.StatusOK {
+		body, err := ioutil.ReadAll(response.Body)
+		response.Body = io.NopCloser(bytes.NewBuffer(body)) // rewind
```
RoyUP9, line 161: Are you creating new reader here? if so, do we need to close the old reader?

gadotroee (approved):

PR #778, opened 2022-02-08
Improve request-response matching and fix some interface conversion errors

mertyildiran:
Improves request-response matching in HTTP and Redis.

Fixes some interface conversion errors in AMQP and Kafka.

PR #793, opened 2022-02-10
Move the request-response matcher's scope from global-level to TCP stream-level

mertyildiran:
Isolates request-response matching logic to the individual TCP streams. This enables us to verify the whole system once we verify the protocols on TCP stream-level and the `tap` module's basic packet capturing functionality.

Reduces the memory usage.

Fixes possible key collisions that are introduced by https://github.com/up9inc/mizu/pull/778

~~It breaks HTTP unit tests because the expected files are changed. See [the diff](https://github.com/up9inc/mizu/compare/develop...fix/matcher-map-tmp?expand=1). So we either need to update the files on Google Cloud bucket just before merging it, or someone with write access merges this PR by force and we update the files later on. (the purpose in here is not breaking the tests on other branches)~~


gadotroee:
I will be able to review it soon, but about the fact that it broke the tests, maybe we can upload the new expected to another folder and change the path that those tests are pull and then check it is working before merge it?
then after merging and when we sure it is not use the old files (maybe only after doing the release (in case we need a hotfix from main that still uses the old files) we will delete the un necessary files.

@mertyildiran  wdyt?

tap/extensions/kafka/matcher.go
```diff
@@ -5,20 +5,19 @@ import (
...
-var reqResMatcher = CreateResponseRequestMatcher() // global
 const maxTry int = 3000
 
 type RequestResponsePair struct {
 	Request  Request
 	Response Response
 }
 
-// Key is {client_addr}:{client_port}->{dest_addr}:{dest_port}::{correlation_id}
+// Key is {client_addr}:{client_port}_{dest_addr}:{dest_port}_{correlation_id}
```
gadotroee, line 15:
```suggestion
// Key is {client_addr_{client_port}_{dest_addr}_{dest_port}_{correlation_id}
```

tap/extensions/http/main_test.go
```diff
@@ -155,14 +154,6 @@ func TestDissect(t *testing.T) {
 
 		stop <- true
 
-		sort.Slice(items, func(i, j int) bool {
```
gadotroee, line 158: where is this sort happening? or maybe it is not needed from some reason?

tap/cleaner.go
```diff
@@ -72,25 +66,3 @@ func (cl *Cleaner) dumpStats() CleanerStats {
 
 	return stats
 }
-
-func deleteOlderThan(matcherMap *sync.Map, t time.Time) int {
```
gadotroee, line 76: how can we remove this logic of deletion? something else is handling the cleaning?

tap/extensions/kafka/matcher.go
gadotroee: I think there is another place (not changed in this PR so I couldn't suggest an edit) that has comment with the old format key "with ":" and not "_". not sure it is super important bit if we already have some comment is should reflect the right explanation IMO

mertyildiran: @gadotroee oh yeah I missed this comment. Thanks!

tap/extensions/http/main_test.go
mertyildiran: @gadotroee with this change order became consistent somehow. So that sorting became redundant.

tap/cleaner.go
mertyildiran: @gadotroee with this change the matcher maps created and destroyed for each TCP stream. So for short-living TCP connections the matcher maps are short-living. But for long-living TCP connections, if they are very long, there is no drop mechanism for unmatched requests or responses implemented in this PR. I'm not sure if there is a need for that.

With this PR there are 3 matcher maps for each TCP stream. So the number of iteration scales with the number of TCP streams. But the `clean` function is triggered every 10 seconds. Therefore I don't know.

mertyildiran:
> I will be able to review it soon, but about the fact that it broke the tests, maybe we can upload the new expected to another folder and change the path that those tests are pull and then check it is working before merge it? then after merging and when we sure it is not use the old files (maybe only after doing the release (in case we need a hotfix from main that still uses the old files) we will delete the un necessary files.
> 
> @mertyildiran wdyt?

Good idea :+1: Let me upload them to a new folder.

tap/cleaner.go
mertyildiran: @gadotroee with https://github.com/up9inc/mizu/commit/6363d6392a5f369f384238cadf57332023ebf2a4 I brought back the  `deleteOlderThan` method.

gadotroee (approved):

PR #819, opened 2022-02-16
TRA-4331 Implement full data streaming over WebSocket

mertyildiran:
Example usage:

```python
import websocket
import _thread
import time
import json

message_counter = 0

def on_message(ws, message):
    global message_counter
    d = json.loads(message)
    # print(message)
    if d["messageType"] == "fullEntry":
        print(message)
        message_counter += 1
        print(message_counter)

def on_error(ws, error):
    print(error)

def on_close(ws, close_status_code, close_msg):
    print("### closed ###")

def on_open(ws):
    def run(*args):
        ws.send(
            'timestamp >= datetime("02/16/2022, 10:38:32.352 AM") and timestamp <= datetime("02/16/2022, 10:38:34.229 AM")'
        ) # query
        ws.send('1') # enable full entries (empty: disable, non-empty: enable)
        time.sleep(10000)
        ws.close()
        print("thread terminating...")
    _thread.start_new_thread(run, ())

if __name__ == "__main__":
    # websocket.enableTrace(True)
    ws = websocket.WebSocketApp("ws://localhost:8899/ws",
        on_open=on_open,
        on_message=on_message,
        on_error=on_error,
        on_close=on_close
    )

    ws.run_forever()
```


ui/src/components/Pages/TrafficPage/TrafficPage.tsx
```diff
@@ -122,6 +122,7 @@ export const TrafficPage: React.FC<TrafficPageProps> = ({setAnalyzeStatus}) => {
         ws.current.onopen = () => {
             setWsConnection(WsConnectionStatus.Connected);
             ws.current.send(query);
+            ws.current.send("0");
```
gadotroee, line 125: can we change it in a way that it will don't need to send the "0" here?
I understand it is disabling the full data streaming but I think we can assume it is disabled until someone enabled it..

(maybe I misunderstand and it is not what the "0" here for

agent/pkg/api/socket_routes.go
```diff
@@ -111,19 +111,35 @@ func websocketHandler(w http.ResponseWriter, r *http.Request, eventHandlers Even
...
+				break
 			}
+		}
 
+		if breakWholeLoop {
 			break
 		}
 
+		enableFullEntries := false
+		if len(params[1]) > 0 && params[1][0] != 48 {
```
gadotroee, line 137: 48 is key code for "0"?
maybe need to explain this consts somehow?

ui/src/components/Pages/TrafficPage/TrafficPage.tsx
mertyildiran: With https://github.com/up9inc/mizu/pull/819/commits/df4405eedd399abc7ac50b72ebff5bcfce610f12 I have made it empty string. We need to send message to the server regardless of the value.

agent/pkg/api/socket_routes.go
mertyildiran: With https://github.com/up9inc/mizu/pull/819/commits/df4405eedd399abc7ac50b72ebff5bcfce610f12 I have changed the criteria to being empty or non-empty.

ui/src/components/Pages/TrafficPage/TrafficPage.tsx
gadotroee: @mertyildiran 
why do we need to send this message?
can't we initilalize the value of  enableFullEntries to false (which is the default of boolena but maybe better to be specifically written) and then enabled by user action (and if needed it will send message of disable) but I don't see reason to send this 0.

am I miss anything?

mertyildiran: @gadotroee it's because in the second iteration of this loop:

https://github.com/up9inc/mizu/blob/df4405eedd399abc7ac50b72ebff5bcfce610f12/agent/pkg/api/socket_routes.go#L118-L130

this line is blocking until seeing a message:

https://github.com/up9inc/mizu/blob/df4405eedd399abc7ac50b72ebff5bcfce610f12/agent/pkg/api/socket_routes.go#L119

So there should be always two messages from client to server.

agent/pkg/api/socket_routes.go
```diff
@@ -111,19 +111,35 @@ func websocketHandler(w http.ResponseWriter, r *http.Request, eventHandlers Even
...
+				} else {
+					logger.Log.Errorf("Error reading message, socket id: %d, error: %v", socketId, err)
+				}
+
+				breakWholeLoop = true
+				break
 			}
+		}
 
+		if breakWholeLoop {
```
undera, line 132: https://stackoverflow.com/questions/51996175/how-to-break-out-of-nested-loops-in-go

mertyildiran: @undera oh there are labels in Go right? Fixed it with https://github.com/up9inc/mizu/commit/7fadd84df85425f7e25e79b594142b92d8459dfc

gadotroee (approved):

PR #820, opened 2022-02-16
Take ownership over DaemonSet

nimrod-up9:
Set mizu-api-server as the owner of mizu-tapper-daemonset.
This will cause k8s to delete mizu-tapper-daemonset if mizu-api-server is deleted.

The problem this PR solves:
In the case where Mizu is installed via Helm and not with the `mizu install` command, then there are left overs after `helm uninstall`. The reason for the leftovers is that Helm creates the api server but doesn't create the tappers. It doesn't know it needs to delete the tapper daemonset.
If mizu-api-server is marked as the owner of mizu-tapper-daemonst, then k8s will take care of deletion on its own.

https://github.com/up9inc/mizu/commit/5484b7c491993cd166fdeb33b5f9c742e21216b2 can be reverted with this PR.

Reading material:
- https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

### DIFF
--- a/cli/resources/createResources.go
+++ b/cli/resources/createResources.go
@@ -80,7 +80,7 @@ func CreateInstallMizuResources(ctx context.Context, kubernetesProvider *kuberne
 	}
 	logger.Log.Infof("configmap/%v created", kubernetes.ConfigMapName)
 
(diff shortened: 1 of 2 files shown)


nimrod-up9:
We used another solution: Helm creates a DaemonSet with 0 pods when it starts. api-server overwrites it, but Helm is still the owner.
Although it is uglier, we didn't have any issues with it for the past few months, so I'm scrapping this PR.

PR #821, opened 2022-02-16
Cypress fix: resizing mizu when needed

AdamKol-up9:


acceptanceTests/cypress/integration/testHelpers/TrafficHelper.js
```diff
@@ -82,27 +84,38 @@ export function checkFilterByMethod(funcDict) {
...
-                            if (summaryDict)
-                                leftTextCheck(entryNum, summaryDict.pathLeft, summaryDict.expectedText);
-
-                            if (!doneCheckOnFirst) {
-                                deepCheck(funcDict, protocolDict, methodDict, entry);
-                                doneCheckOnFirst = true;
+                    cy.get('#entries-length').invoke('text').then(len => {
+                        resizeIfNeeded(len);
+                        listElmWithIdAttr.forEach(entry => {
+                            if (entry?.id && entry.id.match(RegExp(/entry-(\d{2}|\d{1})$/gm))) {
```
gadotroee, line 90: if you know it is going to be number - there is no need in regex,
you can probably convert it to int and do some operation like > < and =

acceptanceTests/cypress/integration/tests/Rabbit.js
```diff
@@ -1,8 +1,8 @@
 import {checkFilterByMethod, valueTabs,} from "../testHelpers/TrafficHelper";
 
-
 it('opening mizu', function () {
     cy.visit(Cypress.env('testUrl'));
+    cy.get('#total-entries').invoke('text').should('match', /^[4-7][0-9]$/m)
```
gadotroee, line 5: same here, convert to number and do >= 40 and < 80 or something like this

gadotroee (changes requested):
added two improvement suggestions

acceptanceTests/cypress/integration/testHelpers/TrafficHelper.js
AdamKol-up9: yes i thought its prettier with regex, but I'll change it

AdamKol-up9: the regex is for checking  that the id is -> 'entry-X', its not only a number

acceptanceTests/cypress/integration/tests/Rabbit.js
AdamKol-up9: I need to use here the should(), If I convert the entries number to int I will need to use expect() instead, which is not good for this case.

gadotroee (approved):

PR #836, opened 2022-02-21
Update relase template for Mac M1 executable

IgorGov:
This should fix this issue:
image (5)

On Mac M1 if you override the executable the OS refuse to execute and will kill it because of caching mechanism


gadotroee (approved):

gadotroee (changes requested):
Changes needed

gadotroee (approved):

PR #843, opened 2022-02-23
Add the ability to set the insertion filter into CLI

mertyildiran:


agent/pkg/app/main.go
```diff
@@ -86,6 +86,11 @@ func ConfigureBasenineServer(host string, port string, dbSize int64, logLevel lo
 			}
 		}
 	}
+
+	// Set the insertion filter that comes from the config
+	if err := basenine.InsertionFilter(host, port, insertionFilter); err != nil {
```
gadotroee, line 91: I tried to run and it is working very well.
just didn't saw a case that this log is happening.. 
even when I use invalid filter it seems like it is not crashing (but it is not showing any information so it is filtered)
am I missing anything here?

gadotroee (commented):
all code looks ok, one comment that I have it about the "panic" in api server.
when the filter is not valid (i used for example "http1" the behavior is pretty weird, I just not get a working mizu and there is no log that helps me understand it.

I'm thinking about one of two options which not seems the best but maybe you will have other options:
1. check the syntax before starting the agent.
2. do not panic and ignore the filter if it is not valid 
each one of them cause some issue: 
first meaning that basenine query validation should happen when you execute the tap (in your laptop and there is no basenine server running - maybe it is possible easily but I'm not sure) 
the second will cause user confusion because he wrote filter and it is ignored 

@mertyildiran  wdyt?

mertyildiran:
@gadotroee for 1 the CLI should programmatically depend on Basenine means that this PR is required https://github.com/up9inc/basenine/pull/11 I'm about to merge that actually so I agree that we need to check against syntax errors on CLI level.

`http1` is not a syntax error without macros (with macros expanded it depends) but these are for example syntax errors in any case;

```
> request.path.= "something"
1:14: unexpected token "=" (expected Expression)
> request.path == "something
1:27: literal not terminated
```

For 2 I understand we shouldn't panic. So, I'm going to make it log an error instead.

gadotroee:
> @gadotroee for 1 the CLI should programmatically depend on Basenine means that this PR is required [up9inc/basenine#11](https://github.com/up9inc/basenine/pull/11) I'm about to merge that actually so I agree that we need to check against syntax errors on CLI level.
> 
> `http1` is not a syntax error without macros (with macros expanded it depends) but these are for example syntax errors in any case;
> 
> ```
> > request.path.= "something"
> 1:14: unexpected token "=" (expected Expression)
> > request.path == "something
> 1:27: literal not terminated
> ```
> 
> For 2 I understand we shouldn't panic. So, I'm going to make it log an error instead.

sounds good.
I will review once those changes will be added.

gadotroee (approved):

gadotroee:
from this PR the DB files will be called *.db and not *.bin. 
we need to be aware of this in cases of upgrade mizu instance because it will not be aware of the files it is looking. (If I understand correctly)

gadotroee (changes requested):
waiting for fixes in the change from *.bin to *.db

gadotroee:
> @gadotroee with [5003be4](https://github.com/up9inc/mizu/commit/5003be420716a3766dd711e98a29861d215eb79f) it will change legacy `.bin` files to `.db` files.

Ok sounds good so this will do the "migration" in case you have old data, right?

mertyildiran:
> > @gadotroee with [5003be4](https://github.com/up9inc/mizu/commit/5003be420716a3766dd711e98a29861d215eb79f) it will change legacy `.bin` files to `.db` files.
> 
> Ok sounds good - so this will do the "migration" in case you have old data, right?

Yeah exactly, it will **not** destroy the old data.

gadotroee (approved):

PR #852, opened 2022-02-24
Fixed location of pre tap permission files

RoyUP9:


cli/cmd/checkRunner.go
```diff
@@ -17,6 +17,11 @@ import (
 	"github.com/up9inc/mizu/shared/semver"
 )
 
+var (
+	//go:embed permissionFiles
+	f embed.FS
```
nimrod-up9, line 22: Does it have to be `f`? Can it be a meaningful name?

nimrod-up9 (approved):

PR #856, opened 2022-02-24
OAS: Endpoint reclustering

undera:
If we learned over time sufficient information to group endpoints together - do it.
Remember historical OperationIDs into custom property.


mertyildiran:
Closing since it's outdated.

PR #860, opened 2022-02-27
fixed install printout with updated helm chart name

haiut:


haiut:
duplicate .. closing without merge.

PR #865, opened 2022-02-28
set bpf filter for pcap

dvdlevanon:


tap/source/packet_source_manager.go
```diff
@@ -2,200 +2,135 @@ package source
...
-		result = append(result, src)
 	}
-
-	return result
 }
 
-func newNetnsPacketSource(pid int, nsh netns.NsHandle, interfaceName string,
-	behaviour TcpPacketSourceBehaviour) (*tcpPacketSource, error) {
+func (m *PacketSourceManager) getRelevantPids(procfs string, pods []v1.Pod) map[string]bool {
+	relevantPids := make(map[string]bool)
```
nimrod-up9, line 85: Why `make(map[string]bool)` and not `[]string`?

tap/source/packet_source_manager.go
```diff
@@ -2,200 +2,135 @@ package source
...
 
-	go func(done chan<- *tcpPacketSource) {
-		// Setting a netns should be done from a dedicated OS thread.
-		//
-		// goroutines are not really OS threads, we try to mimic the issue by
-		//	locking the OS thread to this goroutine
-		//
-		runtime.LockOSThread()
-		defer runtime.UnlockOSThread()
+	if err != nil {
```
nimrod-up9, line 90: Not critical, but here I would do:
```go
if envoyPids, err := discoverRelevantEnvoyPids(procfs, pods); err != nil {
```
because `envoyPids` and `err` are only used in this `if`.

tap/source/packet_source_manager.go
```diff
@@ -2,200 +2,135 @@ package source
...
+	}
 
-		oldnetns, err := netns.Get()
+	linkerdPids, err := discoverRelevantLinkerdPids(procfs, pods)
 
-		if err != nil {
-			logger.Log.Errorf("Unable to get netns of current thread %v", err)
-			errors <- err
-			return
+	if err != nil {
```
nimrod-up9, line 100: Same.
```go
if linkerdPids, err := discoverRelevantLinkerdPids(procfs, pods); err != nil
```

tap/source/packet_source_manager.go
```diff
@@ -2,200 +2,135 @@ package source
...
+func NewPacketSourceManager(procfs string, filename string, interfaceName string,
 	mtls bool, pods []v1.Pod, behaviour TcpPacketSourceBehaviour) (*PacketSourceManager, error) {
-	sources := make([]*tcpPacketSource, 0)
-	sources, err := createHostSource(sources, filename, interfaceName, behaviour)
-
-	if err != nil {
-		return nil, err
-	}
-
-	sources = createSourcesFromPids(sources, procfs, pids, interfaceName, behaviour)
```
nimrod-up9, line 27: Where did this functionality go?
I see that we search for envoy PIDs and Linkerd PIDs, but there is no equivalent for creating sources from some PID (only from host "0" PID).

tap/source/netns_packet_source.go
```diff
@@ -0,0 +1,83 @@
...
+	return src, nil
+}
+
+func newPacketSourceFromNetnsHandle(pid string, nsh netns.NsHandle, interfaceName string,
+	behaviour TcpPacketSourceBehaviour) (*tcpPacketSource, error) {
+
+	done := make(chan *tcpPacketSource)
+	errors := make(chan error)
+
+	go func(done chan<- *tcpPacketSource) {
```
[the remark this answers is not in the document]
nimrod-up9: `runtime.LockOSThread()` locks the routine. We don't want to lock the routine in which `newPacketSourceFromNetnsHandle` runs (which is the main routine).

tap/source/netns_packet_source.go
```diff
@@ -0,0 +1,83 @@
...
+
+		oldnetns, err := netns.Get()
+
+		if err != nil {
+			logger.Log.Errorf("Unable to get netns of current thread %w", err)
+			errors <- err
+			return
+		}
+
+		if err := netns.Set(nsh); err != nil {
```
nimrod-up9, line 53: What happens to the communication channel (ws) with the server?

tap/source/netns_packet_source.go
```diff
@@ -0,0 +1,83 @@
...
+
+		done <- src
+	}(done)
+
+	select {
+	case err := <-errors:
+		return nil, err
+	case source := <-done:
+		return source, nil
+	}
```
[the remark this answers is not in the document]
nimrod-up9: Sorry, not possible.
The code above will just skip waiting.
The original code is an idiom in go.

tap/source/packet_source_manager.go
```diff
@@ -2,200 +2,135 @@ package source
...
-	result := make([]*tcpPacketSource, 0)
+func (m *PacketSourceManager) UpdatePods(mtls bool, procfs string, pods []v1.Pod,
+	interfaceName string, behaviour TcpPacketSourceBehaviour) {
+	if mtls {
+		m.updateMtlsPods(procfs, pods, interfaceName, behaviour)
+	}
 
-	for _, pidstr := range pids {
-		pid, err := strconv.Atoi(pidstr)
+	m.setBPFFilter(pods)
```
nimrod-up9, line 57: - Check very long filter.
- Check if `pods` is only on this node or cluster wide.

tap/source/packet_source_manager.go
nimrod-up9: Because lookup is `O(1)`.
Also because we need a set (only one copy of each PID).

nimrod-up9 (commented):
.

tap/source/netns_packet_source.go
dvdlevanon: from my test it continue working

tap/source/packet_source_manager.go
dvdlevanon: 1. checked and added a limit to max 150 pods - my testings show 200 is also fine but to be on the safe side. I assume it depend on the system configuration somehow, but i'm not sure how
2. its pods per node, so 150 should be fine

nimrod-up9 (approved):

PR #876, opened 2022-03-03
Run PR validation check only when needed & use docker cache during build

IgorGov:


.github/workflows/test.yml
```diff
@@ -19,14 +19,16 @@ jobs:
     name: Unit Tests
     runs-on: ubuntu-latest
     steps:
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
```
gadotroee, line 25: why do we need to fetch-depth 2?

gadotroee (commented):
one small question that I want to understand but it looks good

gadotroee:
also, I see it caused the codecov to fail and I guess it is because we are not uploading the full report (for example all exceptions tests didn't run here).. 
do we have any solution for it?

.github/workflows/test.yml
IgorGov: for git diff to work

IgorGov:
> also, I see it caused the codecov to fail and I guess it is because we are not uploading the full report (for example all exceptions tests didn't run here)..
> do we have any solution for it?
no.

.github/workflows/test.yml
gadotroee: 👍

RamiBerm (approved):

gadotroee (approved):

PR #887, opened 2022-03-12
Improve acceptance tests setup script

gadotroee:


IgorGov (dismissed):

acceptanceTests/setup.sh
```diff
@@ -27,44 +30,52 @@ echo "Starting minikube..."
...
-eval $(minikube docker-env)
+eval "$(minikube docker-env)"
 
 echo "Build agent image"
 docker build -t mizu/ci:0.0 .
 
 echo "Build cli"
 cd cli && make build GIT_BRANCH=ci SUFFIX=ci
 
+# TODO: need to understand how to fail is password is asked (sudo)
```
RoyUP9, line 76:
```suggestion
# TODO: need to understand how to fail is password is asked (sudo)
```
```suggestion
# TODO: need to understand how to fail if password is asked (sudo)
```

RoyUP9, line 76:
```suggestion
# TODO: need to understand how to fail is password is asked (sudo)
```
```suggestion
# TODO: need to understand how to fail if password is asked (sudo)
```

acceptanceTests/setup.sh
```diff
@@ -27,44 +30,52 @@ echo "Starting minikube..."
...
 
 echo "Creating redis service"
-kubectl expose deployment redis --type=LoadBalancer --port=6379 -n mizu-tests
+kubectl expose deployment redis --type=LoadBalancer --port=6379 -n mizu-tests --dry-run=client -o yaml | kubectl apply -f -
 
 echo "Creating rabbitmq service"
-kubectl expose deployment rabbitmq --type=LoadBalancer --port=5672 -n mizu-tests
+kubectl expose deployment rabbitmq --type=LoadBalancer --port=5672 -n mizu-tests --dry-run=client -o yaml | kubectl apply -f -
 
+# TODO: need to understand how to fail if address already in use
```
nimrod-up9, line 60: Maybe check ip tables?
But this is just one possible issue.
We need to make sure that `kubectl proxy` quits on every error.
Maybe check job / process status after running `kubectl proxy`?
Or read from `PROXY_LOG` and search some string that indicates that it started correctly.
Can also be applied to `minikube tunnel` below.

acceptanceTests/setup.sh
```diff
@@ -27,44 +30,52 @@ echo "Starting minikube..."
...
-eval $(minikube docker-env)
+eval "$(minikube docker-env)"
 
 echo "Build agent image"
 docker build -t mizu/ci:0.0 .
 
 echo "Build cli"
 cd cli && make build GIT_BRANCH=ci SUFFIX=ci
 
+# TODO: need to understand how to fail if password is asked (sudo)
```
nimrod-up9, line 76: Why? What's the use case?
Either this is run manually on a local machine, in which case it is ok to prompt for password, or it is run in CI, in which case we won't get asked.

Or is the issue that the password is asked in a detached process, and `minikube tunnel` waits for it without notifying the user?

acceptanceTests/setup.sh
```diff
@@ -27,44 +30,52 @@ echo "Starting minikube..."
 minikube start
 
 echo "Creating mizu tests namespaces"
-kubectl create namespace mizu-tests
-kubectl create namespace mizu-tests2
+kubectl create namespace mizu-tests --dry-run=client -o yaml | kubectl apply -f -
```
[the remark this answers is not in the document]
gadotroee: the idea is to not fail when the resource exists and if you do "kubectl create" it will write error. (not really fails)
but when you do dry run it returns the yaml and then you are running kubectl apply that knows to handle it and will "configure" the resource and as we already know the kubectl apply can handle upsert so it will create if not exists and update if exists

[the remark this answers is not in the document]
gadotroee: if I'm not writing dry-run, it will create the namespace when I'm trying to get the yaml only without creation

acceptanceTests/setup.sh
gadotroee: yes I just don't want to make it complex but I agree that we need to fail on every error

RoyUP9 (approved):

PR #896, opened 2022-03-17
Add `/db/flush` and `/db/reset` API endpoints

mertyildiran:


agent/pkg/utils/utils.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/up9inc/mizu/shared/logger"
 )
 
+var (
+	StartTime int64 // global
```
gadotroee, line 21: why do we have global StartTime here and in https://github.com/up9inc/mizu/pull/896/files#diff-5fc5ab56849ddf72f277967087b00e1d0da19aec8d6116d4d9ea646919627962R23, maybe they are too different variables but it confuses me when i tried to understand

gadotroee (changes requested):
one small confusing line, other code make sense (as long as those endpoints should be internal and used only when we really need to "reset" the end like we had with Alex last week)

agent/pkg/utils/utils.go
mertyildiran: @gadotroee fixed with https://github.com/up9inc/mizu/commit/4ac4b3314e1f23edb16ea50142ec645d65c5a28c

gadotroee (commented):
This change is ok (mostly api server) and just waiting for the acceptance tests fix to be merged

gadotroee (approved):
looking good

PR #897, opened 2022-03-18
TRA-4383 Calculate request and response sizes and display them instead of `BodySize` field

mertyildiran:
It will as far as I understand, require publishing a new version to `@up9/mizu-common` NPM package.

Screenshot from 2022-03-19 12-04-48

Screenshot from 2022-03-19 12-04-52


gadotroee:
You are probably correct about publishing new version of the common library. I will try to talk with @leon-up9 / @lirazyehezkel  and see how we it is done. Also there will be more changes before to fix the broken tests (Leon is working on it and I hope it will be finished tomorrow)

mertyildiran:
> You are probably correct about publishing new version of the common library. I will try to talk with @leon-up9 / @lirazyehezkel and see how we it is done. Also there will be more changes before to fix the broken tests (Leon is working on it and I hope it will be finished tomorrow)

@gadotroee OK :+1:

gadotroee (commented):
all the changes here looking good.
I hope we will finish work on the ui-common fix and then we will be able to publish new version with  this changes and push another commit here to use this new version.

gadotroee (approved):

PR #901, opened 2022-03-20
Update tap targets over ws

nimrod-up9:
Update tappers via websocket instead of by env var. This way the DaemonSet doesn't have to be applied just to notify the tappers that the tap targets changed. The number of tapper restarts is reduced. The DaemonSet still gets applied when there is a need to add/remove a tapper.


agent/pkg/providers/tappers/tappers_provider.go
```diff
@@ -81,3 +83,12 @@ func saveStatus() {
 		logger.Log.Errorf("Error saving tappers status, err: %v", err)
 	}
 }
+
+func SetNodeToTappedPodMap(nodeToTappedPodsMap shared.NodeToPodsMap) {
+	logger.Log.Infof("Setting node to tapped pods map to %v", nodeToTappedPodsMap)
```
nimrod-up9, line 88: Not sure about this. Message could be quite long. This is with two nodes, two pods:
```
[2022-03-21T04:04:48.803+0000] INFO  ▶ Setting node to tapped pods map to map[gke-nimrod-feb-24-1-default-pool-5b71ec16-1df3:[{{ } {httpbin-5d4c65bb96-rpstq      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] [] [] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] [] <nil> nil} { []     10.40.0.3 [] <nil> [] [{ {nil nil nil} {nil nil nil} false 0   containerd://31f50019f46d2a135f99332427ae58a1a31850db02580607fc73b347485b438a <nil>}]  []}}] gke-nimrod-feb-24-1-default-pool-5b71ec16-s65f:[{{ } {load-6f5d6db7b7-9flnr      0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []} {[] [] [] []  <nil> <nil>  map[]   <nil>  false false false <nil> nil []   nil  [] []  <nil> nil [] <nil> <nil> <nil> map[] [] <nil> nil} { []     10.40.1.9 [] <nil> [] [{ {nil nil nil} {nil nil nil} false 0   containerd://3869c1dfc7db4500de73b703bbf169f7be7ce870ce8be8d0039f2dc98bf4c312 <nil>}]  []}}]] ▶ [1 tappers_provider.go:88 SetNodeToTappedPodMap]
```

nimrod-up9: Changed to something less verbose. Maybe instead we should keep all the info but change to `Debug`?

agent/pkg/api/socket_server_handlers.go
```diff
@@ -65,6 +77,16 @@ func BroadcastToBrowserClients(message []byte) {
...
+func BroadcastToTapperClients(message []byte) {
+	for _, socketId := range tapperClientSocketUUIDs {
+		go func(socketId int) {
+			if err := SendToSocket(socketId, message); err != nil {
+				logger.Log.Error(err)
+			}
+		}(socketId)
+	}
+}
+
```
gadotroee, line 89:
```suggestion
func broadcastToClients(message []byte, clientIds []int) {
	for _, socketId := range clientIds {
		go func(socketId int) {
			if err := SendToSocket(socketId, message); err != nil {
				logger.Log.Error(err)
			}
		}(socketId)
	}
}

func BroadcastToTapperClients(message []byte) {
               broadcastToClients(message, tapperClientSocketUUIDs)
}

func BroadcastToBrowserClients(message []byte) {
               broadcastToClients(message, browserClientSocketUUIDs)
}
```

gadotroee: and remove the function before of BroadcastToBrowserClients

shared/models.go
```diff
@@ -75,11 +76,29 @@ type WebSocketStatusMessage struct {
...
 	TapTargets []v1.Pod `json:"pods"`
 }
 
+type NodeToPodsMap map[string][]v1.Pod
+
+func (np NodeToPodsMap) Summary() map[string][]string {
+	summary := make(map[string][]string)
+	for node, pods := range np {
+		for _, pod := range pods {
+			summary[node] = append(summary[node], pod.Name + "." + pod.Namespace)
```
gadotroee, line 95:
```suggestion
			summary[node] = append(summary[node], pod.Namespace + "/" +  pod.Name)
```

gadotroee (commented):
added my thoughts until this point

gadotroee (approved):

gadotroee (approved):

PR #902, opened 2022-03-20
Refactored Dockerfile

adriang-90:
Here I wanted to improve **Dockerfile** to follow best practices, such as:
* pinned version instead of using latest which is prone to errors if the image will ever be updated
* consolidated **RUN** commands so it decrease steps in build phase to 6 from 14
* used **WORKDIR** to switch a directory


haiut:
@adriang-90 hey, thanks for your fix.. we're happy to adopt it, except for the first line's change -- we prefer to keep it as `latest` .. this image is used as base-image for other images we build, and we re-build it periodically after careful review, so there's no real danger in using `latest` tag in our case.. would you please update the PR and I'll approve it ?

adriang-90:
I have changed the tag to *latest* as you wanted. May I also take care of Dockerfile in root directory? I have ran **hadolint** on it and it showed things such as:
* Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages
* Double quote to prevent globbing and word splitting.

IgorGov (approved):

IgorGov:
@adriang-90 Thanks for your contribution!

PR #910, opened 2022-03-22
Add from ARM64 to AMD64 cross-compilation support to the agent image

mertyildiran:
Follow up of https://github.com/up9inc/mizu/pull/659 Adds the build strategy below:

### From ARM64 (AArch64) to AMD64 (x86-64)

```shell
$ docker build . -t up9inc/mizu:develop-amd64 --build-arg BUILDARCH=arm64v8 --build-arg TARGETARCH=amd64
$ docker push up9inc/mizu:develop-amd64
```


gadotroee:
I think you should consider need to improve the Dockerfile based on the changes came from  https://github.com/up9inc/mizu/pull/902

mertyildiran:
@gadotroee the `Dockerfile` is an intermediate builder image. It's not a production image.

gadotroee (approved):

PR #914, opened 2022-03-23
Build ui-common part of CI

IgorGov:


Dockerfile
```diff
@@ -1,12 +1,24 @@
...
+RUN npm run build
+RUN npm pack
+
 ### Front-end
 FROM node:16 AS front-end
 
 WORKDIR /app/ui-build
 
 COPY ui/package.json ui/package-lock.json ./
+COPY --from=front-end-common ["/app/ui-build/up9-mizu-common-1.0.138.tgz", "."]
```
leon-up9, line 21: just to avoid confusion maybe version should be " 0.0.0" and not "1.0.138"

leon-up9: and change the version in ui-common/packge.json ,too.
this file name wont be created ever again

lirazyehezkel (approved):

PR #917, opened 2022-03-23
OAS: rework data feeding + sampleIDs

undera:
1. Change it into reading directly from DB, since we need entryIDs
2. Add sample IDs on different levels


.github/workflows/test.yml
```diff
@@ -18,6 +18,7 @@ jobs:
   run-unit-tests:
     name: Unit Tests
     runs-on: ubuntu-latest
+    timeout-minutes: 30 # TODO: is it low enough?
```
IgorGov, line 21: did you reached timeout with this workflow?

IgorGov, line 21: did you reached timeout with this workflow?

IgorGov: i don't think it should be more than 10 minutes.
if it more then that we need to rethink how we running the tests

IgorGov: i don't think it should be more than 10 minutes.
if it more then that we need to rethink how we running the tests

IgorGov (approved):

.github/workflows/test.yml
undera: I have observed once that the job were taking very long time. I thought that something in unit tests made them hang indefinitely. IMO having safety valve timeout is a good idea. I will change it into 10m

IgorGov (approved):

PR #918, opened 2022-03-23
quick tls update pods solution

dvdlevanon:


nimrod-up9 (approved):

nimrod-up9 (changes requested):
Update tls tapper from https://github.com/up9inc/mizu/blob/2eb5b2308751f471b2d10b2d3105fc000f345b12/tap/passive_tapper.go#L108

nimrod-up9 (approved):

PR #929, opened 2022-03-24
Updated Dockerfile and fixed typo in Makefile

adriang-90:
Small update on **Dockerfile** with a help from hadolint:
* added --no-cache
* consolidated multiple RUN commands
* deleted one **WORKDIR**, was both necessary?


gadotroee:
Hi @adriang-90,

Thanks for opening this PR.

The changes are good and improve our Dockerfile.
About the question of the extra WORKDIR, it will be no necessary in the near future - but currently there are files that this container will be write to /app/data and this writing assumes the folder exists so this step of running the WORKDIR ensures it will happen. I understand it is not best practice and as I said we will remove this limitation when we are doing some other change but if you can "undo" this change I will approve and we will merge this PR (we still need the WORKDIR to be /app in the end so this line of /app/data should be before).

gadotroee:
Hi @adriang-90  I saw your change and planned to approve it but meanwhile there was hotfix that was created and caused one small conflict. I don't think I can do the resolving here because the branch is in a fork in your account but after you solve it (pretty small conflict in the version we are installing of the "basenine" binary) I will approve it and we will insert your change

sorry about the small delay (I see you solved it pretty fast but I didn't approve it, and anyway thanks for you contributions - it is much appreciated.

adriang-90:
I have added the small hotfix to this PR. Can you check if I didnt mess up anything?

gadotroee:
> I have added the small hotfix to this PR. Can you check if I didnt mess up anything?

It seems like this is the correct fix.
The issue is that I still see conflict in the PR so I can't really merge this as it is now.
I think you need to merge / rebase the changes from develop so git will be able with the changes and will not mark conflicts.

adriang-90:
It should be ok now. It was my mistake I didnt resolve conflict, just did new commit

gadotroee (approved):
No problem.
As I said before - Many thanks for the contribution

PR #958, opened 2022-03-31
Remove gRPC related modifications

mertyildiran:


gadotroee (commented):
@mertyildiran 

It looks ok overall but I'm probably missing the problem, what issue caused from this modification that we did?

ui-common/src/components/UI/StatusCode.tsx
```diff
@@ -37,9 +37,9 @@ export function getClassification(statusCode: number): string {
 
     // 1 - 16 HTTP/2 (gRPC) status codes
```
gadotroee, line 38: this comment is probably not relevant is we are not changing the status to be 1-16 (from `Grpc-Status` header value)

nimrod-up9:
This PR affects UI and filters.

Before:
image

After:
image

Status and method changed to the gRPC ones to HTTP/2.
The filters that use status and method also changed.

Maybe it is worth adding macros `grpc.status`, `grpc.method`?

nimrod-up9:
Got `panic` here: https://github.com/up9inc/mizu/blob/60232f1c0af4df9a2d394098e8230956e8ea638d/tap/extensions/http/main.go#L253

```
[GIN] 2022/04/03 - 10:59:28 | 200 |   42.168914ms |       127.0.0.1 | POST     "/status/tappedPods"
panic: runtime error: index out of range [200] with length 17

goroutine 31 [running]:
github.com/up9inc/mizu/tap/extensions/http.dissecting.Analyze({0xc00096deb7, 0xc000335320}, 0xc0003c8380, {0xc000335320, 0x11}, {0xc000c24048, 0x13}, {0xc000bcb140, 0x9})
/app/tap/extensions/http/main.go:253 +0x1190
github.com/up9inc/mizu/agent/pkg/api.startReadingChannel(0xc0001604e0, 0xc0005b4c90)
/app/agent-build/pkg/api/main.go:122 +0x3ae
github.com/up9inc/mizu/agent/pkg/api.StartReadingEntries(0x909090a7b203a22, 0x6665722422090909, 0x64242f2322203a22)
/app/agent-build/pkg/api/main.go:59 +0x3a
created by github.com/up9inc/mizu/agent/pkg/app.GetEntryInputChannel
/app/agent-build/pkg/app/main.go:102 +0x10f
```

Because `statusCode` is now HTTP status and not gRPC status, it is in the >200. The array `grpcStatusCodes` only contains 17 values.

mertyildiran:
@nimrod-up9 I have removed the code that causes the panic.

### Macros

A macro like `grpc.status` expands to `response.headers["Grpc-Status"]` is OK.

For `grpc.method` macro we need to expand to `request.pathSegments[-1]` but we don't support negative indexes yet and I'm not sure whether `request.pathSegments[1]` is always the gRPC method or not. We can't define a regex based macro because macros don't support arguments right now.

In conclusion, the users should figure out how they are going to filter based on gRPC specific properties based on the response header or request path, for a while at least.

Alternatively, we don't merge this PR.

mertyildiran:
@nimrod-up9 also the vertical `gRPC` label shouldn't change to `HTTP/2` with the changes in this PR. At least, this is what I expect. If it's changing then something is wrong.

gadotroee:
> 



> @nimrod-up9 also looking at the screenshots that you have shared; the vertical `gRPC` label shouldn't change to `HTTP/2` with the changes in this PR. At least, this is what I expect. If it's changing then something is wrong.


I'm almost sure it is caused by the check that was removed for content-type (I have added it back here https://github.com/up9inc/mizu/pull/958/commits/690401bdad9a6f4a21a9ae160359941b77d8a2d1)

nimrod-up9:
Roee's fix stets the tag to gRPC.
This is the view now (status and method are HTTP not gRPC).
image

mertyildiran:
@nimrod-up9 @gadotroee If it's OK for everybody, we can merge it?

gadotroee:
> @nimrod-up9 @gadotroee If it's OK for everybody, we can merge it?

I'm planning to talk to Refael today and ask what he prefer (showing the status as "200" and "POST" or grpc status and method).

if he said he prefer gRPC information we might need to add more commits here (but this time change it when returning it from api server and not before inserting)

anyway I hope we will have answer before tomorrow morning if it is ok

mertyildiran:
@gadotroee OK.

gadotroee:
I forget to write here after I talked with refael we will merge this PR like this, still need to understand a little bit about gRPC if the status code (GRPC-status header value) and the method (last part of the "path") are interesting enough and if so doing it separately. 
The issue is that currently tests are broken (from what I know) and to reduce spam alerts I prefer to merge it when they are working so I will ensure we are merging this after tests are fixed.

gadotroee (approved):

PR #961, opened 2022-04-03
Refactor ws

RamiBerm:


gadotroee (commented):
I'm thinking about changing the tapper components to be like the the browser component (struct and no int array)

agent/pkg/api/socket_server_handlers.go
```diff
@@ -88,7 +96,33 @@ func BroadcastToTapperClients(message []byte) {
...
+		if browserClients[socketId] != nil && browserClients[socketId].dataStreamCancelFunc == nil {
+			var params WebSocketParams
+			if err := json.Unmarshal(message, &params); err != nil {
+				logger.Log.Errorf("Error: %v", socketId, err)
+				return
+			}
+
+			entriesStreamer := dependency.GetInstance(dependency.EntriesSocketStreamer).(EntryStreamer)
+			ctx, cancelFunc := context.WithCancel(context.Background())
+			err := entriesStreamer.Get(ctx, socketId, &params)
```
RoyUP9, line 113: check if we have a sync problem

agent/pkg/api/socket_data_streamer.go
```diff
@@ -0,0 +1,141 @@
...
+		logger.Log.Error(err)
+	}
+}
+
+func (e *DefaultEntryStreamerSocketConnector) CleanupSocket(socketId int) {
+	socketObj := connectedWebsockets[socketId]
+	socketCleanup(socketId, socketObj)
+}
+
+type BasenineEntryStreamer struct{}
```
RoyUP9, line 67: separate

RoyUP9 (approved):

PR #964, opened 2022-04-03
Add ability to set query in OAS Generator

nimrod-up9:


agent/pkg/oas/oas_generator.go
```diff
@@ -181,12 +176,25 @@ func (g *defaultOasGenerator) GetServiceSpecs() *sync.Map {
 	return g.serviceSpecs
 }
 
-func NewDefaultOasGenerator(c *basenine.Connection) *defaultOasGenerator {
+func (g *defaultOasGenerator) SetEntriesQuery(query string) {
```
nimrod-up9, line 179: `SetEntriesQuery` has to happen before `Start`. Setting the query after starting won't cause the generator to restart.
I don't like it.

Alternatives in increasing order of code changes:
1. Return an error if trying to set the query while the generator is running.
1. Set the entry on `Start` (pass the query as a parameter to `Start`)
1. Pass the query during creation (as a parameter to `NewDefaultOasGenerator`. Requires some refactoring.
1. Don't pass a `basenine.Connection` to `NewDefaultOasGenerator`. Instead pass an `entriesStreamer` interface (create it if it doesn't exist). Set the DB connection and the filters from outside of the generator class. Requires even more refactoring.

Because this PR blocks the release, I think we should merge as is and immediately begin working on a better solution.

undera: I think options 1 and 2 is a good idea to implement right away.

undera (changes requested):
Unit tests are broken. The code around OAS is supposed to not depend on read DB availability.

undera (approved):
Looks good now

PR #977, opened 2022-04-07
Make minor changes to OasGenerator

nimrod-up9:
* Added log message
* Remove Reset function from OasGenerator interface, use Stop+Start instead
* SetEntriesQuery returns a bool stating whether the query changed

### DIFF
--- a/agent/pkg/oas/oas_generator.go
+++ b/agent/pkg/oas/oas_generator.go
@@ -23,7 +23,7 @@ type OasGenerator interface {
 	Stop()
 	IsStarted() bool
 	GetServiceSpecs() *sync.Map
-	SetEntriesQuery(query string)
+	SetEntriesQuery(query string) bool
 }
 
 type defaultOasGenerator struct {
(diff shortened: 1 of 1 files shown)


agent/pkg/oas/oas_generator.go
```diff
@@ -160,24 +159,28 @@ func (g *defaultOasGenerator) getGen(dest string, urlStr string) *SpecGen {
 	val, found := g.serviceSpecs.Load(dest)
 	var gen *SpecGen
 	if !found {
-		gen = NewGen(u.Scheme + "://" + dest)
+		serviceAddress := u.Scheme + "://" + dest
+		logger.Log.Infof("Create SpecGen for service %s", serviceAddress)
```
RoyUP9, line 163: I don't think we need this log, it could be spamming

RoyUP9:
I don't see the change from reset to stop in the controller

agent/pkg/oas/oas_generator.go
nimrod-up9: https://github.com/up9inc/mizu-enterprise/pull/137/commits/29a819e512052c3d982da7fb82f5fdd5b48b51a3

nimrod-up9:
> I don't see the change from reset to stop in the controller

It's in Enterprise.
https://github.com/up9inc/mizu-enterprise/pull/137/commits/29a819e512052c3d982da7fb82f5fdd5b48b51a3

PR #979, opened 2022-04-07
Increase the OOM risk in exchange of less idle CPU usage

mertyildiran:
This PR will reduce the idle CPU usage. But it will **increase the out of memory (OOM) risk** under heavy traffic load.

60 seconds of CPU profiling:

### Before

Screenshot from 2022-04-07 17-50-24

### After

Screenshot from 2022-04-07 17-50-34

### DIFF
--- a/tap/tcp_streams_map.go
+++ b/tap/tcp_streams_map.go
@@ -62,6 +62,7 @@ func (streamMap *tcpStreamMap) getCloseTimedoutTcpChannelsInterval() time.Durati
 func (streamMap *tcpStreamMap) closeTimedoutTcpStreamChannels() {
 	tcpStreamChannelTimeout := GetTcpChannelTimeoutMs()
 	closeTimedoutTcpChannelsIntervalMs := streamMap.getCloseTimedoutTcpChannelsInterval()
+	logger.Log.Infof("Using %d ms as the close timedout TCP stream channels interval", closeTimedoutTcpChannelsIntervalMs/time.Millisecond)
 	for {
 		time.Sleep(closeTimedoutTcpChannelsIntervalMs)
 		_debug.FreeOSMemory()
@@ -73,7 +74,7 @@ func (streamMap *tcpStreamMap) closeTimedoutTcpStreamChannels() {
 					stream.Close()
 					diagnose.AppStats.IncDroppedTcpStreams()
 					logger.Log.Debugf("Dropped an unidentified TCP stream because of timeout. Total dropped: %d Total Goroutines: %d Timeout (ms): %d",
-						diagnose.AppStats.DroppedTcpStreams, runtime.NumGoroutine(), tcpStreamChannelTimeout/1000000)
+						diagnose.AppStats.DroppedTcpStreams, runtime.NumGoroutine(), tcpStreamChannelTimeout/time.Millisecond)
 				}
 			} else {
 				if !stream.superIdentifier.IsClosedOthers {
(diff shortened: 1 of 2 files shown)


tap/tcp_streams_map.go
```diff
@@ -37,7 +37,7 @@ func (streamMap *tcpStreamMap) nextId() int64 {
 func (streamMap *tcpStreamMap) closeTimedoutTcpStreamChannels() {
 	tcpStreamChannelTimeout := GetTcpChannelTimeoutMs()
 	for {
-		time.Sleep(10 * time.Millisecond)
+		time.Sleep(1000 * time.Millisecond)
```
gadotroee, line 40: maybe we can support some environment variable here? (which will be defaulted to 1000).
so in case we want to change back to 10 it will not require code change (but we probably need to add some validation to this env and not take it as is)

@mertyildiran  wdyt?

mertyildiran: @gadotroee when you have a constant value in programming, it generally means that you have made something wrong. `closeTimedoutTcpStreamChannels` wouldn't exist if we were not trying to dissect in real-time.

That value is determined by the programmer and optimized by taking many situations into account. I don't think we should give the user the ability to mess with this value and add yet another unknown parameter to the equation of failures. If we change this to an environment variable then we're going to add the below question to the list of questions that needs to be asked to the user who reports an issue;

> Did you set the `CLOSE_TIMEDOUT_TCP_STREAM_CHANNELS_INTERVAL_MS` environment variable? If you set it to `1` or `0`, the program will do nothing but executing `closeTimedoutTcpStreamChannels`. If you set it `9999999...` and you're getting OOM frequently, that's a too big of a value, make it small.

Also, this software is a monolith, you can't set the environment variable without altering the source. (unless providing an argument through CLI or assuming people will going to use Kubernetes manifests / `kubectl` or something similar to set it)

Alternatively, we can try to make this value dynamically determined. But we should be really smart on that and I'm not sure about the correct strategy.

haiut: @mertyildiran I agree with @gadotroee, for this parameter we need the ability to control it externally, without rebuilding the binary, docker image, and redeploying everything.. It's necessarily for the user, but for at least for our own flexibility and better troubleshooting. 

It would be great to have this value retrieved via env var, and if the env var doesn't exist - fall back to default value (1000 <-- the one you chose, for example) .. since env vars can't be changed during runtime, the code can retrieve its value at the beginning, do some safety checks (make sure the value in range of 10 .. 10,000, for example) and proceed..

Info log message should be printed when value from env var is used.
Warn log message should be printed when value found in env-var, but out of permitted range
If no env var is defined, use default value and no logs are shown.

mertyildiran: @haiut done with https://github.com/up9inc/mizu/commit/77833659b849705e170cc173aaa2a6677e3ca4b0

tap/tcp_streams_map.go
```diff
@@ -34,10 +36,34 @@ func (streamMap *tcpStreamMap) nextId() int64 {
...
+			} else {
+				return time.Duration(closeTimedoutTcpChannelsIntervalMs) * time.Millisecond
+			}
+		}
+	}
+}
+
 func (streamMap *tcpStreamMap) closeTimedoutTcpStreamChannels() {
 	tcpStreamChannelTimeout := GetTcpChannelTimeoutMs()
+	closeTimedoutTcpChannelsIntervalMs := streamMap.getCloseTimedoutTcpChannelsInterval()
```
gadotroee, line 64: @mertyildiran  do you think it will help to log this value just to be sure we know what it is? 
it can when we are debugging something so we won't guess or go to the code to find the value (especially when it can changed in the future (the default))

Imo, it will be nice to get message like this
`logger.Log.Infof("Using '%d'ms as the close tcp channels timeout", closeTimedoutTcpChannelsIntervalMs)`

Nice change with support the value and range I think it will help

mertyildiran: https://github.com/up9inc/mizu/commit/86ebeec09c43bbbe8300dae3e4ac26e7e90a41eb

gadotroee (approved):

PR #984, opened 2022-04-09
Upgrade Basenine version to `v0.7.1`

gadotroee:


mertyildiran (dismissed):

ui-common/src/hooks/useWS.tsx
```diff
@@ -7,7 +7,7 @@ enum WebSocketReadyState {
   CLOSED
 }
 
-export const DEFAULT_QUERY = "leftOff(-1)";
+export const DEFAULT_QUERY = `leftOff("latest")`;
```
gadotroee, line 10: where it is used? and why not use it in (https://github.com/up9inc/mizu/pull/984/files#diff-d6919ca03cedf72249cdb293e45ab5013085888488398f8a38da8baaef683c68R120)

ui-common/src/components/TrafficViewer/EntriesList.tsx
```diff
@@ -148,14 +148,14 @@ export const EntriesList: React.FC<EntriesListProps> = ({
         case "queryMetadata":
           setTruncatedTimestamp(message.data.truncatedTimestamp);
           setQueriedTotal(message.data.total);
-          if (leftOffTop === null) {
-            setLeftOffTop(message.data.leftOff - 1);
+          if (leftOffTop === "") {
+            setLeftOffTop(message.data.leftOff);
```
gadotroee, line 152: maybe need to decreament the id by one?

mertyildiran: We can't decrement it because it's a string now. This change is also wrong https://github.com/up9inc/mizu/pull/984/commits/2899414f2b82ed8fdb2e84a7b6aa00f8e8b8f5e0#diff-3a5137adc8ec2619a9214ef04911adfb8b3c3e76da4eb75169ee697b899d580aR58-R59

It's not always a zero padded string that's actually an integer. You can't do arithmetic operations or modifications on the `id` or `leftOff` value inside the front-end anymore.

mertyildiran: To clarify again, with this PR, the value of `id`, `leftOff` fields or `leftOff` helper argument can be;

- `000000000000000000000042`
- `507f1f77bcf86cd799439011`

gadotroee: Ok, so maybe I didn't understand it before. 
Anyway the issue with 1 is something that we don't understand something with the leftoff so I will try to explain it later and see if we can find a good option here (with the latest fixed we did)

agent/go.mod
```diff
@@ -149,3 +149,5 @@ replace github.com/up9inc/mizu/tap/extensions/http v0.0.0 => ../tap/extensions/h
 replace github.com/up9inc/mizu/tap/extensions/kafka v0.0.0 => ../tap/extensions/kafka
 
 replace github.com/up9inc/mizu/tap/extensions/redis v0.0.0 => ../tap/extensions/redis
+
+replace github.com/up9inc/basenine/client/go v0.0.0-20220326121918-785f3061c8ce => /home/mertyildiran/Documents/up9/basenine/client/go
```
gadotroee, line 153: I guess it should not be here?

mertyildiran: Removed with https://github.com/up9inc/mizu/pull/984/commits/6596a22fc0199daee8bb32405cdbddf74b89d328

ui-common/src/components/TrafficViewer/EntriesList.tsx
```diff
@@ -55,7 +55,8 @@ export const EntriesList: React.FC<EntriesListProps> = ({
   const [startTime, setStartTime] = useState(0);
   const [truncatedTimestamp, setTruncatedTimestamp] = useState(0);
 
-  const leftOffBottom = entries.length > 0 ? entries[entries.length - 1].id + 1 : -1;
+  const zeroPad = (num, places) => String(num).padStart(places, '0')
+  const leftOffBottom = entries.length > 0 ? zeroPad(+entries[entries.length - 1].id + 1) : "";
```
mertyildiran, line 59: Could you please revert this change? We need to understand caused `-1` in `develop` first. The front-end code shouldn't alter the index values.

mertyildiran (changes requested):

ui-common/src/components/TrafficViewer/EntriesList.tsx
gadotroee: sure I will revert it..

gadotroee: I did it with this commit https://github.com/up9inc/mizu/pull/984/commits/e1ba1cefbde0d11b4c2f3639e5513bab0fde814b I hope this is what you mean

ui-common/src/hooks/useWS.tsx
mertyildiran: https://github.com/up9inc/mizu/commit/93725264c34f1df6b81fd31eaa5dde56194ea5ab

ui-common/src/components/TrafficViewer/EntriesList.tsx
mertyildiran: https://github.com/up9inc/mizu/commit/9e1dccac501c2eae42bf9c76d58c1ebeeef14d3d

mertyildiran (approved):

agent/pkg/models/models.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type EntriesRequest struct {
-	LeftOff   int    `form:"leftOff" validate:"required,min=-1"`
+	LeftOff   string `form:"leftOff" validate:"required,min=-1"`
```
RoyUP9, line 16: min should be -1 after the change?

agent/pkg/oas/oas_generator.go
```diff
@@ -98,8 +101,14 @@ func (g *defaultOasGenerator) runGenerator() {
 	dataChan := make(chan []byte)
 	metaChan := make(chan []byte)
 
-	logger.Log.Infof("Querying DB for OAS generator with query '%s'", g.entriesQuery)
-	g.dbConn.Query(g.entriesQuery, dataChan, metaChan)
+	func() {
```
RoyUP9, line 104: why is this part inside a func?

agent/pkg/models/models.go
mertyildiran: I forgot to remove it. Removed with https://github.com/up9inc/mizu/commit/c2f5894de57c33d1aa6de1d4bdcda6fec3d62b4d

agent/pkg/api/socket_data_streamer.go
```diff
@@ -79,7 +79,9 @@ func (e *BasenineEntryStreamer) Get(ctx context.Context, socketId int, params *W
 	go handleDataChannel(connection, data)
 	go handleMetaChannel(connection, meta)
 
-	connection.Query(query, data, meta)
+	if err = connection.Query(query, data, meta); err != nil {
+		logger.Log.Errorf("Query mode call failed: %v", err)
```
RoyUP9, line 83: We don't need to return nil/switch to panic?

agent/pkg/oas/oas_generator.go
mertyildiran: It's from @undera's commits. `Query` call and the retrieval from `dataChan`/`metaChan` should run in parallel like in [this example](https://github.com/up9inc/basenine#query-1). But I'm not sure why there is a closure instead of a Goroutine.

RoyUP9: So what can we do? I don't think it's a good idea to leave it like that

agent/pkg/api/socket_data_streamer.go
mertyildiran: I'm afraid of crashing the API server based on user action but I agree that `Panicf` is required there. So, fixed with https://github.com/up9inc/mizu/commit/296b924460171166021e2c7b4c4a163594829713

agent/pkg/oas/oas_generator.go
mertyildiran: I'm not familiar with this OAS generator feature. But I understand that `runGenerator` is a Goroutine. So, it's written in a slightly different way than [the original example](https://github.com/up9inc/basenine#query-1). Therefore removed the closure with https://github.com/up9inc/mizu/commit/d93e00076d9be042ea1fc42597bc0917a5f874fb

ui-common/src/components/TrafficViewer/EntriesList.tsx
```diff
@@ -55,7 +55,7 @@ export const EntriesList: React.FC<EntriesListProps> = ({
   const [startTime, setStartTime] = useState(0);
   const [truncatedTimestamp, setTruncatedTimestamp] = useState(0);
 
-  const leftOffBottom = entries.length > 0 ? entries[entries.length - 1].id + 1 : -1;
+  const leftOffBottom = entries.length > 0 ? entries[entries.length - 1].id : "latest";
```
lirazyehezkel, line 58: I think we need the +1 so we won't have the same entry twice

lirazyehezkel (commented):
Added a comment

AmitUp9 (commented):
FE approved

ui-common/src/components/TrafficViewer/EntriesList.tsx
mertyildiran: It's now a string. I have moved that to the database server:
 - https://github.com/up9inc/basenine/blob/c741e4aa1cf2a7105d5cc5709d3c1686d66e23db/server/lib/storages/native.go#L1167
 - https://github.com/up9inc/basenine/blob/c741e4aa1cf2a7105d5cc5709d3c1686d66e23db/server/lib/storages/native.go#L396

RoyUP9 (approved):

PR #1000, opened 2022-04-13
Fix acceptance tests entry id is string and not number

gadotroee:


acceptanceTests/cypress/integration/testHelpers/TrafficHelper.js
```diff
@@ -111,6 +111,10 @@ export function checkFilterByMethod(funcDict) {
     });
 }
 
+export function getEntryId (id) {
+    return id.split('-')[1];
```
RoyUP9, line 115: comment

gadotroee: added in https://github.com/up9inc/mizu/pull/1000/commits/e817b1fc71aabc576ebb1afdd158403ca8eed5ee

RoyUP9 (approved):

PR #1003, opened 2022-04-13
Fix the acceptance tests `noMoreDataTop` issue

mertyildiran:


gadotroee:
@lirazyehezkel did some work with the leftOff so I prefer he will look in this too.
We will take a deeper look and handle the merges in the morning

lirazyehezkel (approved):

PR #1005, opened 2022-04-14
Link component

leon-up9:


ui-common/src/components/UI/index.ts
```diff
@@ -5,10 +5,10 @@ import Tooltip from "./Tooltip";
...
 import { StatusBar } from "./StatusBar";
 import CustomModal from "./CustomModal";
-import { InformationIcon } from "./InformationIcon";
+import { InformationIcon, Link } from "./InformationIcon";
 import SelectList from "./SelectList";
 import NoDataMessage from "./NoDataMessage";
 
 
-export { LoadingOverlay, Select, Tabs, Tooltip, Checkbox, CustomModal, InformationIcon, SelectList, NoDataMessage }
+export { LoadingOverlay, Select, Tabs, Tooltip, Checkbox, CustomModal, InformationIcon, SelectList, NoDataMessage, Link }
```
IgorGov, line 13: can 'StatusBar' be part of this export?

ui-common/src/components/UI/InformationIcon.tsx
```diff
@@ -3,17 +3,24 @@ import styles from "./style/InformationIcon.module.sass"
...
-            <span>Docs</span>
+        <a href={link} className={className} title={title} target="_blank">
+            {children}
         </a>
     </React.Fragment>
 }
 
-
+export const InformationIcon: React.FC<LinkProps> = ({ className }) => {
+    return <React.Fragment>
```
IgorGov, line 21: why do we need to use fragment here?
according to the documentation it is a tool for specific scenario which is not related in this case
https://reactjs.org/docs/fragments.html

ui-common/src/components/UI/InformationIcon.tsx
```diff
@@ -3,17 +3,24 @@ import styles from "./style/InformationIcon.module.sass"
...
+        <Link title="documentation" className={`${styles.linkStyle} ${className}`} link={DEFUALT_LINK}>
+            <span>Docs</span>
+        </Link>
+    </React.Fragment>
+}
```
IgorGov, line 26: new line

ui-common/src/components/UI/index.ts
leon-up9: yes technically,
its was a aesthetic separation for a easily identify UI components and others

lirazyehezkel (approved):

PR #1026, opened 2022-04-20
Refactor `tap` module to achieve synchronously closing other protocol dissectors upon identification

mertyildiran:
This PR defines `TcpReader` and `TcpStream` interfaces in the API and does refactor under `tap` module for calling the below function synchronously;

https://github.com/up9inc/mizu/blob/69538cb7e3f4bf5ab0385667bb49acbe41f9a4b2/tap/tcp_stream.go#L100-L124

Such that upon identification, it closes the other protocol dissectors directly without relying `closeTimedoutTcpStreamChannels` periodic job.

Related PRs:

- https://github.com/up9inc/mizu/pull/257
- https://github.com/up9inc/mizu/pull/277
- https://github.com/up9inc/mizu/pull/260
- https://github.com/up9inc/mizu/pull/979


tap/api/tcp_streams_map.go
```diff
@@ -0,0 +1,60 @@
...
+	streamMap.Streams.Store(key, value)
+}
+
+func (streamMap *TcpStreamMap) Delete(key interface{}) {
+	streamMap.Streams.Delete(key)
+}
+
+func (streamMap *TcpStreamMap) NextId() int64 {
+	streamMap.streamId++
+	return streamMap.streamId
```
gadotroee, line 34: do we need any locker here to be sure this method to be "atomic"? (maybe I miss something important that prevent this call to be in a way that return the same id to 2 different calls)

gadotroee (commented):
I have tried to add comments of what I know about.
anyway it worth to get review here from @dvdlevanon  / @nimrod-up9  (at least one of them).

tap/passive_tapper.go
```diff
@@ -31,11 +32,8 @@ var maxcount = flag.Int64("c", -1, "Only grab this many packets, then exit")
 var decoder = flag.String("decoder", "", "Name of the decoder to use (default: guess from capture)")
 var statsevery = flag.Int("stats", 60, "Output statistics every N seconds")
 var lazy = flag.Bool("lazy", false, "If true, do lazy decoding")
-var nodefrag = flag.Bool("nodefrag", false, "If true, do not do IPv4 defrag")
-var checksum = flag.Bool("checksum", false, "Check TCP checksum")                                                      // global
-var nooptcheck = flag.Bool("nooptcheck", true, "Do not check TCP options (useful to ignore MSS on captures with TSO)") // global
-var ignorefsmerr = flag.Bool("ignorefsmerr", true, "Ignore TCP FSM errors")                                            // global
```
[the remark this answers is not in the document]
mertyildiran: Yeah, those flags are used in the code without argument propagation so a Go module change made them unusable as a flag. As far as I know, we're not setting those flags in CLI.

If you want to keep them, we need to pass them to all the related functions in `tap/api`.

tap/api/tcp_streams_map.go
mertyildiran: I didn't add that function, I just moved it from [here](https://github.com/up9inc/mizu/pull/1026/files#diff-2ed99ead342ab539b2997815d8fe2f3a1b1f5a0bf52997a1d1ae9609fe772141L34-L37) to there. It's only called in [here](https://github.com/up9inc/mizu/blob/fix/better-close-other-dissectors/tap/tcp_stream_factory.go#L73) and those calls are synchronous.

tap/settings.go
```diff
@@ -12,11 +11,8 @@ const (
 	MemoryProfilingTimeIntervalSeconds        = "MEMORY_PROFILING_TIME_INTERVAL"
 	MaxBufferedPagesTotalEnvVarName           = "MAX_BUFFERED_PAGES_TOTAL"
 	MaxBufferedPagesPerConnectionEnvVarName   = "MAX_BUFFERED_PAGES_PER_CONNECTION"
-	TcpStreamChannelTimeoutMsEnvVarName       = "TCP_STREAM_CHANNEL_TIMEOUT_MS"
-	CloseTimedoutTcpChannelsIntervalMsEnvVar  = "CLOSE_TIMEDOUT_TCP_STREAM_CHANNELS_INTERVAL_MS"
 	MaxBufferedPagesTotalDefaultValue         = 5000
 	MaxBufferedPagesPerConnectionDefaultValue = 5000
-	TcpStreamChannelTimeoutMsDefaultValue     = 10000
```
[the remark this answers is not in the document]
mertyildiran: While separating them, I imagined;
 - the env vars and default values related to `tap` module goes into `tap/settings.go`
 - the env vars and default values related to `tap/api` module goes into `tap/api/settings.go`
 
Everything goes into `tap/api/settings.go` would also work though.

tap/passive_tapper.go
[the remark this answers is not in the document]
mertyildiran: Brought back those flags with https://github.com/up9inc/mizu/commit/0c5678976c71ae46996f18aef30ec0631d96f71d

tap/settings.go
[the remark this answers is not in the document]
gadotroee: yes it sounds like a good separation as long as we know what is inside "tap/api/settings" and what is inside "tap/settings".

I think it make sense to move to tap/api/settings but I want another one to go over and see all other code make sense and then we can decide if we want to move it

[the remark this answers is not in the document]
mertyildiran: OK then I'll be waiting for the decision.

tap/api/tcp_stream.go
```diff
@@ -1,43 +1,45 @@
...
 
 /* It's a connection (bidirectional)
  * Implements gopacket.reassembly.Stream interface (Accept, ReassembledSG, ReassemblyComplete)
  * ReassembledSG gets called when new reassembled data is ready (i.e. bytes in order, no duplicates, complete)
- * In our implementation, we pass information from ReassembledSG to the tcpReader through a shared channel.
+ * In our implementation, we pass information from ReassembledSG to the TcpReader through a shared channel.
  */
-type tcpStream struct {
-	id              int64
+type TcpStream struct {
```
dvdlevanon, line 19: I see we moved TcpStream and TcpReader to the api package, which is a good initiative. However, currently those structs coupled with the traditional tapper and gopacket. There is no clean way to use them from other tappers, like TLS and future tappers.

For example, there is no `TcpState *reassembly.TCPSimpleFSM` in TLS, and faking it is ugly. Indeed many members of those structs are not relevant for TLS.

Maybe it need to be an interface, with some public functions, allowing every implementation to implement its specific.

dvdlevanon (changes requested):

tap/api/tcp_stream.go
mertyildiran: The struct is moved to `tap/api` package without changing its fields so much. `tap/api` package is the package shared between `tap` and the protocol dissectors. I couldn't understand the example of `TcpState        *reassembly.TCPSimpleFSM` field. That field is not populated in here and it does not need to be populated:

https://github.com/up9inc/mizu/blob/0c5678976c71ae46996f18aef30ec0631d96f71d/tap/tlstapper/tls_poller.go#L171-L174

I agree that `TcpStream` struct has fields which are specific to `gopacket`. I think there can be a better abstraction based on commonalities between the network sniffer and tracers. But this PR is not aiming to achieve that abstraction. It rather moves the code from here to there.

dvdlevanon: Yes, I understand the PR doesn't introduced those structs. Yet moving a struct to an "api" package, make it an api.

mertyildiran: Is the problem exported fields? If so, we need define some functions to make them private again.

dvdlevanon: If the motivation of the PR is to allow the extensions to call to `CloseOtherProtocolDissectors`, we should define an interface with this method, instead of exposing a lot of internal logic.

In tls, we have tls_reader.go, which is the equivalent to tcp_reader.go. After this change, tls suddenly need to have both, without a good reason. 

---

I still think that we should unify the tcp_reader and tls_reader, and move the core logic to the api. But this should be done in a generic way, not just by copy paste a single implementation to the api.

mertyildiran: As long as there is this logic:

https://github.com/up9inc/mizu/blob/0c5678976c71ae46996f18aef30ec0631d96f71d/tap/passive_tapper.go#L72-L75

The eBPF - OpenSSL tracer has nothing to do with the protocol dissector API. Because it's not complying with the dissection of multiple application layer protocols criteria. I would first remove that logic then try to unify everything under `tap/api`. But that's off-topic for this PR.

If you think the exported structs `TcpReaderDataMsg`, `TcpReader` and `TcpStream` and their exported fields are too much. For the sake of good code, we can make them private and export only the methods.

> There's no performance difference between exported and unexported fields/methods/etc.

As I stated above, the TLS tapper is neither compliant with multiple protocol support nor it's compliant with the `TcpStream` struct. So, I don't see a clear interface that unifies everything at this moment which can easily be implemented. Even if there is such an interface possible, it's outside of the scope of this PR.

dvdlevanon: The tls use the http extension via the extension api, It doesn't matter if it use a single extension or many of them.
Those 3 struct relevant to pcap capture only - TcpReaderDataMsg, TcpReader and TcpStream.

Unification, while possible, takes time.

 ---

The whole motivation of the PR is to be able to call `CloseOtherProtocolDissectors` from the extensions, which is possible using an interface. I can't understand why we need to expose and break the encapsulation

dvdlevanon: This commit converted pcap specific structs to pcap specific interfaces, its better to stay with the structs.

A single interface, like this once, may do the work without exposing pcap logic to the api.
```
type ExtensionHandler interface {
	Close(Protocol protocol)
}
```

The Dissect method may get it as another parameter, keeping the code the exact way it was, while achieving the same goal.

mertyildiran: or maybe keep the interfaces in `tap/api` but move the structs to `tap`?

mertyildiran: Did that in here https://github.com/up9inc/mizu/commit/7543db557c6a76835632fc4266146d5fda7193d5 but ended up with an import cycle:

```
building mizu agent ..
package command-line-arguments
	imports github.com/up9inc/mizu/tap
	imports github.com/up9inc/mizu/tap/tlstapper
	imports github.com/up9inc/mizu/tap: import cycle not allowed
make: *** [Makefile:36: agent] Error 1
```

so I had to move them to a new package called `tap/tcp`.

`CloseOtherProtocolDissectors` needs to be a pointer receiver which holds references to all the client and server `TcpReader`s such that it can close all of the other `TcpReaderDataMsg` channels. That indicates `TcpStream` is the most suitable struct to have that method. That's why `TcpStream` is passed to `Dissect` and its method signature is changed.

mertyildiran: That required the protocol dissectors being depend on `tap` package here https://github.com/up9inc/mizu/commit/367a812ec64359db76bf1180c3dd20003f411f76 So, I guess I will have to revert https://github.com/up9inc/mizu/commit/7543db557c6a76835632fc4266146d5fda7193d5 and https://github.com/up9inc/mizu/commit/367a812ec64359db76bf1180c3dd20003f411f76

dvdlevanon: I'll try to elaborate.

In `tap/api/api.go` add 
```
type ExtensionHandler interface {
	Close(protocol Protocol)
}
```
Changed `Dissector.Dissect` signature to looks like 
```
Dissect(b *bufio.Reader, progress *ReadProgress, capture Capture, isClient bool, tcpID *TcpID, counterPair *CounterPair, superTimer *SuperTimer, superIdentifier *SuperIdentifier, emitter Emitter, options *TrafficFilteringOptions, reqResMatcher RequestResponseMatcher, handler ExtensionHandler) error
```

In `tap/tcp_reader.go` introduce a new struct 
```
type tcpExtensionHandler struct {
	reader *tcpReader
}

func (t *tcpExtensionHandler) Close(protocol api.Protocol) {
	fmt.Printf("Close the extension using both protocl + reader information")
}
```

Then change the call dissect like this:
```
h.extension.Dissector.Dissect(b, h.progress, h.parent.origin, h.isClient, h.tcpID, h.counterPair, h.superTimer,
		h.parent.superIdentifier, h.emitter, filteringOptions, h.reqResMatcher, &tcpExtensionHandler{reader: h})
```

---

Its just an example, names, locations etc may be anywhere. But the idea is to be able to let the extension communicate its status, without the need to give it all the internal information.

mertyildiran: @dvdlevanon but close happens on arrays of TCP readers:

```
	clients         []api.TcpReader
	servers         []api.TcpReader
```

It's not just single reader:


```go
func (t *tcpStream) CloseOtherProtocolDissectors(protocol *api.Protocol) {
	t.Lock()
	defer t.Unlock()

	if t.protoIdentifier.IsClosedOthers {
		return
	}

	t.protoIdentifier.Protocol = protocol

	for i := range t.clients {
		reader := t.clients[i]
		if reader.GetExtension().Protocol != t.protoIdentifier.Protocol {
			reader.Close()
		}
	}
	for i := range t.servers {
		reader := t.servers[i]
		if reader.GetExtension().Protocol != t.protoIdentifier.Protocol {
			reader.Close()
		}
	}

	t.protoIdentifier.IsClosedOthers = true
}
```

dvdlevanon: If that's the case, we can give `tcpExtensionHandler` the tcpStream instead of tcpReader. We can also initialize it in the tcpStream instead of tcpReader.

mertyildiran: @dvdlevanon OK then what's the difference between passing `tcpStream` inside `tcpExtensionHandler` argument or passing `tcpStream` argument?

dvdlevanon: We don't pass `tcpStream` insdie `tcpExtensionHandler`.
We pass `ExtensionHandler interface`, which is an interface containing a single method.

For the API this interface should be opaque, they don't know and shouldn't count on its concrete implementation.

mertyildiran: @dvdlevanon there were 11 arguments in the `Dissect` method signature. With that it becomes 12. Most of the arguments are actually either sourced from `tcpStream` or `tcpReader` that's why I decided to pass a `tcpReader` instance.

I don't think with;

```
h.extension.Dissector.Dissect(b, h.progress, h.parent.origin, h.isClient, h.tcpID, h.counterPair, h.superTimer,
		h.parent.superIdentifier, h.emitter, filteringOptions, h.reqResMatcher, &tcpExtensionHandler{reader: h})
```

It will be cleaner.

mertyildiran: I mean if we're going to pass the reader with `&tcpExtensionHandler{reader: h})` anyway and the reader holds the information of all the other arguments then why do we just don't pass the reader like it's already implemented in this PR?

dvdlevanon: Yes I agree we should do a refactor and create a struct for those arguments.
This struct should be generic and shouldn't contain any pcap/tls related members.

If we have the time to do this refactor, great. If not, a 12th param is better than breaking the encapsulation.

mertyildiran: @dvdlevanon if you think that;

https://github.com/up9inc/mizu/blob/9aed5b294129e69f4fcce5bcfd9f0bbc3f90f433/tap/tlstapper/tls_reader.go#L10-L16

should implement;

https://github.com/up9inc/mizu/blob/9aed5b294129e69f4fcce5bcfd9f0bbc3f90f433/tap/api/api.go#L415-L432

I can do it as well. If I do that, is it OK?

dvdlevanon: What `GetParent() TcpStream` returns for TlsReader?

mertyildiran: @dvdlevanon returns `TcpStream`

https://github.com/up9inc/mizu/blob/9aed5b294129e69f4fcce5bcfd9f0bbc3f90f433/tap/api/api.go#L434-L450

dvdlevanon: This breaks the encapsulation, again.

tcpStream contains information not relevant to extensions, and moreover, if they'll accidentally use it it may break tls.

The fact it called `dummy` by itself, proves the point.

mertyildiran: > tcpStream contains information not relevant to extensions, and moreover, if they'll accidentally use it it may break tls.

It's relevant to the protocol dissectors because they elevate into TCP stream level and close the other readers with `CloseOtherProtocolDissectors` call.

https://github.com/up9inc/mizu/blob/9aed5b294129e69f4fcce5bcfd9f0bbc3f90f433/tap/api/api.go#L435-L437

are required by [this interface](https://github.com/google/gopacket/blob/3eaba08943250fd212520e5cff00ed808b8fc60a/reassembly/tcpassembly.go#L362-L380) defined in `gopacket`.

> The fact it called `dummy` by itself, proves the point.

It's suffixed `Dummy` because the `tlstapper` package does not make any use of it other than setting the `origin` field.

mertyildiran: Upon looking at again, I think this;

https://github.com/up9inc/mizu/blob/2f1e98d7fe226ed0243c72b09713b7ead70f5a01/tap/tlstapper/tls_poller.go#L25-L34

corresponds to;

https://github.com/up9inc/mizu/blob/2f1e98d7fe226ed0243c72b09713b7ead70f5a01/tap/api/api.go#L436-L452

since it holds references to `api.TcpReader`;

https://github.com/up9inc/mizu/blob/2f1e98d7fe226ed0243c72b09713b7ead70f5a01/tap/tlstapper/tls_poller.go#L27

just like;

https://github.com/up9inc/mizu/blob/2f1e98d7fe226ed0243c72b09713b7ead70f5a01/tap/tcp/tcp_stream.go#L37-L38

also;

https://github.com/up9inc/mizu/blob/2f1e98d7fe226ed0243c72b09713b7ead70f5a01/tap/tlstapper/tls_poller.go#L29

vs.

https://github.com/up9inc/mizu/blob/2f1e98d7fe226ed0243c72b09713b7ead70f5a01/tap/tcp/tcp_stream.go#L41

So, I can make `tlsPoller` implement `TcpStream`? WDYT?

dvdlevanon: `It's relevant to the protocol dissectors because they elevate into TCP stream level and close the other readers with CloseOtherProtocolDissectors call.`

With the interface they achieve the same goal, without actually closing them by themselves. Thats why interfaces invented.

---

The fact that `tcpStream` needs to implement `gopacket` internals, is yet another prove why it shouldn't be part of the api.

---

`It's suffixed Dummy because the tlstapper package does not make any use of it other than setting the origin field`

Yet is another prove why it shouldn't be part of the api

mertyildiran: Adding yet another argument and defining a specific, smaller interface achieves the same goal. I didn't say it does not achieve the same goal. But TCP is a common property for the network sniffer, eBPF tracer and protocol dissectors. So, the TCP related definitions can reside in `tap/api` IMHO.

`TcpStream` interface defines methods that're related to TCP reassembly which is again a common property for the network sniffer and eBPF tracer (because they both should build the TCP stream). I guess `gopacket`'s TCP reassembly is not used in `tlstapper` package because the data/packets does not come in out of order. So, those methods not being required inside `tlstapper` does not make them completely irrelevant.

dvdlevanon: OK that's make sense. 
Let's go with it.

mertyildiran: With https://github.com/up9inc/mizu/commit/97072427f67b8c00982a1abc4c99060eab8971e4 `tap/api` does not depend on `gopacket` and `gopacket` specific methods are now removed from `TcpStream` interface. So, it looks like this now;

https://github.com/up9inc/mizu/blob/97072427f67b8c00982a1abc4c99060eab8971e4/tap/api/api.go#L433-L449

mertyildiran: @dvdlevanon with https://github.com/up9inc/mizu/commit/b0fbaec2f4b534421775975326ff50615df2ab48 I have made `tlsPoller` implement `TcpStream` interface and remove the call to `NewTcpStreamDummy` method.

tap/tcp/tcp_reader.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (reader *tcpReader) SendMsgIfNotClosed(msg api.TcpReaderDataMsg) {
+	reader.Lock()
+	if !reader.isClosed {
+		reader.msgQueue <- msg
+	}
+	reader.Unlock()
+}
+
+func (reader *tcpReader) SendChunk(chunk api.TlsChunk) {}
```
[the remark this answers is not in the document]
mertyildiran: @nimrod-up9 it does not really implement it since it has an empty function body. `TcpReader` defines the necessary method signatures for both `tcpReader` and `tlsReader`. We can try to further unify them with a method like `SendMsg(msg interface{}` or defining intermediate interfaces. But right now `TcpReader` is valid and encompasses all the necessary functionality.

nimrod-up9 (approved):

dvdlevanon (approved):

PR #1034, opened 2022-04-21
Fix/ui/Service-map-GUI-issues_TRA-4499

leon-up9:


ui-common/src/components/style/LoadingOverlay.module.sass
```diff
@@ -8,6 +8,7 @@
     align-items: center
     justify-content: center
     background-color: rgba(25, 25, 25, 0.5)
+    z-index: 1000
```
lirazyehezkel, line 11: This change makes sense to me, but are we sure we want this change?
we use this component all over the code and I'm afraid it might break something...
Is this a must for a feature you worked on?

leon-up9: it should fix the loader when you save the general settings.
your call

lirazyehezkel (approved):

PR #1053, opened 2022-04-28
Feature/reorginize some tls code after refactor

dvdlevanon:
This PR fixes tls, which currently broken in develop.

The tls is broken because of:
1. panic because `GetReqResMatcher()` returns nil
2. tls readers remain opened because `setProtocol` does nothing
3. tls not recognized because `GetProtoIdentifier()` return nothing

Bullets 2 and 3 are because tls doesn't implement TcpStream right. And it doesn't implement it right because that in tls there is no such concept like TcpStream and TcpReader, there is only a TlsReader. The refactor tries to make TlsPoller to implement and behave like a TcpStream, but it can't. There is only a single instance of TlsPoller in the program, and there is no way to let it implement TcpStream, which is per stream.

The only reason why it had to implement TcpStream in the first place, is because the extensions do something like `reader.GetParent().XXX`. And this is wrong because:
1. It couples the extensions to pcap implementation, where it has a concept of a parent.
2. It forces tls to implement a "fake" or "dummy" TcpStream like interface.

A better solution would be to call `reader.XXX` from the extensions, the pcap implementation can delegate the method to the parent, while other implementations can do what they need to do, without the need of creating a fake class.

So the first thing the PR do, beside solving the panic, is to get rid of TcpStream, and implement `SetProtocol` and `GetProtoIdentifier` for tls in TlsReader.

---

The other thing this PR do, and it have to do it as part of this PR, is to stop using interfaces in the concrete implementation. Take a look at the function `NewTlsReader` in develop:
1. It public and there is not reason for making it public
2. It returns an interface, forcing the whole internal code to work with an interface, why?
3. Because it forces to work with interfaces, there are various casts in the code, why?

The whole concept of interface, is to be an interface across components. There is no need to use the interface in the package that create it, it just cause unnecessary casts and expose unnecessary logic to the api.

So this second thing this PR do, is to use concrete implementation internally, making the code cleaner and easier to write and read.

### DIFF
--- a/tap/api/api.go
+++ b/tap/api/api.go
@@ -413,22 +413,15 @@ type TcpReader interface {
 	GetReqResMatcher() RequestResponseMatcher
 	GetIsClient() bool
 	GetReadProgress() *ReadProgress
-	GetParent() TcpStream
 	GetTcpID() *TcpID
 	GetCounterPair() *CounterPair
 	GetCaptureTime() time.Time
 	GetEmitter() Emitter
 	GetIsClosed() bool
 	GetExtension() *Extension
-}
-
-type TcpStream interface {
 	SetProtocol(protocol *Protocol)
-	GetOrigin() Capture
 	GetProtoIdentifier() *ProtoIdentifier
-	GetReqResMatcher() RequestResponseMatcher
-	GetIsTapTarget() bool
-	GetIsClosed() bool
+	GetOrigin() Capture
 }
 
 type TcpStreamMap interface {
--- a/tap/tcp_stream_factory.go
+++ b/tap/tcp_stream_factory.go
@@ -61,15 +61,14 @@ func (factory *tcpStreamFactory) New(net, transport gopacket.Flow, tcpLayer *lay
 	stream := NewTcpStream(isTapTarget, factory.streamsMap, getPacketOrigin(ac))
 	reassemblyStream := NewTcpReassemblyStream(fmt.Sprintf("%s:%s", net, transport), tcpLayer, fsmOptions, stream)
 	if stream.GetIsTapTarget() {
-		_stream := stream.(*tcpStream)
-		_stream.setId(factory.streamsMap.NextId())
+		stream.setId(factory.streamsMap.NextId())
 		for i, extension := range extensions {
 			reqResMatcher := extension.Dissector.NewResponseRequestMatcher()
 			counterPair := &api.CounterPair{
 				Request:  0,
 				Response: 0,
 			}
-			_stream.addClient(
+			stream.addClient(
 				NewTcpReader(
 					make(chan api.TcpReaderDataMsg),
 					&api.ReadProgress{},
@@ -90,7 +89,7 @@ func (factory *tcpStreamFactory) New(net, transport gopacket.Flow, tcpLayer *lay
 					reqResMatcher,
 				),
 			)
-			_stream.addServer(
+			stream.addServer(
 				NewTcpReader(
 					make(chan api.TcpReaderDataMsg),
 					&api.ReadProgress{},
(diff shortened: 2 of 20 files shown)


tap/tcp_stream_factory.go
```diff
@@ -61,15 +61,14 @@ func (factory *tcpStreamFactory) New(net, transport gopacket.Flow, tcpLayer *lay
...
-		_stream.setId(factory.streamsMap.NextId())
+		stream.setId(factory.streamsMap.NextId())
 		for i, extension := range extensions {
 			reqResMatcher := extension.Dissector.NewResponseRequestMatcher()
 			counterPair := &api.CounterPair{
 				Request:  0,
 				Response: 0,
 			}
-			_stream.addClient(
+			stream.addClient(
```
nimrod-up9, line 71: This is only a change in style.
I think we should decide on a convention as a time instead of running over each other's code.

nimrod-up9 (commented):
Could you please explain what this PR fixes?

tap/api/api.go
```diff
@@ -413,22 +413,15 @@ type TcpReader interface {
...
-	GetParent() TcpStream
 	GetTcpID() *TcpID
 	GetCounterPair() *CounterPair
 	GetCaptureTime() time.Time
 	GetEmitter() Emitter
 	GetIsClosed() bool
 	GetExtension() *Extension
-}
-
-type TcpStream interface {
```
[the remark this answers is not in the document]
dvdlevanon: In `tls` there is no concept of `TcpStream`, in the end the extensions need `TcpReader` only.

tap/tcp_stream_factory.go
dvdlevanon: What do you mean only change in style?

Before there were both `_stream` and `stream`, one was the concrete struct while the other was the interface.
Now, we only use the concrete struct, as there is no need to use the `interface` internally.

tap/tlstapper/tls_poller.go
```diff
@@ -135,24 +126,12 @@ func (p *tlsPoller) handleTlsChunk(chunk *tlsChunk, extension *api.Extension,
...
-
 	if !exists {
-		reader = p.startNewTlsReader(chunk, ip, port, key, extension, newReader, options)
+		reader = p.startNewTlsReader(chunk, ip, port, key, emitter, extension, options)
 		p.readers[key] = reader
 	}
 
-	tlsReader := reader.(*tlsReader)
-
-	tlsReader.setCaptureTime(time.Now())
```
mertyildiran, line 154: The place where we call `time.Now()` was there;

https://github.com/up9inc/mizu/commit/b5c665b60236b7868833b2c1d31a7510e1b54ddc#diff-2c7816ba77c385ccf0193c52e07bdc1a012be29c6c0f33389735f77d2a9f1ea2R134

Now it's moved to somewhere else. Is it intended?

tap/tlstapper/tls_poller.go
```diff
@@ -279,27 +273,3 @@ func (p *tlsPoller) logTls(chunk *tlsChunk, ip net.IP, port uint16) {
 		srcIp, srcPort, dstIp, dstPort,
 		chunk.Recorded, chunk.Len, chunk.Start, str, hex.EncodeToString(chunk.Data[0:chunk.Recorded]))
 }
-
-func (p *tlsPoller) SetProtocol(protocol *api.Protocol) {
-	// TODO: Implement
```
mertyildiran, line 284: The existence and implementation of this method might be useful in the future once this logic is removed;

https://github.com/up9inc/mizu/blob/9bddfc19376684a66d0dbea5e21207bfe70d9a23/tap/passive_tapper.go#L74-L77

tap/tcp_stream_factory.go
nimrod-up9: Sorry, missed it.

tap/tlstapper/tls_poller.go
dvdlevanon: Correct, I just moved it to another place, because tlsPoller is global and not per stream

tap/tlstapper/tls_poller.go
dvdlevanon: No, it wasn't.
Thanks, I put it back there.

PR #1057, opened 2022-04-30
Fix `panic: interface conversion: api.RequestResponseMatcher is nil, not *http.requestResponseMatcher` error

mertyildiran:
Fixes the error below that's introduced by https://github.com/up9inc/mizu/pull/1026

```
panic: interface conversion: api.RequestResponseMatcher is nil, not *http.requestResponseMatcher

goroutine 74 [running]:
github.com/up9inc/mizu/tap/extensions/http.dissecting.Dissect({0xc0000716c0, 0xc0000716c0}, 0x44f2f2, {0x2815250, 0xc006897340}, 0xc006897301)
	/app/tap/extensions/http/main.go:90 +0xd3f
github.com/up9inc/mizu/tap/tlstapper.dissect(0xc000df4870, {0x2815250, 0xc006897340}, 0x0)
	/app/tap/tlstapper/tls_poller.go:185 +0x1a6
created by github.com/up9inc/mizu/tap/tlstapper.(*tlsPoller).startNewTlsReader
	/app/tap/tlstapper/tls_poller.go:177 +0x2ef
```

Also fixes the request-response matcher maps iteration in `clean()` method.

**No stable releases are affected.**

### DIFF
--- a/tap/api/api.go
+++ b/tap/api/api.go
@@ -426,7 +426,7 @@ type TcpStream interface {
 	SetProtocol(protocol *Protocol)
 	GetOrigin() Capture
 	GetProtoIdentifier() *ProtoIdentifier
-	GetReqResMatcher() RequestResponseMatcher
+	GetReqResMatchers() []RequestResponseMatcher
 	GetIsTapTarget() bool
 	GetIsClosed() bool
 }
--- a/tap/cleaner.go
+++ b/tap/cleaner.go
@@ -34,12 +34,14 @@ func (cl *Cleaner) clean() {
 	cl.assemblerMutex.Unlock()
 
 	cl.streamsMap.Range(func(k, v interface{}) bool {
-		reqResMatcher := v.(api.TcpStream).GetReqResMatcher()
-		if reqResMatcher == nil {
-			return true
+		reqResMatchers := v.(api.TcpStream).GetReqResMatchers()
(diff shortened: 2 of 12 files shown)


tap/api/api.go
```diff
@@ -426,7 +426,7 @@ type TcpStream interface {
 	SetProtocol(protocol *Protocol)
 	GetOrigin() Capture
 	GetProtoIdentifier() *ProtoIdentifier
-	GetReqResMatcher() RequestResponseMatcher
```
gadotroee, line 429: why do we need to change this to array of matchers? 
if the exception is casting the request response matcher to the interface because it is nil can't we just add condition for the "nil" of the value before casting? 

maybe there is another reason that I don't see now and you can explain.

gadotroee:
I think this is not needed after https://github.com/up9inc/mizu/pull/1053 (which solve the panic by fixing the tls issues and will not get to this error)

mertyildiran:
> I think this is not needed after #1053 (which solve the panic by fixing the tls issues and will not get to this error)

https://github.com/up9inc/mizu/pull/1053 is reverting a bunch of changes that are introduced with https://github.com/up9inc/mizu/pull/1026 and not fixing the `clean` method in `tap/cleaner.go`.

mertyildiran:
Closing in favor of https://github.com/up9inc/mizu/pull/1058

PR #1058, opened 2022-05-01
fix tls + creating tls_stream

dvdlevanon:


mertyildiran (approved):
Looking much better with the concrete types instead of abstract ones. The network sniffer side working fine. Couldn't test the TLS side but assuming the PR fixes it.

PR #1062, opened 2022-05-01
Spawn only two Goroutines per TCP stream

mertyildiran:
With this PR the number of Goroutines does not scale with the number the protocol dissectors and it will only spawn two Goroutines per TCP stream.

### DIFF
--- a/tap/tcp_reader.go
+++ b/tap/tcp_reader.go
@@ -3,11 +3,9 @@ package tap
 import (
 	"bufio"
 	"io"
-	"io/ioutil"
 	"sync"
 	"time"
 
-	"github.com/up9inc/mizu/logger"
 	"github.com/up9inc/mizu/tap/api"
 )
 
@@ -23,44 +21,44 @@ type tcpReader struct {
 	isClient      bool
 	isOutgoing    bool
 	msgQueue      chan api.TcpReaderDataMsg // Channel of captured reassembled tcp payload
+	buffer        []byte
+	exhaustBuffer bool
 	data          []byte
 	progress      *api.ReadProgress
 	captureTime   time.Time
 	parent        *tcpStream
 	packetsSeen   uint
-	extension     *api.Extension
 	emitter       api.Emitter
 	counterPair   *api.CounterPair
 	reqResMatcher api.RequestResponseMatcher
 	sync.Mutex
 }
 
-func NewTcpReader(msgQueue chan api.TcpReaderDataMsg, progress *api.ReadProgress, ident string, tcpId *api.TcpID, captureTime time.Time, parent *tcpStream, isClient bool, isOutgoing bool, extension *api.Extension, emitter api.Emitter, counterPair *api.CounterPair, reqResMatcher api.RequestResponseMatcher) *tcpReader {
+func NewTcpReader(msgQueue chan api.TcpReaderDataMsg, progress *api.ReadProgress, ident string, tcpId *api.TcpID, captureTime time.Time, parent *tcpStream, isClient bool, isOutgoing bool, emitter api.Emitter) *tcpReader {
 	return &tcpReader{
-		msgQueue:      msgQueue,
-		progress:      progress,
-		ident:         ident,
-		tcpID:         tcpId,
-		captureTime:   captureTime,
-		parent:        parent,
-		isClient:      isClient,
-		isOutgoing:    isOutgoing,
-		extension:     extension,
-		emitter:       emitter,
-		counterPair:   counterPair,
-		reqResMatcher: reqResMatcher,
+		msgQueue:    msgQueue,
+		progress:    progress,
+		ident:       ident,
+		tcpID:       tcpId,
+		captureTime: captureTime,
+		parent:      parent,
+		isClient:    isClient,
+		isOutgoing:  isOutgoing,
+		emitter:     emitter,
 	}
 }
 
 func (reader *tcpReader) run(options *api.TrafficFilteringOptions, wg *sync.WaitGroup) {
 	defer wg.Done()
-	b := bufio.NewReader(reader)
-	err := reader.extension.Dissector.Dissect(b, reader, options)
-	if err != nil {
-		_, err = io.Copy(ioutil.Discard, reader)
-		if err != nil {
-			logger.Log.Errorf("%v", err)
+	for i, extension := range extensions {
+		reader.reqResMatcher = reader.parent.reqResMatchers[i]
+		reader.counterPair = reader.parent.counterPairs[i]
+		b := bufio.NewReader(reader)
+		extension.Dissector.Dissect(b, reader, options)
+		if reader.parent.protocol != nil {
+			break
 		}
+		reader.exhaustBuffer = true
 	}
 }
 
@@ -81,7 +79,17 @@ func (reader *tcpReader) sendMsgIfNotClosed(msg api.TcpReaderDataMsg) {
 	reader.Unlock()
 }
 
+func (reader *tcpReader) isProtocolIdentified() bool {
+	return reader.parent.protocol != nil
+}
+
 func (reader *tcpReader) Read(p []byte) (int, error) {
+	if reader.exhaustBuffer {
+		l := copy(p, reader.buffer)
+		reader.exhaustBuffer = false
+		return l, nil
+	}
+
 	var msg api.TcpReaderDataMsg
 
 	ok := true
@@ -101,6 +109,9 @@ func (reader *tcpReader) Read(p []byte) (int, error) {
 	}
 
 	l := copy(p, reader.data)
+	if !reader.isProtocolIdentified() {
+		reader.buffer = append(reader.buffer, reader.data...)
+	}
 	reader.data = reader.data[l:]
 	reader.progress.Feed(l)
 
@@ -141,8 +152,4 @@ func (reader *tcpReader) GetEmitter() api.Emitter {
 
 func (reader *tcpReader) GetIsClosed() bool {
 	return reader.isClosed
-}
-
-func (reader *tcpReader) GetExtension() *api.Extension {
-	return reader.extension
 }
--- a/tap/api/api.go
+++ b/tap/api/api.go
@@ -104,11 +104,6 @@ type OutputChannelItem struct {
 	Namespace      string
 }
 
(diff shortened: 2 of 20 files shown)


nimrod-up9:
Copied from slack.

I think that we could end up skipping some bytes.
Suppose we got a Kafka message and that the HTTP extension just quit.
Let’s say this is the state of the `tcpReader`:
```go
reader.exhaustBuffer == true
reader.buffer == {1, 2, 3, 4}
reader.data == {5, 6, 7, 8}
```
Kafka extension calls `Read`. For some reason it gets only the first two bytes `(l == 2)`. `reader.exhaustBuffer == false`.
The extension calls `Read` again. This time it reads from `reader.data` and not from `reader.buffer` and returns 4 bytes.
The Kafka extension now holds the buffer `{1, 2, 5, 6, 7, 8}`.

Proposed solution: Always read from the same buffer `reader.data`, but to have the ability to rewind it. Instead of setting `reader.exhaustBuffer = true` after an extension quits we could call `reader.Rewind` (or some other name).

I think this is a start. Do not remove bytes from reader.data.
```go
func (reader *tcpReader) Read(p []byte) (int, error) {
	var msg api.TcpReaderDataMsg

	ok := true
	for ok && len(reader.data) == 0 {
		msg, ok = <-reader.msgQueue
		if msg != nil {
			reader.data = msg.GetBytes()
			reader.captureTime = msg.GetTimestamp()
		}

		if len(reader.data) > 0 {
			reader.packetsSeen += 1
		}
	}
	if !ok || len(reader.data) == 0 {
		return 0, io.EOF
	}

	l := copy(p, reader.data[reader.position:])
	reader.position += l
	// reader.progress.Feed(l) - not sure what to do with this

	return l, nil
}

func (reader *tcpReader) Rewind() {
	reader.position = 0
}
```

- We don’t want to keep in memory `reader.data` forever (until the stream is closed), so  we could still remove bytes if `reader.isProtocolIdentified() == true`. Need to be careful here not to mess up the `for ok && len(reader.data) == 0` loop.

mertyildiran:
I reproduced the skipping bytes issue...

mertyildiran:
@nimrod-up9 I believe I have fixed the skipping bytes issue with https://github.com/up9inc/mizu/commit/b0e95e2ea74ebdd26f85e880f0a42ebeeb28bb23

tap/tcp_reader.go
```diff
@@ -23,44 +21,44 @@ type tcpReader struct {
...
-	if err != nil {
-		_, err = io.Copy(ioutil.Discard, reader)
-		if err != nil {
-			logger.Log.Errorf("%v", err)
+	for i, extension := range extensions {
+		reader.reqResMatcher = reader.parent.reqResMatchers[i]
+		reader.counterPair = reader.parent.counterPairs[i]
+		b := bufio.NewReader(reader)
+		err := extension.Dissector.Dissect(b, reader, options)
+		if err == nil || reader.isProtocolIdentified() {
```
nimrod-up9, line 58:
```suggestion
		if reader.isProtocolIdentified() {
```
I get that if there is an error, we still want to process through all of the other extensions because maybe the error is specific to the extension. So we don't break if `err != nil`.
Does `Dissect` guarantee that if the protocol is not identified, then it will return `err == nil`? If not, then this will cause a premature exit from `run`. If yes, then it is sufficient to only check `isProtocolIdentified`.

tap/tcp_reader.go
```diff
@@ -23,44 +21,44 @@ type tcpReader struct {
 	isClient      bool
 	isOutgoing    bool
 	msgQueue      chan api.TcpReaderDataMsg // Channel of captured reassembled tcp payload
+	buffer        []byte
```
nimrod-up9, line 24:
```suggestion
	pastData        []byte
```

tap/tcp_reader.go
```diff
@@ -81,14 +79,23 @@ func (reader *tcpReader) sendMsgIfNotClosed(msg api.TcpReaderDataMsg) {
...
+	}
+
 	var msg api.TcpReaderDataMsg
 
 	ok := true
 	for ok && len(reader.data) == 0 {
 		msg, ok = <-reader.msgQueue
 		if msg != nil {
-			reader.data = msg.GetBytes()
+			reader.data = append(reader.data, msg.GetBytes()...)
```
nimrod-up9, line 98: Why append to `reader.data`? Its length is `0` here.

tap/tcp_reader.go
```diff
@@ -81,14 +79,23 @@ func (reader *tcpReader) sendMsgIfNotClosed(msg api.TcpReaderDataMsg) {
...
 
+func (reader *tcpReader) isProtocolIdentified() bool {
+	return reader.parent.protocol != nil
+}
+
 func (reader *tcpReader) Read(p []byte) (int, error) {
+	if reader.exhaustBuffer {
+		reader.data = reader.buffer
+		reader.exhaustBuffer = false
+	}
```
nimrod-up9, line 90:
```suggestion
func (reader *tcpReader) rewind() {
	reader.data = reader.buffer
}

func (reader *tcpReader) Read(p []byte) (int, error) {
```

No need for `reader.exhaustBuffer`.
Call `reader.rewind` where you would set `reader.exhaustBuffer = true`.

tap/tcp_reader.go
```diff
@@ -101,6 +108,9 @@ func (reader *tcpReader) Read(p []byte) (int, error) {
 	}
 
 	l := copy(p, reader.data)
+	if !reader.isProtocolIdentified() {
+		reader.buffer = append(reader.buffer, reader.data...)
+	}
```
nimrod-up9, line 113: This could duplicate data on `reader.buffer`.
```go
reader.data = reader.buffer
reader.buffer = append(reader.buffer, reader.data...)
```
is the same as
```
reader.buffer = append(reader.buffer, reader.buffer...)
```

Only run this when getting new messages (`msg.GetBytes()`).

tap/tcp_reader.go
```diff
@@ -81,14 +79,23 @@ func (reader *tcpReader) sendMsgIfNotClosed(msg api.TcpReaderDataMsg) {
 	reader.Unlock()
 }
 
+func (reader *tcpReader) isProtocolIdentified() bool {
+	return reader.parent.protocol != nil
+}
+
 func (reader *tcpReader) Read(p []byte) (int, error) {
+	if reader.exhaustBuffer {
+		reader.data = reader.buffer
```
nimrod-up9, line 88: I'm worried that copying a slice could have unexpected results when appending. Because both slices are views to the same underlaying array, appending to one could overwrite appends to the other.
Not sure if this is the case here or if go is smart enough to avoid this effect.

tap/tcp_reader.go
```diff
@@ -101,6 +108,9 @@ func (reader *tcpReader) Read(p []byte) (int, error) {
 	}
 
 	l := copy(p, reader.data)
+	if !reader.isProtocolIdentified() {
```
nimrod-up9, line 111: Stop hanging on to bytes when the protocol is already identified is a nice optimization.
Why not also drop `reader.buffer` when the protocol is identified?

mertyildiran: @nimrod-up9 it's actually done in here:

https://github.com/up9inc/mizu/blob/b0e95e2ea74ebdd26f85e880f0a42ebeeb28bb23/tap/tcp_stream.go#L72-L73

tap/tcp_reader.go
mertyildiran: @nimrod-up9 done with https://github.com/up9inc/mizu/commit/c176f4c7377f1b5625b6c96d02f97cbbf83a0be3

tap/tcp_reader.go
mertyildiran: @nimrod-up9 fixed with https://github.com/up9inc/mizu/commit/6eb29988b7155f75053c38bbff57344745773566

tap/tcp_reader.go
mertyildiran: @nimrod-up9 done with https://github.com/up9inc/mizu/commit/8c8eb20ad6ea7324c2fa518fc836fa0ab31f4b6d

tap/tcp_reader.go
mertyildiran: @nimrod-up9 fixed with https://github.com/up9inc/mizu/commit/ce840940b3bd1012f431233742fc0c73c3f0afa7

tap/tcp_reader.go
mertyildiran: @nimrod-up9 fixed with https://github.com/up9inc/mizu/commit/11e535cd27eaff11895cba920d6d857739d7ad20

nimrod-up9 (approved):

nimrod-up9 (dismissed):

mertyildiran:
Changed the approach from buffering the bytes to buffering the TCP packet's (`api.TcpReaderDataMsg`) since the timestamp `captureTime` should be rewinded as well.

Fixed an issue of the early timing out of unidentified TCP streams because of uninitialized `createdAt` field.

Now the acceptance tests are passed.

tap/tcp_reader.go
```diff
@@ -15,35 +15,34 @@ import (
...
-	counterPair   *api.CounterPair
-	reqResMatcher api.RequestResponseMatcher
+	ident           string
+	tcpID           *api.TcpID
+	isClosed        bool
+	isClient        bool
+	isOutgoing      bool
+	msgQueue        chan api.TcpReaderDataMsg // Channel of captured reassembled tcp payload
+	msgBuffer       []api.TcpReaderDataMsg
+	msgBufferMaster []api.TcpReaderDataMsg
```
nimrod-up9, line 25: What's the benefit of buffering the messages over buffering the bytes?

mertyildiran: > Changed the approach from buffering the bytes to buffering the TCP packet's (`api.TcpReaderDataMsg`) since the timestamp `captureTime` should be rewinded as well.

from https://github.com/up9inc/mizu/pull/1062#issuecomment-1126286372

Also it's cleaner but probably using slightly more memory.

nimrod-up9: Got it. It does make more sense.

nimrod-up9 (approved):

PR #1063, opened 2022-05-01
Bring back `debug.FreeOSMemory()` call to prevent `OOMKilled`(s)

mertyildiran:
Brings `debug.FreeOSMemory()` call that originally added with:
- https://github.com/up9inc/mizu/pull/277

then later removed with:
- https://github.com/up9inc/mizu/pull/1026

Apparently these changes are still required despite the recent optimization efforts. Otherwise the tapper pod gets easily `OOMKilled`.

**The removal of these changes didn't make into any stable releases.**

### DIFF
--- a/.github/workflows/static_code_analysis.yml
+++ b/.github/workflows/static_code_analysis.yml
@@ -10,7 +10,7 @@ permissions:
   contents: read
 
 jobs:
-  golangci:
+  go-lint:
     name: Go lint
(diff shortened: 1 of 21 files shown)


gadotroee (approved):

mertyildiran:
@gadotroee I realized that only `debug.FreeOSMemory()` is necessary to prevent `OOMKilled`(s).

PR #1064, opened 2022-05-02
Move "es Lint" to different step and add go lint on logger module

gadotroee:


IgorGov (approved):

IgorGov:
@gadotroee we should consider doing the same in mizu-ent

PR #1055, opened 2022-04-30
UI/Service-map-GUI-improvements

leon-up9:
servicemap new design

PR #1072, opened 2022-05-09
UI/servicemap-dynamic-protocol-filters

leon-up9:


ui-common/src/components/ServiceMapModal/ServiceMapModal.module.sass
```diff
@@ -1,64 +1,106 @@
 @import "../../variables.module"
```
lirazyehezkel, line 1: this code looks very similar to the CSS code you declared in #1055 
why do you need to declare it in both PRs?

lirazyehezkel (approved):

PR #1080, opened 2022-05-13
Fix the Kafka `ApiKey` query and add `ApiKeyName` field (human-readable `ApiKey`)

mertyildiran:
### Before
Screenshot from 2022-05-13 00-35-57
Screenshot from 2022-05-13 00-36-01
Screenshot from 2022-05-13 00-36-04

### After
Screenshot from 2022-05-13 21-49-59
Screenshot from 2022-05-13 21-50-04
Screenshot from 2022-05-13 21-50-08
Screenshot from 2022-05-13 21-50-11


gadotroee (commented):
@mertyildiran 
Do we need to make any changes to the unit tests files?

gadotroee:
Now I see that they failed and I guess it makes sense (which shows that the fix we did in the past to run then when the folder is changed are working ok)

mertyildiran:
> Now I see that they failed and I guess it makes sense (which shows that the fix we did in the past to run then when the folder is changed are working ok)

@gadotroee updated the dataset for Kafka unit tests with https://github.com/up9inc/mizu/commit/b9cab623668c9e45affb2d4acd454c0a1a4ce827

gadotroee (approved):

PR #1087, opened 2022-05-16
Add profiling tools

nimrod-up9:
- Run pprof server on api-server with `mizu tap --profiler` flag
- Add cpu, rss and number of cores to periodic stats
- Add script to plot performance metrics from logs
- Disable parts of the taping pipleline with env vars
  - MIZU_TAPPER_DISABLE_PCAP                          disable pcap
  - MIZU_TAPPER_DISABLE_TCP_REASSEMBLY    disable assembler
  - MIZU_TAPPER_DISABLE_TAP_TARGET              disable tcp reader creation
  - MIZU_TAPPER_DISABLE_DISSECTORS             disable sending payload to dissectors
  - MIZU_TAPPER_DISABLE_EMITTING                  disable emitting from extensions (marshaling + sending over ws)
  - MIZU_TAPPER_DISABLE_SENDING                   disable sending traffic from tapper to api-agent


agent/pkg/app/main.go
```diff
@@ -24,6 +24,8 @@ var (
 )
 
 func LoadExtensions() {
+	// (DEBUG_PERF 5) Comment out paragraphs to disable extensions
+	//   Remember to fix indices: Length of Extensions slice, index in assignment to Extensions
```
nimrod-up9, line 28: Delete

performance_analysis/plot_from_tapper_logs.py
```diff
@@ -0,0 +1,98 @@
...
+import pandas as pd
+import pathlib
+import re
+import sys
+import typing
+
+COLORMAP = plt.get_cmap('turbo')
+
+# Extract cpu and rss samples from log files and plot them
+# Input: List of log files
```
nimrod-up9, line 12: Add info

tap/passive_tapper.go
```diff
@@ -52,7 +55,8 @@ var staleTimeoutSeconds = flag.Int("staletimout", 120, "Max time in seconds to k
 var servicemesh = flag.Bool("servicemesh", false, "Record decrypted traffic if the cluster is configured with a service mesh and with mtls")
 var tls = flag.Bool("tls", false, "Enable TLS tapper")
 
-var memprofile = flag.String("memprofile", "", "Write memory profile")
+var cpuprofile = flag.String("tap.cpuprofile", "", "Write cpu profile") // cpuprofile flag clashes with one of the dependencies
```
nimrod-up9, line 58: Remove. Not flexible enough to make sense except for a specific task.

tap/source/tcp_packet_source.go
```diff
@@ -116,6 +117,10 @@ func (source *tcpPacketSource) close() {
 }
 
 func (source *tcpPacketSource) readPackets(ipdefrag bool, packets chan<- TcpPacketInfo) {
+	if os.Getenv("MIZU_TAPPER_NO_PCAP") == "true" {
+		return
+	}
+
```
mertyildiran, line 123:
```suggestion

```

tap/source/tcp_packet_source.go
```diff
@@ -116,6 +117,10 @@ func (source *tcpPacketSource) close() {
 }
 
 func (source *tcpPacketSource) readPackets(ipdefrag bool, packets chan<- TcpPacketInfo) {
+	if os.Getenv("MIZU_TAPPER_NO_PCAP") == "true" {
```
mertyildiran, line 120: I'm not sure what's the exact purpose of this disabling but it basically nullifies the capture. So, I would not initialize the handle at all if we want to disable the capturing. It disables this `libpcap` FFI call https://github.com/google/gopacket/blob/v1.1.19/pcap/pcap_unix.go#L398

tap/tcp_assembler.go
```diff
@@ -89,7 +89,9 @@ func (a *tcpAssembler) processPackets(dumpPacket bool, packets <-chan source.Tcp
 			}
 			diagnose.InternalStats.Totalsz += len(tcp.Payload)
 			a.assemblerMutex.Lock()
-			a.AssembleWithContext(packet.NetworkLayer().NetworkFlow(), tcp, &c)
+			if os.Getenv("MIZU_TAPPER_NO_ASSEMBLER") != "true" {
+				a.AssembleWithContext(packet.NetworkLayer().NetworkFlow(), tcp, &c)
+			}
```
mertyildiran, line 94:
```suggestion
			if os.Getenv("MIZU_TAPPER_DISABLE_TCP_REASSEMBLY") != "true" {
			        a.assemblerMutex.Lock()
				a.AssembleWithContext(packet.NetworkLayer().NetworkFlow(), tcp, &c)
				a.assemblerMutex.Unlock()
			}
```

tap/tcp_reader.go
```diff
@@ -48,6 +49,13 @@ func NewTcpReader(ident string, tcpId *api.TcpID, parent *tcpStream, isClient bo
 
 func (reader *tcpReader) run(options *api.TrafficFilteringOptions, wg *sync.WaitGroup) {
 	defer wg.Done()
+
+	if os.Getenv("MIZU_TAPPER_NO_DISSECTORS") == "true" {
+		b := bufio.NewReader(reader)
+		io.ReadAll(b)
+		return
+	}
```
mertyildiran, line 57: I would not call `run` at all. `io.ReadAll(b)` will consume memory.

tap/tcp_stream.go
```diff
@@ -89,6 +90,10 @@ func (t *tcpStream) GetReqResMatchers() []api.RequestResponseMatcher {
 }
 
 func (t *tcpStream) GetIsTapTarget() bool {
+	if os.Getenv("MIZU_TAPPER_NO_TAP_TARGET") == "true" {
+		return false
+	}
+
```
mertyildiran, line 96:
```suggestion
	
```

tap/tcp_stream.go
```diff
@@ -89,6 +90,10 @@ func (t *tcpStream) GetReqResMatchers() []api.RequestResponseMatcher {
 }
 
 func (t *tcpStream) GetIsTapTarget() bool {
+	if os.Getenv("MIZU_TAPPER_NO_TAP_TARGET") == "true" {
+		return false
+	}
```
mertyildiran, line 95: This will halt TCP reassembly in a bad place:

https://github.com/up9inc/mizu/blob/4ec98ff76bb7130f69dae9eb874236c342ad7de3/tap/tcp_reassembly_stream.go#L135-L147

Such that it will cause a lot of hanging dissector Goroutines.

mertyildiran (commented):
In every place you do `if os.Getenv(...`, you add a good amount of CPU instructions to a busy loop unnecessarily. They are unnecessary for the application to work really. They are there for debugging purposes. I suggest, at least do it for one time upon initialization wherever it's possible.

IMHO, this is kind of a code that would only exist temporarily on developer's machine for debugging purposes. But this PR wants to add such code into trunk. That's my comment on this.

tap/tcp_reader.go
nimrod-up9: We have to empty tcpReader's message queue.
Otherwise `tcpReassemblyStream.ReassembledSG` hands on `reader.sendMsgIfNotClosed`.

mertyildiran: `reader.close()` closes the queue. But anyway, it depends what you really want to achieve.

nimrod-up9:
> In every place you do `if os.Getenv(...`, you add a good amount of CPU instructions to a busy loop unnecessarily. They are unnecessary for the application to work really. They are there for debugging purposes. I suggest, at least do it for one time upon initialization wherever it's possible.

Good input. Fixed with https://github.com/up9inc/mizu/pull/1087/commits/6baffbb7a6b25dcfc7fbf01a188fd48b3175f4f2. Also makes things more tidy. You can easily see all of the vars in one place.

> IMHO, this is kind of a code that would only exist temporarily on developer's machine for debugging purposes. But this PR wants to add such code into trunk. That's my comment on this.

We find that time and time again we have to rewrite this kind of control in order to debug/measure performance. With this PR we do it once in order to save time in the future. Also, we are trying to have a common method to measure performance, because if each of us runs things differently then it is harder to reason about.

tap/tcp_stream.go
nimrod-up9: Shouldn't be a problem because of https://github.com/up9inc/mizu/blob/11e1286286e547813febffba90bb335387f02790/tap/tcp_stream_factory.go#L62. The routines won't be created to begin with.

tap/api/api.go
```diff
@@ -149,6 +149,10 @@ type Emitter interface {
 }
 
 func (e *Emitting) Emit(item *OutputChannelItem) {
+	if os.Getenv("MIZU_TAPPER_NO_EMITTER") == "true" {
+		return
+	}
+	
 	e.OutputChannel <- item
 	e.AppStats.IncMatchedPairs()
 }
```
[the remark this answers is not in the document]
nimrod-up9: https://github.com/up9inc/mizu/pull/1087/commits/4a0081b946f73f8434796b5a473645846f24692c

tap/passive_tapper.go
```diff
@@ -139,11 +141,27 @@ func printPeriodicStats(cleaner *Cleaner) {
...
+		
+		sysInfo, err := pidusage.GetStat(os.Getpid())
+		if err != nil {
+			sysInfo = &pidusage.SysInfo{
+				CPU:    -1,
+				Memory: -1,
+			}
+		}
+		
+		numCores, err := cpu.Counts(true)
```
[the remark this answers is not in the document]
nimrod-up9: Count cores only once with e06e178ba15828e399c344d4edd7039c98f8625e.
Count both virtual and physical with bc421693a27e4162f3a983b0220e15cf78970a56.

tap/source/tcp_packet_source.go
nimrod-up9: It's a baseline for other measurements where everything is initialized but no work is done.
This is on purpose.

tap/tcp_assembler.go
nimrod-up9: @dvdlevanon, can we skip `Lock`/`Unlock` if `MIZU_TAPPER_NO_ASSEMBLER == "true"`?

dvdlevanon (approved):

tap/tcp_assembler.go
dvdlevanon: yes

dvdlevanon (approved):

mertyildiran (approved):

PR #1088, opened 2022-05-16
Handle the wait for fetch in acceptance tests better

mertyildiran:
Related to:
- https://github.com/up9inc/mizu/pull/1074
- https://github.com/up9inc/mizu/pull/1056

### DIFF
--- a/acceptanceTests/cypress/integration/tests/UiTest.js
+++ b/acceptanceTests/cypress/integration/tests/UiTest.js
@@ -195,8 +195,6 @@ function checkFilter(filterDetails) {
     it(`checking the filter: ${filter}`, function () {
         waitForFetchAndPause();
 
-        const totalEntries = number.text();
-
         cy.get(`#list [id^=entry]`).last().then(elem => {
             const element = elem[0];
             const entryId = getEntryId(element.id);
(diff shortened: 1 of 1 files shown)


acceptanceTests/cypress/integration/tests/UiTest.js
```diff
@@ -193,58 +193,60 @@ function checkFilter(filterDetails) {
...
+}
 
-            // reloading then waiting for the entries number to load
-            resizeToNormalMizu();
-            cy.reload();
-            cy.get('#total-entries', {timeout: refreshWaitTimeout}).should('have.text', totalEntries);
-        })
+function waitForFetch() {
+    cy.get('#entries-length', {timeout: refreshWaitTimeout}).should((el) => {
+        expect(parseInt(el.text().trim(), 10)).to.be.greaterThan(20);
```
gadotroee, line 245: where is this "20" comes from?
I guess it should be some consts or parameter we are passing to the function but I know that we are sending 50 messages so I'm not sure why we are waiting only for "more than 20.."

mertyildiran: There are multiple tests using this function and I'm not sure what tests expects how many records. 20 was my guesstimation.

If all tests are 50+, we can say it's 50. Since WebSocket fetches up to 50 records.

I'm not sure what parameter and how exactly we should pass to here for the expected `greaterThan` value.

gadotroee: I'm almost sure it is always 50 messages.
but I'm not sure we are sure it always catches all entries so I wouldn't go to 50.
lets start with a working tests and change it if we see that this can be changed.

PR #1096, opened 2022-05-19
Reduce the sleep in AMQP tests to 5 seconds

mertyildiran:
We drop TCP streams that are unidentified for 10 seconds. The `time.Sleep(10 * time.Second)` was causing sporadic failures in the tests.

For some reason, the TCP reassembly does not work without that a timeout. So, I reduced it to 5 seconds.


acceptanceTests/extensions_test.go
```diff
@@ -167,7 +168,7 @@ func TestAmqp(t *testing.T) {
 	defer conn.Close()
 
 	// Temporary fix for missing amqp entries
-	time.Sleep(10 * time.Second)
+	time.Sleep(5 * time.Second)
```
gadotroee, line 171: from what I know the value 10 was not chosen with no reason, 
I remember that this test was started with no timeout and as you. saw it didn't work,
then adding the timeout helped but with values lower than 10 it sporadically failed too.

```We drop TCP streams that are unidentified for 10 seconds``` - can you point for the place that cleans after 10 seconds? (I want to see if this was changed in the past).


I see you changed in the Redis but not in AMQP so I'm confused.

mertyildiran: @gadotroee in here:

https://github.com/up9inc/mizu/blob/e6a2bef720c5b51cbed98ddf3fb07453b26968c4/tap/tcp_streams_map.go#L63-L68

and that code is pretty old actually.

The TCP assembler does not call `New()` immediately that's what I understand but I'm not sure why;

https://github.com/up9inc/mizu/blob/e6a2bef720c5b51cbed98ddf3fb07453b26968c4/tap/tcp_stream_factory.go#L49

mertyildiran: Actually `New()` is being called immediately. It's something else.

PR #1097, opened 2022-05-19
Call `SetProtocol` in AMQP faster and remove `GetProtocol` method

mertyildiran:
Reduces TCP stream drop probability of AMQP (improves acceptance tests stability) by calling `SetProtocol` faster.

Removes the redundant `GetProtocol()` method.


gadotroee (commented):
@mertyildiran 
This seems to be good (setting the protocol first thing after we are getting "valid" amqp information. 

only questions is to be sure that I understand, the 
```
		if reader.GetParent().GetProtocol() != nil && reader.GetParent().GetProtocol() != &protocol {
			return errors.New("Identified by another protocol")
		}
```

that removed from protocols can be done because we are not starting them in the same time, we will start each protocol dissector after we know that it failed to be processed by another protocol dissector, right?

mertyildiran:
> @mertyildiran This seems to be good (setting the protocol first thing after we are getting "valid" amqp information.
> 
> only questions is to be sure that I understand, the
> 
> ```
> 		if reader.GetParent().GetProtocol() != nil && reader.GetParent().GetProtocol() != &protocol {
> 			return errors.New("Identified by another protocol")
> 		}
> ```
> 
> that removed from protocols can be done because we are not starting them in the same time, we will start each protocol dissector after we know that it failed to be processed by another protocol dissector, right?

Yes, those checks are forgotten to be removed in https://github.com/up9inc/mizu/pull/1062 There is no chance for a dissector to be called while also the protocol is identified because of the check below:

https://github.com/up9inc/mizu/blob/41dc081f3bd8bbb91109b0c1b483254013dcf96f/tap/tcp_reader.go#L63-L66

gadotroee (approved):

PR #1100, opened 2022-05-19
Remove the using of leftOff from websocket

lirazyehezkel:


ui-common/src/components/TrafficViewer/EntriesList.tsx
```diff
@@ -156,7 +159,6 @@ export const EntriesList: React.FC<EntriesListProps> = ({
         case "queryMetadata":
           setTruncatedTimestamp(message.data.truncatedTimestamp);
           setQueriedTotal(message.data.total);
-          setLeftOffTop(leftOffState => leftOffState === "" ? message.data.leftOff : leftOffState);
```
mertyildiran, line 159: Left off value should come from the metadata stream. Why do we remove this `setLeftOffTop` call?

PR #1104, opened 2022-05-22
Change redact to opt in

nimrod-up9:


gadotroee (approved):

gadotroee (changes requested):
we need to check that acceptance tests pass

gadotroee (approved):

PR #1105, opened 2022-05-23
close finished live streams

dvdlevanon:


tap/tcp_streams_map.go
```diff
@@ -67,6 +66,9 @@ func (streamMap *tcpStreamMap) CloseTimedoutTcpStreamChannels() {
 						diagnose.AppStats.DroppedTcpStreams, runtime.NumGoroutine(), tcpStreamChannelTimeoutMs/time.Millisecond)
 				}
 			}
+			if stream.client.doneReading && stream.server.doneReading {
```
nimrod-up9, line 69: Function name `CloseTimedoutTcpStreamChannels` doesn't suite this function anymore because it also closes done streams, not only timed-out streams.
Rename. I suggest `RunTcpStreamChannelsCleanup`.

tap/tcp_reader.go
```diff
@@ -66,6 +68,8 @@ func (reader *tcpReader) run(options *api.TrafficFilteringOptions, wg *sync.Wait
 		}
 		reader.rewind()
 	}
+
+	reader.doneReading = true
```
nimrod-up9, line 72: Why not close the channel here instead of in `CloseTimedoutTcpStreamChannels`?

PR #1124, opened 2022-06-02
Fix cypress files to cypress version 10 (migration)

gadotroee:


acceptanceTests/cypress/e2e/testHelpers/TrafficHelper.js
```diff
@@ -4,8 +4,6 @@ export const valueTabs = {
     none: null
 }
 
-const maxEntriesInDom = 13;
```
RoyUP9, line 7: was this not used?

gadotroee: correct.

acceptanceTests/Makefile
```diff
@@ -1,2 +1,3 @@
 test: ## Run acceptance tests.
+	@npm install cypress@10.0.1 -y
```
RoyUP9, line 2: can this be done in the setup file?

gadotroee: no. it should be installed when we are in the directory of the acceptance test.
so it feels like this place is the best place to put it in

gadotroee: (if it is installed before (like I tried) it fails with error of "module not found" when we are running the "npx cypress run.."

RoyUP9 (approved):

PR #1127, opened 2022-06-02
Ui/TRA-4586_add-insertion-filter-to-settings-page

leon-up9:


ui-common/src/components/TrafficViewer/Filters.tsx
```diff
@@ -47,18 +54,59 @@ export const modalStyle = {
...
 
-export const QueryForm: React.FC<QueryFormProps> = ({backgroundColor, reopenConnection}) => {
+export const CodeEditorWrap: FC<QueryFormProps> = ({ query, onQueryChange, validateQuery, onValidationChanged }) => {
+    const [queryBackgroundColor, setQueryBackgroundColor] = useState("#f5f5f5");
+    const handleQueryChange = useMemo(
+        () =>
+            debounce(async (query: string) => {
+                if (!query) {
+                    setQueryBackgroundColor("#f5f5f5");
+                    onValidationChanged && onValidationChanged({ query: query, message: "", valid: true })
```
AmitUp9, line 64: why onCalidationChanged Twice?

leon-up9: you mean called twice?
what would you suggest?

AmitUp9: I missed the "?" in interface. I got it now. Is there a case where you don't want to use "onValidationChanged" in CodeEditorWrap?

ui-common/src/components/TrafficViewer/Filters.tsx
```diff
@@ -1,36 +1,43 @@
...
         <QueryForm
-            backgroundColor={backgroundColor}
+            query={query}
             reopenConnection={reopenConnection}
-        />
+            onQueryChange={(query) => { setQuery(query?.trim()); }} validateQuery={api?.validateQuery} />
     </div>;
 };
 
+type OnQueryChnage = { valid: boolean, message: string, query: string }
```
lirazyehezkel, line 33: OnQueryChange

ui-common/src/components/TrafficViewer/Filters.tsx
```diff
@@ -47,18 +54,59 @@ export const modalStyle = {
...
 
     const [openModal, setOpenModal] = useState(false);
 
     const handleOpenModal = () => setOpenModal(true);
     const handleCloseModal = () => setOpenModal(false);
 
-    const handleChange = async (e) => {
-        setQuery(e.target.value.trim());
+    const handleChange = async (queryObj) => {
+        onQueryChange(queryObj);
```
lirazyehezkel, line 109: I think this is a redundant func, you can call onQueryChange directly

PR #1131, opened 2022-06-07
TRA-4276 Folder structure refactor

lirazyehezkel:


ui-common/src/components/EntryDetailed/EntryDetailed.tsx
```diff
@@ -1,18 +1,18 @@
...
 import { RecoilState, useRecoilValue } from "recoil";
 import focusedEntryIdAtom from "../../recoil/focusedEntryId";
-import TrafficViewerApi from "./TrafficViewerApi";
+import TrafficViewerApi from "../TrafficViewer/TrafficViewerApi";
 import TrafficViewerApiAtom from "../../recoil/TrafficViewerApi/atom";
 import queryAtom from "../../recoil/query/atom";
 import useWindowDimensions, { useRequestTextByWidth } from "../../hooks/WindowDimensionsHook";
 import { TOAST_CONTAINER_ID } from "../../configs/Consts";
-import spinner from "assets/spinner.svg";
+import spinner from "../assets/spinner.svg";
```
leon-up9, line 15: are you sure its working?
if so what changed?

leon-up9:
how about adding an index.ts to folders of common components and 
changing import from  "import Queryable from "../Queryable/Queryable";
to "../Queryable";
for example

ui-common/src/components/EntryDetailed/EntryDetailed.tsx
lirazyehezkel: yes, the files were moved and this is the new hierarchy between the img and the component

lirazyehezkel:
> 

fixed the comments about the tgz file and the spinner.
And I created a follow-up ticket regarding your suggestion with the index files - https://up9.atlassian.net/browse/TRA-4605
I think it's a good idea but I would do it in a different PR, we have already too many changes on this one

AmitUp9 (approved):

PR #1151, opened 2022-06-15
Add ARM64 support to eBPF

mertyildiran:
This PR adds ARM64 to eBPF for the build strategies below:
- `docker build . -t mizutest:arm64v8-cross --build-arg BUILDARCH=amd64 --build-arg TARGETARCH=arm64v8`
- `docker build . -t mizutest:arm64v8-native --build-arg BUILDARCH=arm64v8 --build-arg TARGETARCH=arm64v8`

It builds and loads the BPF without any issues on both x86_64 and ARM64. I couldn't verify the capture's itself on an Amazon EC2 A1 instance because of the issue below:

`bpf_get_current_pid_tgid() >> 32` returns an incorrect PID on ARM64 `minikube` Docker driver, the problem does not exist on x86_64 `minikube` VirtualBox driver. It's either specific to `minikube` ARM64 or `minikube` Docker driver. It seems like a bug of either `minikube` or `libbpf`.

~~So, it needs to be tested on a cluster with ARM64 nodes using a cloud provider.~~ See https://github.com/up9inc/mizu/pull/1151#issuecomment-1159726744


dvdlevanon (approved):

mertyildiran:
I have observed entries sourced from OpenSSL on ARM64 with [minikube none driver](https://minikube.sigs.k8s.io/docs/drivers/none/). But didn't see any Go TLS. Therefore Go implementation needs further improvement after this PR. Probably related to [these lines](https://github.com/up9inc/mizu/blob/feat/ebpf-arm64/tap/tlstapper/bpf/include/go_abi_internal.h#L92-L123).

Apparently, the incorrect PID issue of `bpf_get_current_pid_tgid() >> 32` is specific to [minikube docker driver](https://minikube.sigs.k8s.io/docs/drivers/docker/) on ARM64.

dvdlevanon (approved):

PR #1153, opened 2022-06-16
TRA-4602 Traffic statistics - Pie Chart

lirazyehezkel:
image

image


gadotroee:
waiting for https://github.com/up9inc/mizu/pull/1144

ui-common/src/components/modals/TrafficStatsModal/TrafficStatsModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+interface TrafficStatsModalProps {
+  isOpen: boolean;
+  onClose: () => void;
+  getTrafficStatsDataApi: () => Promise<any>
+}
+
+export const TrafficStatsModal: React.FC<TrafficStatsModalProps> = ({ isOpen, onClose, getTrafficStatsDataApi }) => {
+
+  const modes = Object.keys(StatsMode).filter(x => !(parseInt(x) >= 0));
```
AmitUp9, line 35: Why do you need to filter StatsMode like that?

lirazyehezkel: Object.keys of an enum return the string keys but also return numerical keys.
for example if we have the enum {a: a, b: b}
Object.keys(enum) will return [a,b,0,1] and we want to filter those numbers

https://www.crojach.com/blog/2019/2/6/getting-enum-keys-in-typescript

AmitUp9: ok got it. thanks

AmitUp9 (approved):

PR #1120, opened 2022-05-30
Add Go `crypto/tls` eBPF tracer for TLS connections

mertyildiran:
# Go `crypto/tls` eBPF Tracer for Tracing TLS Connections

This is an eBPF-based tracer that's programmed to trace binaries written
in Go which depends on `crypto/tls` standard library to encrypt their TLS payloads.

The program attaches `uprobe`(s) to the offsets below:
 - [`crypto/tls.(*Conn).Write`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1099)
 - [`crypto/tls.(*Conn).Write+972`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1104)
 - [`crypto/tls.(*Conn).Write+1385`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1113)
 - [`crypto/tls.(*Conn).Write+1497`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1120)
 - [`crypto/tls.(*Conn).Write+1609`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1124)
 - [`crypto/tls.(*Conn).Write+1707`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1128)
 - [`crypto/tls.(*Conn).Write+1787`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1145)
 - [`crypto/tls.(*Conn).Write+1847`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1152)
 - [`crypto/tls.(*Conn).Read`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1263)
 - [`crypto/tls.(*Conn).Read+260`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1265)
 - [`crypto/tls.(*Conn).Read+305`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1270)
 - [`crypto/tls.(*Conn).Read+434`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1278)
 - [`crypto/tls.(*Conn).Read+765`](https://github.com/golang/go/blob/go1.17.6/src/crypto/tls/conn.go#L1282)
[...]

PR #1155, opened 2022-06-19
Fix a `slice bounds out of range` panic in Go offsets calculation

mertyildiran:
Introduced by https://github.com/up9inc/mizu/pull/1120

### DIFF
--- a/tap/tlstapper/go_offsets.go
+++ b/tap/tlstapper/go_offsets.go
@@ -10,6 +10,7 @@ import (
 	"github.com/Masterminds/semver"
 	"github.com/cilium/ebpf/link"
 	"github.com/knightsc/gapstone"
+	"github.com/up9inc/mizu/logger"
 )
 
 type goOffsets struct {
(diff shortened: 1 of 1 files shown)


tap/tlstapper/go_offsets.go
```diff
@@ -145,6 +146,16 @@ func getOffsets(filePath string) (offsets map[string]*goExtendedOffset, err erro
 		symEndingIndex := symStartingIndex + sym.Size
 
 		// collect the bytes of the symbol
+		textSectionDataLen := uint64(len(textSectionData) - 1)
+		if symEndingIndex > textSectionDataLen {
+			logger.Log.Errorf(
```
IgorGov, line 151: i can see that we have now many crashes, does it mean that it will be printed as error now? if yes, I am wondering if it have to be error?

mertyildiran: We can print as `Infof` instead? It's caused by the lack of disassembly info in the binaries.

mertyildiran: or `Warningf` maybe?

mertyildiran: Changed it to `Warningf` with https://github.com/up9inc/mizu/commit/f56e1b135ba0109bf9cdfeb14d76fdd033d51af5

IgorGov (approved):

PR #1157, opened 2022-06-21
Add timing endpoint (timeline data) for stats

gadotroee:


agent/pkg/providers/stats_provider.go
```diff
@@ -45,9 +45,15 @@ type AccumulativeStatsProtocol struct {
...
+	ProtocolsData []*AccumulativeStatsProtocol `json:"protocols"`
+	Time          int64                        `json:"timestamp"`
+}
+
 var (
-	generalStats = GeneralStats{}
-	bucketsStats = BucketStats{}
+	generalStats            = GeneralStats{}
+	bucketsStats            = BucketStats{}
+	internalBucketThreshold = time.Minute * 1
```
IgorGov, line 56: can it be const?

agent/pkg/providers/stats_provider.go
```diff
@@ -113,6 +119,97 @@ func GetAccumulativeStats() []*AccumulativeStatsProtocol {
 	return finalResult
 }
 
+func GetAccumulativeStatsTiming(intervalSeconds int, numberOfBars int) []*AccumulativeStatsProtocolTime {
+	bucketStatsCopy := BucketStats{}
+	if err := copier.Copy(&bucketStatsCopy, bucketsStats); err != nil {
```
IgorGov, line 124: should this section be locked? to make sure no changed to the bucketsStats during the copy

gadotroee: I have changed those lines to be in function with https://github.com/up9inc/mizu/pull/1157/commits/5909395b053adc932ee67e2c4547b98bc77962a8 so both places that use the object will get it from this function

agent/pkg/providers/stats_provider.go
gadotroee: Sure, changed in this https://github.com/up9inc/mizu/pull/1157/commits/5909395b053adc932ee67e2c4547b98bc77962a8

agent/pkg/providers/stats_provider.go
```diff
@@ -113,6 +119,97 @@ func GetAccumulativeStats() []*AccumulativeStatsProtocol {
 	return finalResult
 }
 
+func GetAccumulativeStatsTiming(intervalSeconds int, numberOfBars int) []*AccumulativeStatsProtocolTime {
+	bucketStatsCopy := BucketStats{}
```
IgorGov: 1. private?
2. add 'copy' to its name?

IgorGov (approved):

PR #1158, opened 2022-06-21
Fix problems that golang inspection reported  

gadotroee:


agent/pkg/api/contract_validation.go
```diff
@@ -24,13 +24,13 @@ const (
 
 func loadOAS(ctx context.Context) (doc *openapi3.T, contractContent string, router routers.Router, err error) {
 	path := fmt.Sprintf("%s%s", shared.ConfigDirPath, shared.ContractFileName)
-	bytes, err := ioutil.ReadFile(path)
+	bytesValue, err := ioutil.ReadFile(path)
```
gadotroee, line 27: bytes is imported module so there is name collision

agent/pkg/controllers/entries_controller.go
```diff
@@ -44,10 +44,10 @@ func GetEntries(c *gin.Context) {
 	}
 
 	entriesProvider := dependency.GetInstance(dependency.EntriesProvider).(entries.EntriesProvider)
-	entries, metadata, err := entriesProvider.GetEntries(entriesRequest)
+	entryWrappers, metadata, err := entriesProvider.GetEntries(entriesRequest)
```
gadotroee, line 47: entries is imported module so there is name collision

agent/pkg/dependency/container.go
```diff
@@ -1,11 +1,11 @@
 package dependency
 
-var typeIntializerMap = make(map[DependencyContainerType]func() interface{}, 0)
+var typeInitializerMap = make(map[ContainerType]func() interface{}, 0)
```
gadotroee, line 3: typo

agent/pkg/dependency/type_names.go
```diff
@@ -1,12 +1,12 @@
 package dependency
 
-type DependencyContainerType string
+type ContainerType string
```
gadotroee, line 3: variable starts with package name

agent/pkg/providers/tappedPods/tapped_pods_provider.go
```diff
@@ -44,9 +44,10 @@ func Set(tappedPodsToSet []*shared.PodInfo) {
 
 func GetTappedPodsStatus() []shared.TappedPodStatus {
 	tappedPodsStatus := make([]shared.TappedPodStatus, 0)
+	tapperStatus := tappers.GetStatus()
```
gadotroee, line 47: calling the function only once (instead of every loop)

cli/telemetry/telemetry.go
```diff
@@ -73,11 +73,7 @@ func shouldRunTelemetry() bool {
 		return false
 	}
 
-	if mizu.Branch != "main" && mizu.Branch != "develop" {
-		return false
-	}
-
-	return true
+	return mizu.Branch == "main" || mizu.Branch == "develop"
```
gadotroee, line 76: simplified condition

tap/extensions/amqp/main.go
```diff
@@ -74,13 +74,13 @@ func (d dissecting) Dissect(b *bufio.Reader, reader api.TcpReader, options *api.
 	var lastMethodFrameMessage Message
 
 	for {
-		frame, err := r.ReadFrame()
+		frameVal, err := r.readFrame()
```
gadotroee, line 77: imported module with name frame exists (name collision)

acceptanceTests/setup.sh
```diff
@@ -57,10 +57,9 @@ kubectl expose deployment redis --type=LoadBalancer --port=6379 -n mizu-tests --
 echo "Creating rabbitmq service"
 kubectl expose deployment rabbitmq --type=LoadBalancer --port=5672 -n mizu-tests --dry-run=client -o yaml | kubectl apply -f -
 
-# TODO: need to understand how to fail if address already in use
 echo "Starting proxy"
 rm -f ${PROXY_LOG}
-kubectl proxy --port=8080 > ${PROXY_LOG} &
+kubectl proxy --port=8080 >  ${PROXY_LOG} &
```
RoyUP9, line 62: is this addition intendant?

gadotroee: no, it is mistake, will remove it

tap/extensions/amqp/read.go
```diff
@@ -299,7 +299,7 @@ func readArray(r io.Reader) ([]interface{}, error) {
 
 	var (
 		lim   = &io.LimitedReader{R: r, N: int64(size)}
-		arr   = []interface{}{}
+		arr   []interface{}
```
RoyUP9, line 302: not sure this is the best way, and it could break something (there is a difference between empty array and nil)

tap/extensions/http/http2_assembler.go
```diff
@@ -89,7 +89,7 @@ type Http2Assembler struct {
 
 func (ga *Http2Assembler) readMessage() (streamID uint32, messageHTTP1 interface{}, isGrpc bool, err error) {
 	// Exactly one Framer is used for each half connection.
-	// (Instead of creating a new Framer for each ReadFrame operation)
+	// (Instead of creating a new Framer for each readFrame operation)
```
RoyUP9, line 92: i don't think this name needs to change

tap/tlstapper/tls_tapper.go
```diff
@@ -123,31 +123,31 @@ func (t *TlsTapper) ClearPids() {
 }
 
 func (t *TlsTapper) Close() []error {
-	errors := make([]error, 0)
+	errorsResult := make([]error, 0)
```
RoyUP9, line 126: rename to returnValue (same as all the other close funcs)

tap/tlstapper/tls_tapper.go
```diff
@@ -160,7 +160,8 @@ func setupRLimit() error {
 	return nil
 }
 
-func (t *TlsTapper) tapSsllibPid(pid uint32, sslLibrary string, namespace string) error {
+func (t *TlsTapper) tapSSLLibPid(pid uint32, sslLibrary string, namespace string) error {
+
```
RoyUP9, line 164: remove space

tap/extensions/amqp/read.go
gadotroee: append is working when nil, so it is ok to create it as nil..

tap/extensions/amqp/main.go
mertyildiran: @gadotroee what's module that it collides with?

gadotroee: I don't see any module that has this name so maybe it was mistake in the comment, I will check again now

gadotroee: I think it is this one - https://github.com/up9inc/mizu/blob/develop/tap/extensions/amqp/types.go#L255 (here it is not a module but a struct with this name (in the same package)

mertyildiran (approved):

PR #1162, opened 2022-06-21
Fix Go `crypto/tls` eBPF tracer on ARM64

mertyildiran:
Follow up on:
- https://github.com/up9inc/mizu/pull/1120
- https://github.com/up9inc/mizu/pull/1151#issuecomment-1160221146


mertyildiran:
Seen Go `crypto/tls` sourced entries on ARM64 and verified that the build strategies which are targeting ARM64 works correctly.

mertyildiran:
Updated the object files.

dvdlevanon (approved):

PR #1163, opened 2022-06-23
Remove OAS contract monitoring feature

RoyUP9:
Removes the feature added with https://github.com/up9inc/mizu/pull/325

Part of https://up9.atlassian.net/browse/TRA-4610

PR #1166, opened 2022-06-25
Remove Summary field from output channel item

gadotroee:
This fields is never populated so it is more confusing that it is here (follow up from https://github.com/up9inc/mizu/pull/1144#discussion_r901420701)


gadotroee:
test data files should be updated so I will wait with this after https://github.com/up9inc/mizu/pull/1163

PR #1170, opened 2022-06-27
Move general functions from timeline stats to helpers utils

AmitFainholts:
padTo2Digits, getHoursAndMinutes and creatUniqueObjArrayByProp now in helpers for general use.


ui-common/src/helpers/Utils.ts
```diff
@@ -26,4 +26,20 @@ export class Utils {
 
     return bytes.toFixed(dp) + ' ' + units[u];
   }
+
+static padTo2Digits = (num) => {
```
leon-up9, line 30: only i see weird indentation?

AmitUp9: i am not sure that I see it

leon-up9: can we fix it?

AmitUp9: I did format document, I hope it fix what you meant

leon-up9 (approved):

PR #1174, opened 2022-06-30
Merge traffic stats endpoints to one and add auto interval logic

gadotroee:


agent/pkg/providers/stats_provider.go
```diff
@@ -201,13 +251,24 @@ func convertAccumulativeStatsDictToArray(methodsPerProtocolAggregated map[string
...
-			Color:   protocolToColor[protocolName],
+			Color:   getColor(protocolName),
 			Methods: methods,
 		})
 	}
 	return protocolsData
 }
 
+func getColor(protocolName string) string {
+	if protocolToColor == nil {
```
IgorGov, line 262: can we do it in 'SetProtocolMap'? - it seems like 'SetProtocolMap' will be called once on init, and here we need locks for it to work properly

lirazyehezkel (approved):

PR #1177, opened 2022-07-03
Improve tls info with kprobes

nimrod-up9:
Instead of going through the socket fd, addresses are obtained in kprobe/tcp_sendmsg on ssl write and kprobe/tcp_recvmsg on ssl read. The tcp kprobes and the openssl uprobes communicate through the id->sslInfo bpf map.


tap/tlstapper/bpf/common.c
```diff
@@ -104,6 +104,8 @@ static __always_inline void output_ssl_chunk(struct pt_regs *ctx, struct ssl_inf
     chunk->len = count_bytes;
     chunk->fd = info->fd;
 
+    (void)memcpy(&chunk->kprobe_address_pair, &info->kprobe_address_pair, sizeof(chunk->kprobe_address_pair));
```
dvdlevanon, line 107: I wonder why not to use the `bpf_probe_read` which is the both the standard method in eBPF and used in our code base.

Indeed loops in eBPF are limited, new Kernels allows bounded loops (loop from X to known and small Y). While older kernels don't allow it and force the use of Unrolling loops, which is basically a copy paste of the loop body.
I assume the compiler's built it `__builtin_memcpy` do small bounded loop so new kernels allow it, but its better to use the `bpf_probe_read` which does the same and supported by all kernel versions.

tap/tlstapper/bpf/include/maps.h
```diff
@@ -32,7 +41,8 @@ struct tls_chunk {
     __u32 recorded;
     __u32 fd;
     __u32 flags;
-    __u8 address[16];
+    __u8 fd_address[16];
```
dvdlevanon, line 44: I think we can get rid of `fd_address`

tap/tlstapper/chunk.go
```diff
@@ -12,8 +12,8 @@ import (
 const FLAGS_IS_CLIENT_BIT uint32 = (1 << 0)
 const FLAGS_IS_READ_BIT uint32 = (1 << 1)
 
-func (c *tlsTapperTlsChunk) getAddress() (net.IP, uint16, error) {
-	address := bytes.NewReader(c.Address[:])
+func (c *tlsTapperTlsChunk) getFdAddress() (net.IP, uint16, error) {
```
dvdlevanon, line 15: Let's get rid of this

tap/tlstapper/tls_poller.go
```diff
@@ -161,6 +156,28 @@ func (p *tlsPoller) handleTlsChunk(chunk *tlsTapperTlsChunk, extension *api.Exte
 	return nil
 }
 
+func (p *tlsPoller) getAddressPair(chunk *tlsTapperTlsChunk) (addressPair, error) {
```
dvdlevanon, line 159: This function should use `getKprobeAddressPair`, no fallbacks, and remove the old sockFd and addressFd

tap/tlstapper/bpf/common.c
nimrod-up9: Fixed in a4948dbd9f8d44355b997ac7e2efcb5eae4e5dd0.

I thought that `bpf_probe_read` is meant for reading from kernel space, and therefore it would be confusing to see it used otherwise. (for example, here it might imply to the reader that `chunk->kprobe_address_pair` is in kernel space. It is just a matter of convention. If it is used in our code as a replacement for memcpy then should be used here as well.

tap/tlstapper/bpf/include/maps.h
nimrod-up9: Fixed 1819f9b134e35d4bf7b160e608beb6383f165632.

tap/tlstapper/bpf/include/common.h
```diff
@@ -7,6 +7,20 @@ Copyright (C) UP9 Inc.
 #ifndef __COMMON__
 #define __COMMON__
 
+#define AF_INET	2	/* Internet IP Protocol */
+
+#ifndef memset
```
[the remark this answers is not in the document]
nimrod-up9: Removed in a4948dbd9f8d44355b997ac7e2efcb5eae4e5dd0.

tap/tlstapper/chunk.go
nimrod-up9: Removed cf6598b8e5f6eecb568c7566122cba75cdf8d3db.

tap/tlstapper/tls_poller.go
nimrod-up9: Somewhat fixed cf6598b8e5f6eecb568c7566122cba75cdf8d3db.
Didn't remove but improved the code. I think it is more straightforward now.
Couldn't remove because the Go TLS tapper still fetches the addresses from sock fd.

dvdlevanon (approved):

PR #1184, opened 2022-07-05
Add available protocols to the stats endpoint (colors for methods are coming from server)

gadotroee:
This PR adds the following: 
1. Show only protocols with traffic in the filtering options 
2. Methods colors are returned from server 
3. Better view when there are low number of bars


ui-common/src/components/modals/TrafficStatsModal/TimelineBarChart/TimelineBarChart.tsx
```diff
@@ -42,58 +42,61 @@ export const TimelineBarChart: React.FC<TimelineBarChartProps> = ({ timeLineBarC
...
         const commandsNames = [];
         const protocolsCommands = [];
         data.sort((a, b) => a.timestamp < b.timestamp ? -1 : 1).forEach(protocolObj => {
-            let newCommandlbj: { [k: string]: any } = {};
-            newCommandlbj.timestamp = Utils.getHoursAndMinutes(protocolObj.timestamp);
-            protocolObj.protocols.find(protocol => protocol.name === selectedProtocol)?.methods.forEach(command => {
-                newCommandlbj[`${command.name}`] = command[StatsMode[timeLineBarChartMode]]
-                if (commandsNames.indexOf(command.name) === -1)
-                    commandsNames.push(command.name);
+            let newMethodobj: { [k: string]: any } = {};
```
lirazyehezkel, line 52: newMethodObj

ui-common/src/components/modals/TrafficStatsModal/TimelineBarChart/TimelineBarChart.tsx
```diff
@@ -42,58 +42,61 @@ export const TimelineBarChart: React.FC<TimelineBarChartProps> = ({ timeLineBarC
 
     useEffect(() => {
         if (selectedProtocol === ALL_PROTOCOLS) {
-            setCommandStats(null);
-            setcommandNames(null);
+            setMethodsStats(null);
+            setMethodsNamesAndColors(null);
             return;
         }
         const commandsNames = [];
```
lirazyehezkel, line 49: if you rename command with method, I think it would be better to change in all the places

ui-common/src/components/modals/TrafficStatsModal/TrafficPieChart/TrafficPieChart.tsx
```diff
@@ -60,10 +60,11 @@ export const TrafficPieChart: React.FC<TrafficPieChartProps> = ({ pieChartMode,
       setCommandStats(null);
       return;
     }
-    const commandsPieData = data.find(protocol => protocol.name === selectedProtocol)?.methods.map(command => {
+    const commandsPieData = data.find(protocol => protocol.name === selectedProtocol)?.methods.map(method => {
```
lirazyehezkel, line 63: command to method?

ui-common/src/helpers/Utils.ts
```diff
@@ -51,17 +51,4 @@ export class Utils {
     return true;
   }
 
-  static stringToColor = (str) => {
```
lirazyehezkel, line 54: we get the method color from BE?

lirazyehezkel (commented):
some comments

ui-common/src/helpers/Utils.ts
AmitUp9: we get the colors from BE

lirazyehezkel (approved):

PR #1185, opened 2022-07-05
move icon to right side

leon-up9:


ui-common/src/components/EntryDetailed/EntryViewer/AutoRepresentation.tsx
```diff
@@ -27,7 +27,7 @@ export const AutoRepresentation: React.FC<any> = ({ representation, color, opene
         const arr = [
             {
                 tab: 'Request',
-                badge: isReplayDisplayed() && <span title="Replay Request"><ReplayIcon fill={color} stroke={color} style={{ marginLeft: "10px", cursor: "pointer", height: "22px" }} onClick={() => setIsOpenRequestModal(true)} /></span>
+                badge: null
```
AmitUp9, line 30: why do we need badge here?

leon-up9: the question is why used badge: null?
it was used in other tabs

AmitUp9 (approved):

PR #1190, opened 2022-07-07
Delete ebpf object files

nimrod-up9:
We no longer need to store files in git history.
We used to need it to simplify development. Most developers don't edit
the ebpf programs and we wanted to avoid having everyone having to set
up a build environment.
Now we have the mizu-ebpf-builder docker which is an easy way to compile
ebpf.
When building the agent image, we used to take the local object files
from the working dir. Now the objects are compiled when building the
docker image.


mertyildiran:
We might want to `.gitignore` the `*.o` or `tap/tlstapper/*.o` for local builds.

nimrod-up9:
> > When running the linting tap module there are errors if there are no `*.o` files. Fixed by adding the ebpf go generated files to the list of files ignored by the linter. The files are analyzed by their errors are not reported.
> > There is an error when trying to lint agent when there are no `*.o` files:
> > ```
> > WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package tlstapper: could not load export data: no export data for "github.com/up9inc/mizu/tap/tlstapper"
> > ERRO Running error: 1 error occurred:
> >         * can't run linter goanalysis_metalinter: buildir: failed to load package tlstapper: could not load export data: no export data for "github.com/up9inc/mizu/tap/tlstapper"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Note that if you build the object files (`tap/tlstapper/bpf-builder/build.sh`) then run the linter (`cd agent && golangci-lint run`), then delete the object files and run the linter again, it will pass because of its cache.
> > I'm not sure what this error means or how to fix. Found some hints [here](https://youtrack.jetbrains.com/issue/GO-11249/golangci-lint-file-watcher-fails-to-run-failed-to-load-package-unsafeheader#focus=Comments-27-5254493.0-0) but none of them worked.
> > Made a workaround by generating the eBPF objects and go bindings before linting. It adds about 30 seconds to the "Static code analysis" GitHub action.
> > Any thoughts @mertyildiran @dvdlevanon ?
> 
> @nimrod-up9 `make build` does not pass on your machine. First `make build` should pass.

@mertyildiran I think that the workaround of building the object files before linting is not actually a workaround, it is the correct solution. The linter can't be expected to succeeded if the build fails, and the object files are a prerequisite of the build. Therefore I'm keeping it instead of trying to make the linter ignore the missing o files. I added `make bpf` make target that is a prerequisite of `make agent`, so developers don't have to remember to build the bpf before compiling the go part.

mertyildiran (approved):

PR #1191, opened 2022-07-07
TRA_4625- Add dates to timeline ticks

AmitUp9:
Screen Shot 2022-07-07 at 14 01 32

### DIFF
--- a/ui-common/src/helpers/Utils.ts
+++ b/ui-common/src/helpers/Utils.ts
@@ -37,6 +37,20 @@ export class Utils {
     return hoursAndMinutes;
   }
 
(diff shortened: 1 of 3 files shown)


ui-common/src/helpers/Utils.ts
```diff
@@ -37,6 +37,20 @@ export class Utils {
     return hoursAndMinutes;
   }
 
+  static formatDate = (date) => {
+    let d = new Date(date),
```
lirazyehezkel, line 41: const

lirazyehezkel (approved):

PR #1196, opened 2022-07-11
Added redact using insertion filter

RoyUP9:
Removed old redaction implementation and replaced with redaction using insertion filter


cli/config/configStructs/tapConfig.go
```diff
@@ -71,9 +77,48 @@ func (config *TapConfig) GetInsertionFilter() string {
...
+		return fmt.Sprintf("(%s) and %s", insertionFilter, redactFilter)
+	} else if insertionFilter == "" && redactFilter != "" {
+		return redactFilter
+	}
+
 	return insertionFilter
 }
 
+func getRedactFilter(config *TapConfig) string {
+	if !config.EnableRedaction {
```
nimrod-up9, line 92: Is the `--redact` flag necessary? We can simplify the command by removing it. The rest of this function remains the same.

cli/config/configStructs/tapConfig.go
```diff
@@ -71,9 +77,48 @@ func (config *TapConfig) GetInsertionFilter() string {
 			}
 		}
 	}
+
+	redactFilter := getRedactFilter(config)
+	if insertionFilter != "" && redactFilter != "" {
+		return fmt.Sprintf("(%s) and %s", insertionFilter, redactFilter)
```
nimrod-up9, line 83:
```suggestion
		return fmt.Sprintf("(%s) and (%s)", insertionFilter, redactFilter)
```

cli/config/configStructs/tapConfig.go
nimrod-up9: And because this CLI doesn't pass `config.EnableRedaction` anymore, then I'm not sure how this works.
https://github.com/up9inc/mizu/pull/1196/files#diff-3cb0ddb013779515a67d80f2ff08e2d4978c08c96ddb05ca3f65dcda083426ceL249

cli/config/configStructs/tapConfig.go
RoyUP9: redact filter is only one command - redact(<list_of_values_to_redact>)
so i'm not sure it's necessary

nimrod-up9 (approved):

cli/config/configStructs/tapConfig.go
nimrod-up9: There will be default values for redact filters, but we want the feature to be disabled by default. Therefore keep the `--redact` flag.

PR #1205, opened 2022-07-14
Handle `item.Pair.Response.Payload` is `nil` case in AMQP

mertyildiran:


gadotroee:
We can close this because it happens only when there is incompatible versions of api-server and tapper. 
when the versions are compatible it will never happen.

PR #1319, opened 2023-03-18
Upgrade github actions versions to latest and update deprecated options

chrisRedwine:
This updates several GitHub Actions versions and uses latest syntax for options on them:

- Upgrade `actions/checkout` to latest (`v3`) to avoid node version warnings
- Upgrade `actions/setup-go` to latest (`v4`) and use `go.mod` as the single source of truth for the go version
- Upgrade `golangci/golangci-lint-action` to latest (`v3`)
- Upgrade `codecov/codecov-action` to latest (`v3`)
- Upgrade `goreleaser/goreleaser-action` to latest (`v4`) and use `--clean` instead of deprecated `--rm-dist` (see [notice](https://goreleaser.com/deprecations/#-rm-dist))

If this would be better as separate PRs, definitely let me know.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -16,16 +16,18 @@ jobs:
     name: Golint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v3
         with:
           fetch-depth: 2
 
-      - uses: actions/setup-go@v2
+      - name: Set up Go
+        uses: actions/setup-go@v4
         with:
-          go-version: '^1.17'
+          go-version-file: 'go.mod'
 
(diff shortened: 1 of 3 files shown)


chrisRedwine:
@mertyildiran Hmm, looking at the action runs, it looks like previously it was setting up the _latest_ version of go even though it was marked `Set up Go 1.17` - whereas the way I changed it causes it to run on `1.17`. If the intention is in fact to run against the latest version for all three workflows, then it should probably be changed to `go-version: 'stable'` and I can remove the `go-version-file: 'go.mod'`. Thoughts?

mertyildiran (approved):

mertyildiran:
@chrisRedwine looks fine. Thanks for the contribution!

PR #1350, opened 2023-04-28
Set default Storageclass, rewritten namespace and chart version

GeiserX:
Fixes:
- https://github.com/kubeshark/kubeshark/issues/1349
- https://github.com/kubeshark/kubeshark/issues/1348
- https://github.com/kubeshark/kubeshark/issues/1343


GeiserX:
I deleted the `00-namespace.yml` file because in Helm charts this is usually provisioned by other external means, breaking compatibility with usual tooling. Another option would be to create it depending on a value in `values.yaml`

GeiserX:
The thing is, that the `ReadWriteMany` setting on the PVC is incompatible with any of the default StorageClasses that I have in my clusters in GKE, EKS nor Rancher... This is a feature not much implemented in the principal Kubernetes providers. I would suggest rewriting the app so as to use some kind of external DB. Otherwise we will need to install NFS servers.

frimik:
On AWS EKS, *EFS* with dynamic provisioning would be required for this PV sharing to work. Having all of that in place is rather rare.

helm-chart/Chart.yaml
```diff
@@ -22,4 +22,4 @@ name: kubeshark
 sources:
   - https://github.com/kubeshark/kubeshark/tree/master/helm-chart
 type: application
-version: "40.0"
+version: "0.1.0"
```
mertyildiran, line 25: This version should match the CLI version `kubeshark version` which is `40.0` at this point.

I guess it's related to https://github.com/kubeshark/kubeshark/issues/1348

Please see the note in [here](https://github.com/kubeshark/kubeshark/wiki/CHANGELOG) to understand how we ended up with something other than SemVer. So, the repository version scheme should change for this.

helm-chart/templates/01-service-account.yaml
```diff
@@ -9,4 +9,4 @@ metadata:
     kubeshark-created-by: kubeshark
     kubeshark-managed-by: kubeshark
   name: kubeshark-service-account
-  namespace: '{{ .Values.tap.selfnamespace }}'
+  namespace: '{{ default "kubeshark" .Release.Namespace }}'
```
mertyildiran, line 12: The template files are generated automatically. So, please don't modify them directly. See the first line:

https://github.com/kubeshark/kubeshark/blob/08751add1180e7d89bada9228b059739d1e70b27/helm-chart/templates/01-service-account.yaml#L1

The correct way to modify the line is through changing this line:

https://github.com/kubeshark/kubeshark/blob/2563cc19221fb5007411613694777c0e80b717c1/cmd/helmChart.go#L116

Related to https://github.com/kubeshark/kubeshark/issues/1349

We have an existing way of configuring the Kubeshark deployment through `config.yaml` / `kubeshark.yaml` file. So, the namespace should be sourced from the `config.yaml`. [See the docs](https://docs.kubeshark.co/en/config).

helm-chart/values.yaml
```diff
@@ -17,9 +17,8 @@ tap:
     host: 127.0.0.1
   regex: .*
   namespaces: []
-  selfnamespace: kubeshark
   storagelimit: 200Mi
-  storageclass: standard
+  storageclass: ""
```
mertyildiran, line 21: This file is also generated automatically and shouldn't be modified directly. The `values.yaml` should match the output of `kubeshark config` command which prints the `config.yaml`. (prints a default structure if `$HOME/.kubeshark/config.yaml` is not present)

`values.yaml` is automatically generated on these lines: https://github.com/kubeshark/kubeshark/blob/2563cc19221fb5007411613694777c0e80b717c1/cmd/helmChart.go#L350-L360

mertyildiran (changes requested):

helm-chart/templates/01-service-account.yaml
mertyildiran: To clarify; we should generate such a template directive that prioritizes `.Release.Namespace` over `.Values.tap.selfnamespace`. We should also not remove `00-namespace.yaml` but instead generate such a template directive that disables `00-namespace.yaml` if `.Release.Namespace` is present or `.Values.tap.selfnamespace` is something other than `kubeshark`.

mertyildiran:
Closing in favor of current `HEAD` (dfe5605032fa34799a123ef88acd4b35b7d1399e)

PR #1392, opened 2023-07-19
✨Allow to disable IPv6

papanito:
I figured it would make sense to configure whether to use ipv6 or not. Currently - as we have not enabled ipv6 yet - I am unable to get kubeshark running

### DIFF
--- a/helm-chart/README.md
+++ b/helm-chart/README.md
@@ -61,6 +61,7 @@ helm install kubeshark kubeshark/kubeshark \
   --set "tap.ingress.auth.approveddomains={gmail.com}" \
   --set license=LICENSE_GOES_HERE
 ```
+
 You can get your license [here](https://console.kubeshark.co/).
 
 ## Installing with Persistent Storage Enabled
@@ -70,4 +71,14 @@ helm install kubeshark kubeshark/kubeshark \
   --set tap.persistentstorage=true \
   --set license=LICENSE_GOES_HERE
 ```
+
 You can get your license [here](https://console.kubeshark.co/).
+
+## Disabling IPV6
+
+Not all have IPV6 enabled, hence this has to be disabled as follows:
+
+```shell
+helm install kubeshark kubeshark/kubeshark \
+  --set tap.ipv6=false
+```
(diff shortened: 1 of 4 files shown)


helm-chart/README.md
```diff
@@ -70,4 +71,14 @@ helm install kubeshark kubeshark/kubeshark \
...
+
 You can get your license [here](https://console.kubeshark.co/).
+
+## Disabling IPV6
+
+Not all have IPV6 enabled, hence this has to be disabled as follows:
+
+```shell
+helm install kubeshark kubeshark/kubeshark \
+  --set tap.ingress.ipv6=false
```
mertyildiran, line 83: IPv6 support is not Ingress specific. Should be moved to `tap.ipv6`

helm-chart/templates/12-nginx-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ConfigMap
+metadata:
+  name: kubeshark-nginx-config
+  namespace: {{ .Release.Namespace }}
+  apiVersion: v1
+data:
+  default.conf: |
+    server {
+      listen 80;
+{{- if .Values.tap.ingress.ipv6 }}
```
mertyildiran, line 12: IPv6 support is not Ingress specific. Should be moved to `tap.ipv6`

helm-chart/values.yaml
```diff
@@ -55,6 +55,7 @@ tap:
     auth:
       approveddomains: []
     certmanager: letsencrypt-prod
+    ipv6: true
```
mertyildiran, line 58: IPv6 support is not Ingress specific. Should be moved to `tap.ipv6`

helm-chart/templates/12-nginx-config.yaml
```diff
@@ -0,0 +1,26 @@
...
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;
+        expires -1;
+      }
+      error_page   500 502 503 504  /50x.html;
+      location = /50x.html {
+        root   /usr/share/nginx/html;
+      }
+    }
```
mertyildiran, line 26: There has to be a new line at the end of each file.

mertyildiran (changes requested):

mertyildiran (approved):

Issue #1394, opened 2023-07-21
[Feature Request:] Helm chart overhaul


InputObject2:
### Contact Details

InputObject2 (github), InputObject (discord)

### Is your feature request related to a problem? Please describe.

I was looking at the helm chart and I was wondering why some choices were made, for example using pods directly for the hub and front (instead of a deployment or statefulset) or the services using nodeports when we're also exposing via an ingress. 

There's a fair bit of changes I'd like to contribute that wouldn't all be backwards compatible with the current value file. I see that it is used a lot in the go code so I might have to change some of it to match. 

### Original Thread

Discord #kubernetes

### Describe the solution you'd like to see

In no particular order, here are some of the things I was considering:

- using deployments instead of single pods for hub and front, allowing updates and rollouts
- splitting the front and hub configs out of the tap block
- using resource and tolerations blocks everywhere instead of making them separate fields
- allow customizing the services for hub and front
- splitting the ingress block out of the tap block
- reworking the ingress class
- giving each component it's own optional labels and annotations inside their own block
- group and isolate the app config options (like namespaces or packetcapture) inside the tap block so that they are not mixed in with the kubernetes options (like resources or storage)

### Provide additional context

I think it could be a net gain overall. I found the current setup is kinda hard to grasp without going deep into the code.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Ziul:
Here are several steps we need to take in order to achieve these desired outcomes. I'm listing them in a sequence that I believe will minimize any disruptions caused by necessary adjustments to the Go lang code.

1. Adding labels to all resources. Currently, resources are identified by their names in the CLI, hub, and workers, but they should be sorted by labels. By adding labels to all resources, we can modify the types and structures of the helm chart without affecting the CLI's functionality.
2. Replacing pods with deployments. Without utilizing labels or name prefixes, this step becomes challenging to execute without causing communication breakdowns among services.
3. Loading application options/configurations from secrets/configmaps, with the option to specify an existing secret/configmap's name. This approach should simplify the subsequent step and enable the secure passage of parameters such as licenses or flags.
4. Separating the tap block configurations from Kubernetes configurations. This step is crucial for achieving a cleaner `values.yaml` file.
5. Removing the ingress class manifest by using a pre-defined ingress class specified by the user.
6. Distinctly `labeling`, `annotating`, allocating `resources`, and setting `tolerations` for each component within the `values.yaml` file.

PR #1395, opened 2023-07-21
helm chart overhaul

InputObject2:
First draft for #1394 here, mainly looking for feedback, some more things will need to change before this is fully functional.

helm template works and most of the chart settings I wanted to change are there.

## Global
- Removed creationTimestamp from everything as it will be setup automatically anyway
- Removed namespace from everything since helm will set the value to .Release.Namespace by default
- Removed the annotations and labels field from everything if not used (the "if" clause starts above it so setting `labels: {}` in most cases will result in no `labels` field to be created.
- Replaced "kubeshark" from most of the resource names to replace with the .Release.Name instead to match what is being done in other projects.

## RBAC
- Created the `rbac` structure to isolate settings to only the `serviceAccount`, `ClusterRole` and `ClusterRoleBinding`

### Workload
- Created the `hub` structure to isolate settings for the hub.
- Converted the pod to a deployment instead.
- Replaced the individual fields for resources with an ad-lib block. This will allow setting only some (or none) of the fields.
- Replaced the tolerations with an ad-lib block. The default should be that the deployment has no toleration and experienced users with specific use-cases may add some as needed.
- Added more customization to the container image definition
- Added the containerPort definition to explicitly expose the port that was being used for the hub.

### Service
- Modified the service to be able to set another port if needed
- Targeted the `http-api` on the hub deployment

### Workload
- Created the `front` structure to isolate settings for the front.
- Converted the pod to a deployment instead.
- Replaced the hardcoded resources with an ad-lib block. This will allow setting custom resources instead of having them hardcoded.
- Replaced the tolerations with an ad-lib block. The default should be that the deployment has no toleration and experienced users with specific use-cases may add some as needed.
- Added more customization to the container image definition
- Added the containerPort definition to explicitly expose the port that was being used for the front.
- Set the `REACT_APP_HUB_PORT` to use the hub port defined in the hub service instead.

### Service
- Modified the service to be able to set another port if needed
- Targeted the `http-frontend` on the front deployment. 

## Tap workers

- Renamed the daemonset to `{{ .Release.Name}}-worker-tap` since having `daemonset` in the name is redundant.
- Replaced the individual fields for resources with an ad-lib block. This will allow setting only some (or none) of the fields.
- Replaced the tolerations with an ad-lib block. The default should be that the deployment has no toleration and experienced users with specific use-cases may add some as needed.
- Added more customization to the container image definition

## Ingress
- Moved the `ingress` and `ingressClass` out of the `tap` structure.
- Separated the ingress and ingressClass structures to allow for specific values.
- Enabling ingressClass will now force using the custom ingressClass, while `ingressClass.enabled: false` will result in using the value set in `ingress.classname`

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -1,61 +1,122 @@
+# Configs for the worker-tap daemonset
 tap:
-  docker:
+  image:
     registry: docker.io/kubeshark
     tag: latest
-    imagepullpolicy: Always
-    imagepullsecrets: []
-  proxy:
-    worker:
-      srvport: 8897
-    hub:
-      port: 8898
-      srvport: 8898
-    front:
-      port: 8899
-    host: 127.0.0.1
-  regex: .*
-  namespaces: []
-  release:
-    repo: https://helm.kubeshark.co
-    name: kubeshark
-    namespace: default
-  persistentstorage: false
-  storagelimit: 200Mi
-  storageclass: standard
-  dryrun: false
-  pcap: ""
+    name: worker
+  imagepullpolicy: Always
+  port: 8897
+  persistence:
+    enabled: false
+    size: 200Mi
+    storageclass: standard
   resources:
-    worker:
-      limits:
-        cpu: 750m
-        memory: 1Gi
-      requests:
-        cpu: 50m
-        memory: 50Mi
-    hub:
-      limits:
-        cpu: 750m
-        memory: 1Gi
-      requests:
-        cpu: 50m
-        memory: 50Mi
-  servicemesh: true
-  tls: true
-  packetcapture: libpcap
-  ignoretainted: false
+    limits:
+      cpu: 750m
+      memory: 1Gi
+    requests:
+      cpu: 450m
+      memory: 150Mi
   labels: {}
   annotations: {}
+  tolerations: {}
   nodeselectorterms: []
-  ingress:
-    enabled: false
-    classname: kubeshark-ingress-class
-    controller: k8s.io/ingress-nginx
-    host: ks.svc.cluster.local
-    tls: []
-    auth:
-      approveddomains: []
-    certmanager: letsencrypt-prod
-  debug: false
+  
+  # these settings are used to configure the tap's capture settings
+  config:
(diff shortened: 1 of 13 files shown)


helm-chart/values.yaml
```diff
@@ -1,61 +1,122 @@
+# Configs for the worker-tap daemonset
 tap:
-  docker:
+  image:
```
mertyildiran, line 3: The structure of `values.yaml` should match this Go struct:

https://github.com/kubeshark/kubeshark/blob/470ab3d7ed192a9c52b4b16f9638a037c36f16b2/config/configStruct.go#L29-L39

In fact, `values.yaml` shouldn't be edited manually but instead generated automatically using `make generate-helm-values` command:

https://github.com/kubeshark/kubeshark/blob/470ab3d7ed192a9c52b4b16f9638a037c36f16b2/Makefile#L72-L73

mertyildiran (changes requested):

InputObject2:
Alright, I've been working on this but it feels like anytime I make a change it get bigger:
- helm chart changes require changes in values.yaml
- values.yaml is generated by config and uses a configStruct
- configStruct defines a bunch of sub structs to define the different aspects of the config
- these values cannot be generalized because we're using "defaults" to generate the values file with the default values hardcoded in the go code matching each of these.
- can't really do otherwise because we use the user-provided values and add the default or null values for all the fields they didn't fill in
- on the subject of resources, usually projects don't specify them to leave that up to the users to figure out instead. I guess you're going for a one-step-everything-works setup that wouldn't need any config. Same applies for tolerations, annotations and labels usually.

Some solutions I'm seeing:
- we could add a default value file to fill in all the different values (for example different default resources or images for hub/front/workers) but then what's the point of generating all this using go and not just template them using the helm lib to combine the default yaml and then override them using the customer's value file?
- we could add a lot of logic in the helm chart code to account for default values here instead, but it would make the helm chart almost unreadable for anyone unfamiliar with helm trying to debug this
- we could add a bunch of go code duplicating the structures but this isn't very DRY

Anyway this is turning out to be a major refactor,

PR #1400, opened 2023-07-31
Create pod.go

Achyut-Krishna:
perform the deployment, labeling, and pod creation with hooks, respectively.

### DIFF
--- a/pod/pod.go
+++ b/pod/pod.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"context"
+	"fmt"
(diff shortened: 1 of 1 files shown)


mertyildiran:
@Achyut-Krishna the code in `pod/pod.go` is a separate program by itself and it does integrate with the existing codebase. Overall, I couldn't understand how the code in the PR helpful or achieves any purpose.

Additionally, PR has missing `go.mod` and `go.sum` files.

PR #1406, opened 2023-08-08
:hammer: Add `_helpers.tpl` and `NOTES.txt` to Helm chart and refactor `labels`

Ziul:
This PR includes:

- standard labels using a `_helpers.tpl`  file to create those labels.
- removed all the `creationTimestamp` inputs.
- Included a `NOTES.txt` file to help users after the installing.
- Service Account Name also generated by `_helpers.tpl`  file.

### DIFF
--- a/helm-chart/templates/01-service-account.yaml
+++ b/helm-chart/templates/01-service-account.yaml
@@ -8,5 +8,5 @@ metadata:
   {{- if .Values.tap.annotations }}
     {{- toYaml .Values.tap.annotations | nindent 4 }}
   {{- end }}
-  name: kubeshark-service-account
+  name: {{ include "kubeshark.serviceAccountName" . }}
   namespace: {{ .Release.Namespace }}
--- a/helm-chart/templates/01-service-account.yaml
+++ b/helm-chart/templates/01-service-account.yaml
@@ -2,11 +2,8 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  creationTimestamp: null
   labels:
-  {{- if .Values.tap.labels }}
-    {{- toYaml .Values.tap.labels | nindent 4 }}
-  {{- end }}
+    {{- include "kubeshark.labels" . | nindent 4 }}
   annotations:
   {{- if .Values.tap.annotations }}
     {{- toYaml .Values.tap.annotations | nindent 4 }}
--- a/helm-chart/templates/11-ingress.yaml
+++ b/helm-chart/templates/11-ingress.yaml
@@ -6,9 +6,7 @@ metadata:
   annotations:
     certmanager.k8s.io/cluster-issuer: {{ .Values.tap.ingress.certmanager }}
     nginx.ingress.kubernetes.io/rewrite-target: /$2
-  {{- if .Values.tap.annotations }}
-    {{- toYaml .Values.tap.annotations | nindent 4 }}
-  {{- end }}
+    {{- include "kubeshark.labels" . | nindent 4 }}
   creationTimestamp: null
   labels:
   {{- if .Values.tap.labels }}
--- a/helm-chart/templates/11-ingress.yaml
+++ b/helm-chart/templates/11-ingress.yaml
@@ -6,12 +6,11 @@ metadata:
   annotations:
     certmanager.k8s.io/cluster-issuer: {{ .Values.tap.ingress.certmanager }}
     nginx.ingress.kubernetes.io/rewrite-target: /$2
-    {{- include "kubeshark.labels" . | nindent 4 }}
-  creationTimestamp: null
-  labels:
-  {{- if .Values.tap.labels }}
-    {{- toYaml .Values.tap.labels | nindent 4 }}
+  {{- if .Values.tap.annotations }}
+    {{- toYaml .Values.tap.annotations | nindent 4 }}
   {{- end }}
+  labels:
+    {{- include "kubeshark.labels" . | nindent 4 }}
   name: kubeshark-ingress
   namespace: {{ .Release.Namespace }}
 spec:
--- a/helm-chart/templates/NOTES.txt
+++ b/helm-chart/templates/NOTES.txt
@@ -1,0 +1,27 @@
+Thank you for installing {{ .Chart.Name }}.
+
+Your deployment has been successful. The release is named {{ .Release.Name }} and it has been deployed in the {{ .Release.Namespace }} namespace.
+
+{{- if .Values.tap.ingress.enabled }}
+
+{{ if not .Values.license -}}
+warning:
+> Ingress option enabled but license not set. The application should not work as expected.
+> Get a lisence at https://console.kubeshark.co/
+{{- else }}
+You can now access the application through the following URL: 
+http{{ if .Values.tap.ingress.tls }}s{{ end }}://{{ .Values.tap.ingress.host }}
+{{- end -}}
+
+{{- else }}
+To access the application, follow these steps:
+
+1. Perform port forwarding with the following commands:
+
+    kubectl port-forward -n {{ .Release.Namespace }} service/kubeshark-hub 8898:80 & \
+    kubectl port-forward -n {{ .Release.Namespace }} service/kubeshark-front 8899:80
+
+2. Once port forwarding is done, you can access the application by visiting the following URL in your web browser:
+    http://0.0.0.0:8899
+
+{{ end }}
--- a/helm-chart/templates/NOTES.txt
+++ b/helm-chart/templates/NOTES.txt
@@ -9,7 +9,7 @@ warning:
 > Ingress option enabled but license not set. The application should not work as expected.
 > Get a lisence at https://console.kubeshark.co/
 {{- else }}
-You can now access the application through the following URL: 
+You can now access the application through the following URL:
 http{{ if .Values.tap.ingress.tls }}s{{ end }}://{{ .Values.tap.ingress.host }}
 {{- end -}}
 
--- a/helm-chart/templates/NOTES.txt
+++ b/helm-chart/templates/NOTES.txt
@@ -7,7 +7,7 @@ Your deployment has been successful. The release is named {{ .Release.Name }} an
 {{ if not .Values.license -}}
 warning:
 > Ingress option enabled but license not set. The application should not work as expected.
-> Get a lisence at https://console.kubeshark.co/
+> Get a license at https://console.kubeshark.co/
 {{- else }}
 You can now access the application through the following URL:
 http{{ if .Values.tap.ingress.tls }}s{{ end }}://{{ .Values.tap.ingress.host }}
--- a/helm-chart/templates/NOTES.txt
+++ b/helm-chart/templates/NOTES.txt
@@ -1,4 +1,4 @@
-Thank you for installing {{ .Chart.Name }}.
+Thank you for installing {{ title .Chart.Name }}.
 
 Your deployment has been successful. The release is named {{ .Release.Name }} and it has been deployed in the {{ .Release.Namespace }} namespace.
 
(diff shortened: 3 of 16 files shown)


helm-chart/templates/01-service-account.yaml
```diff
@@ -2,14 +2,11 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  creationTimestamp: null
   labels:
-  {{- if .Values.tap.labels }}
-    {{- toYaml .Values.tap.labels | nindent 4 }}
-  {{- end }}
+    {{- include "kubeshark.labels" . | nindent 4 }}
```
mertyildiran, line 6: Setting:

```yaml
tap:
  labels:
    foo: bar
```

in `values.yaml` gives the error below:

```
Error: YAML parse error on kubeshark/templates/01-service-account.yaml: error converting YAML to JSON: yaml: line 9: mapping values are not allowed in this context
```

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- toYaml .Values.tap.labels}}
+{{- end }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "kubeshark.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "kubeshark.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
```
mertyildiran, line 56:
```suggestion
app.kubernetes.io/instance: {{ eq .Release.Name "release-name" | ternary "kubeshark" .Release.Name }}
```

Wouldn't it be better if I was something like this because when I do `helm template .` I see the label: `app.kubernetes.io/instance: release-name`

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "kubeshark.serviceAccountName" -}}
+{{- if and .Values.serviceAccount .Values.serviceAccount.create }}
+{{- default (include "kubeshark.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- printf "%s-service-account" .Release.Name }}
```
mertyildiran, line 66:
```suggestion
{{- printf "%s-service-account" (eq .Release.Name "release-name" | ternary "kubeshark" .Release.Name) }}
```

Wouldn't it be better if I was something like this because when I do `helm template .` I see the service account name as: `release-name-service-account` while it should be `kubeshark-service-account`.

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- define "kubeshark.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "kubeshark.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "kubeshark.serviceAccountName" -}}
+{{- if and .Values.serviceAccount .Values.serviceAccount.create }}
```
mertyildiran, line 63: `serviceAccount.create` does not exist in `values.yaml`.

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- default (include "kubeshark.fullname" .) .Values.serviceAccount.name }}
```
mertyildiran, line 64: `serviceAccount.name` does not exist in `values.yaml`.

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "kubeshark.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
```
mertyildiran, line 5: `nameOverride` field does not exist in `values.yaml`.

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "kubeshark.fullname" -}}
+{{- if .Values.fullnameOverride }}
```
mertyildiran, line 14: `fullnameOverride` field does not exist in `values.yaml`.

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
```
mertyildiran, line 15: `fullnameOverride` field does not exist in `values.yaml`.

helm-chart/templates/_helpers.tpl
```diff
@@ -0,0 +1,68 @@
...
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
```
mertyildiran, line 17: `nameOverride` field does not exist in `values.yaml`.

helm-chart/templates/11-ingress.yaml
```diff
@@ -6,9 +6,7 @@ metadata:
   annotations:
     certmanager.k8s.io/cluster-issuer: {{ .Values.tap.ingress.certmanager }}
     nginx.ingress.kubernetes.io/rewrite-target: /$2
-  {{- if .Values.tap.annotations }}
-    {{- toYaml .Values.tap.annotations | nindent 4 }}
-  {{- end }}
+    {{- include "kubeshark.labels" . | nindent 4 }}
```
mertyildiran, line 9: The labels are incorrectly templated into the `annotations` field instead of `labels` field.

helm-chart/templates/01-service-account.yaml
```diff
@@ -2,14 +2,11 @@
...
   annotations:
   {{- if .Values.tap.annotations }}
     {{- toYaml .Values.tap.annotations | nindent 4 }}
   {{- end }}
-  name: kubeshark-service-account
+  name: {{ include "kubeshark.serviceAccountName" . }}
```
mertyildiran, line 11: What is the motivation to template the service account name?

helm-chart/templates/NOTES.txt
```diff
@@ -0,0 +1,27 @@
+Thank you for installing {{ .Chart.Name }}.
```
mertyildiran, line 1:
```suggestion
Thank you for installing {{ title .Chart.Name }}.
```

helm-chart/templates/NOTES.txt
```diff
@@ -0,0 +1,27 @@
...
+
+Your deployment has been successful. The release is named {{ .Release.Name }} and it has been deployed in the {{ .Release.Namespace }} namespace.
+
+{{- if .Values.tap.ingress.enabled }}
+
+{{ if not .Values.license -}}
+warning:
+> Ingress option enabled but license not set. The application should not work as expected.
+> Get a lisence at https://console.kubeshark.co/
```
mertyildiran, line 10:
```suggestion
> Get a license at https://console.kubeshark.co/
```

typo.

mertyildiran (changes requested):

mertyildiran:
Also please be aware of the recently added `12-ngingx-config.yaml` file (https://github.com/kubeshark/kubeshark/pull/1392)

helm-chart/templates/_helpers.tpl
Ziul: You see `release-name` because you installed it as `helm install release-name .` And That's what we expect when we install a helm chart. If you install cilium as `helm install release-name cilium/cilium` it will create every resource and label with a `release-name` on it. That's one reason why we use labels to filter resources, as [by default](https://helm.sh/docs/chart_best_practices/labels/#standard-labels), the `app.kubernetes.io/instance` label should be `{{ .Release.Name }}`.

helm-chart/templates/_helpers.tpl
Ziul: Yes, it doesn't exist yet. Service Account gonna need future work so the user can have more flexibility with it for security concerns. For now, the way I set still creates the same service account name using the  `.Release.Name` as value

helm-chart/templates/_helpers.tpl
Ziul: No, it doesn't yet.

helm-chart/templates/01-service-account.yaml
Ziul: Templating the Service Account name in a Helm chart brings consistency, automation, scalability, and adaptability to the deployment process. It aligns with best practices for resource management, enhances security through naming conventions, and supports the efficient management of resources across diverse environments.  it's important to carefully consider the potential drawbacks in terms of complexity, standardization, compatibility, debugging, flexibility, maintenance, and the learning curve.

Ziul:
Nice! I'll include the labels on that config than.

helm-chart/templates/_helpers.tpl
Ziul: The `fullnameOverride` and `nameOverride` fields in Helm charts provide a mechanism to customize and control the naming of Kubernetes resources generated by the chart during deployment. These fields are particularly useful when you want to fine-tune the naming of resources to align with your specific project or organization's naming conventions, or to prevent naming conflicts in multi-environment deployments.
This can be especially valuable in scenarios where you are deploying multiple instances of the same chart in the same cluster or across different clusters or slight variations based on the deployment context.

We don't have it now on `values.yaml`, but the inclusion of these terms is willingly embraced and should be included in the future.

helm-chart/templates/_helpers.tpl
mertyildiran: @Ziul we have a Makefile rule named `generate-manifests` in here: https://github.com/kubeshark/kubeshark/blob/65bda4e84432a1f8306d75f9abddc51b09e55acb/Makefile#L75-L76
Which runs `helm template ./helm-chart` command. So, we would want to replace the default name `release-name` with `kubeshark`.

helm-chart/templates/01-service-account.yaml
Ziul: - A company that requires that every service account have the name of the team on it
- A policy by a third-party team that restricts access based on a pattern in the names

I saw those two things happen in different companies. In both cases, the team requesting a tool to be installed wasn't able to change/suggest internal policies. The question is why the user couldn't change the default name.

helm-chart/templates/_helpers.tpl
Ziul: Resolved updating the Makefile

helm-chart/templates/_helpers.tpl
Ziul: keeping it next to the [thread](https://github.com/kubeshark/kubeshark/pull/1406/files#r1288464716) conversation.

helm-chart/templates/_helpers.tpl
Ziul: keeping it next to the [thread](https://github.com/kubeshark/kubeshark/pull/1406/files#r1288464716) conversation.

helm-chart/templates/_helpers.tpl
Ziul: keeping it next to the [thread](https://github.com/kubeshark/kubeshark/pull/1406/files#r1288464716) conversation.

helm-chart/templates/01-service-account.yaml
mertyildiran: I understood, thanks!

mertyildiran (approved):

PR #1412, opened 2023-08-15
:rocket: Change Hub's and Front's resource type from `Pod` to `Deployment`

Ziul:


helm-chart/templates/04-hub-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+            - ./hub
+            {{ .Values.tap.debug | ternary "- -debug" "" }}
+          envFrom:
+            - configMapRef:
+                name: kubeshark-hub-config
+            - secretRef:
+                name: kubeshark-hub-secret
+          env:
+            - name: POD_REGEX
+              value: '{{ .Values.tap.regex }}'
```
mertyildiran, line 38:
```suggestion
```

Duplicate. Already defined in `kubeshark-hub-config`.

kubernetes/provider.go
```diff
@@ -127,6 +127,30 @@ func (provider *Provider) ListAllRunningPodsMatchingRegex(ctx context.Context, r
...
+
+func (provider *Provider) ListAllNamespaces(ctx context.Context) ([]core.Namespace, error) {
+	namespaces, err := provider.clientSet.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return namespaces.Items, err
+}
+
```
mertyildiran, line 153:
```suggestion
```

Dead code.

kubernetes/proxy.go
```diff
@@ -24,6 +24,7 @@ const selfServicePort = 80
 
 func StartProxy(kubernetesProvider *Provider, proxyHost string, srcPort uint16, selfNamespace string, selfServiceName string) (*http.Server, error) {
 	log.Info().
+		Str("proxyHost", proxyHost).
```
mertyildiran, line 27:
```suggestion
		Str("proxy-host", proxyHost).
```

We use `kebab-case` style for debugging values in `log` calls project-wide.

mertyildiran (changes requested):

helm-chart/templates/04-hub-deployment.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1
+kind: Deployment
```
mertyildiran, line 2:
```suggestion
kind: StatefulSet
```

Hub should be `StatefulSet`.

mertyildiran (changes requested):

helm-chart/templates/04-hub-deployment.yaml
Ziul: From what I saw its uses, it doesn't need to be StatefulSet. Workers can easily find it, as it keeps watching.

mertyildiran: OK, let's keep it as `Deployment`.

mertyildiran (approved):

PR #1417, opened 2023-08-24
:building_construction: Give the user ability to set ingress as needed

Ziul:
- Removed unecessary IngressClass.
- If no IngressClassName passed, use cluster's default class
- Renamed `ingressclass` with `IngressClassName`. Is the standard name used for it.
- Included custom annotations for Ingress. This way user can set any custom annotation for the ingress only.


helm-chart/templates/11-ingress.yaml
```diff
@@ -4,17 +4,18 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   annotations:
-    certmanager.k8s.io/cluster-issuer: {{ .Values.tap.ingress.certmanager }}
     nginx.ingress.kubernetes.io/rewrite-target: /$2
-  {{- if .Values.tap.annotations }}
-    {{- toYaml .Values.tap.annotations | nindent 4 }}
+  {{- if .Values.tap.ingress.annotations }}
+    {{- toYaml .Values.tap.ingress.annotations | nindent 4 }}
```
mertyildiran, line 9:
```suggestion
{{- if .Values.tap.annotations }}
    {{- toYaml .Values.tap.annotations | nindent 4 }}
{{- end }}
  {{- if .Values.tap.ingress.annotations }}
    {{- toYaml .Values.tap.ingress.annotations | nindent 4 }}
```

config/configStructs/tapConfig.go
```diff
@@ -87,11 +87,10 @@ type AuthConfig struct {
 
 type IngressConfig struct {
 	Enabled     bool                    `yaml:"enabled" json:"enabled" default:"false"`
-	ClassName   string                  `yaml:"classname" json:"classname" default:"kubeshark-ingress-class"`
-	Controller  string                  `yaml:"controller" json:"controller" default:"k8s.io/ingress-nginx"`
+	ClassName   string                  `yaml:"ingressClassName" json:"ingressClassName" default:"kubeshark-ingress-class"`
```
mertyildiran, line 90:
```suggestion
	ClassName   string                  `yaml:"classname" json:"classname" default:"kubeshark-ingress-class"`
```

helm-chart/values.yaml
```diff
@@ -28,12 +28,11 @@ tap:
   dryrun: false
   ignoretainted: false
   ingress:
-    certmanager: letsencrypt-prod
-    classname: kubeshark-ingress-class
-    controller: k8s.io/ingress-nginx
     enabled: false
-    host: ks.svc.cluster.local
-    tls: null
+    # ingressClassName: nginx
```
mertyildiran, line 32:
```suggestion
    ingressclass: kubeshark-ingress-class
```

helm-chart/templates/11-ingress.yaml
```diff
@@ -4,17 +4,18 @@ apiVersion: networking.k8s.io/v1
...
   {{- end }}
   labels:
     {{- include "kubeshark.labels" . | nindent 4 }}
   name: kubeshark-ingress
   namespace: {{ .Release.Namespace }}
 spec:
-  ingressClassName: {{ .Values.tap.ingress.classname }}
+  {{- if .Values.tap.ingress.ingressClassName }}
+  ingressClassName: {{ .Values.tap.ingress.ingressClassName }}
+  {{- end }}
```
mertyildiran, line 18:
```suggestion
  {{- if .Values.tap.ingress.classname }}
  ingressClassName: {{ .Values.tap.ingress.classname }}
  {{- end }}
```

mertyildiran (changes requested):

helm-chart/values.yaml
Ziul: That ingress-class does not exist anymore. The default value should be empty, so the cluster will use the default ingress class of the cluster.

mertyildiran (approved):

PR #1420, opened 2023-08-24
🔧  Add an option to disable telemetry into config

manfioLP:
now Telemetry can be set on/off based on config flag

if `tap.noTelemetry: true` no Telemetry will run, if `false` (default), telemetry will be sent.

### DIFF
--- a/cmd/tapRunner.go
+++ b/cmd/tapRunner.go
@@ -107,7 +107,7 @@ func tap() {
 	go watchHubPod(ctx, kubernetesProvider, cancel)
 	go watchFrontPod(ctx, kubernetesProvider, cancel)
 	if !config.Config.Tap.NoTelemetry {
-		// todo: add telemtry routine here
+		log.Info().Msg("Starting Telemetry...")
 	}
 
 	// block until exit signal or error
--- a/cmd/tapRunner.go
+++ b/cmd/tapRunner.go
@@ -106,6 +106,9 @@ func tap() {
 	go watchHubEvents(ctx, kubernetesProvider, cancel)
 	go watchHubPod(ctx, kubernetesProvider, cancel)
 	go watchFrontPod(ctx, kubernetesProvider, cancel)
+	if !config.Config.Tap.NoTelemetry {
+		// todo: add telemtry routine here
+	}
 
 	// block until exit signal or error
 	utils.WaitForTermination(ctx, cancel)
--- a/cmd/tapRunner.go
+++ b/cmd/tapRunner.go
@@ -75,6 +75,8 @@ func tap() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel() // cancel will be called when this function exits
 
+	log.Info().Bool("enabled", !config.Config.Tap.NoTelemetry).Msg("Telemetry")
+
 	state.targetNamespaces = kubernetesProvider.GetNamespaces()
 
 	log.Info().Strs("namespaces", state.targetNamespaces).Msg("Targeting pods in:")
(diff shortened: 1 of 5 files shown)


alongir:
Can config value be more simple, eg: tap.telemetry ?
Default is true. If set to false then it stops.

manfioLP:
> 

it can and I thought of, but after speaking with @Ziul he mentioned it's bad behavior to have a flag=false in Helm deployment

manfioLP:
> Can config value be more simple, eg: tap.telemetry ? Default is true. If set to false then it stops.

it can and I thought of, but after speaking with @Ziul he mentioned it's bad behavior to have a flag=false in Helm deployment

cmd/tapRunner.go
```diff
@@ -106,6 +106,9 @@ func tap() {
 	go watchHubEvents(ctx, kubernetesProvider, cancel)
 	go watchHubPod(ctx, kubernetesProvider, cancel)
 	go watchFrontPod(ctx, kubernetesProvider, cancel)
+	if !config.Config.Tap.NoTelemetry {
+		log.Info().Msg("Starting Telemetry...")
+	}
```
mertyildiran, line 111:
```suggestion
	log.Info().Bool("enabled", !config.Config.Tap.NoTelemetry).Msg("Telemetry")
```

Also print it before this line;

https://github.com/kubeshark/kubeshark/blob/7afb1d8b9b16011431da95e05dfcbee749d33bd8/cmd/tapRunner.go#L80

config/config.go
```diff
@@ -28,6 +28,7 @@ const (
 	FieldNameTag   = "yaml"
 	ReadonlyTag    = "readonly"
 	DebugFlag      = "debug"
+	TelemetryFlag  = "no-telemetry"
```
mertyildiran, line 31: Instead, use the `TELEMETRY_DISABLED` environment variable mentioned in https://github.com/kubeshark/hub/pull/4#discussion_r1305936600

helm-chart/templates/04-hub-deployment.yaml
```diff
@@ -28,6 +28,7 @@ spec:
           command:
             - ./hub
             {{ .Values.tap.debug | ternary "- -debug" "" }}
+            {{ .Values.tap.notelemetry | ternary "- -no-telemetry" "" }}
```
mertyildiran, line 31: Instead, use the `TELEMETRY_DISABLED` environment variable mentioned in https://github.com/kubeshark/hub/pull/4#discussion_r1305936600

mertyildiran (changes requested):

cmd/tapRunner.go
manfioLP: done

config/config.go
manfioLP: I think it wasn't used at all here, removed.

helm-chart/templates/04-hub-deployment.yaml
manfioLP: changed to:
```yaml
         command:
            - ./hub
            {{ .Values.tap.debug | ternary "- -debug" "" }}
            - e
            - TELEMETRY_DISABLED={{ .Values.tap.notelemetry }}
```

helm-chart/templates/04-hub-deployment.yaml
```diff
@@ -28,6 +28,8 @@ spec:
           command:
             - ./hub
             {{ .Values.tap.debug | ternary "- -debug" "" }}
+            - e
+            - TELEMETRY_DISABLED={{ .Values.tap.notelemetry }}
```
mertyildiran, line 32:
```suggestion
```

Add it to here instead:

https://github.com/kubeshark/kubeshark/blob/bf182b633041ebcb975728f41bce86f214328e8f/helm-chart/templates/13-config-map.yaml#L10-L16

mertyildiran (changes requested):

helm-chart/templates/04-hub-deployment.yaml
mertyildiran: Since outdated, continuing at https://github.com/kubeshark/kubeshark/pull/1420#discussion_r1309452513

mertyildiran:
Closing because of https://github.com/kubeshark/kubeshark/commit/fa1e7bcf0195f0c5a2e3d523730237c8c380139a

PR #1469, opened 2023-12-13
Kubeshark Operator for Openshift

kcns008:
TESTED LOCALLY

### Prerequisites

- Access to Openshift Cluster.
- Login as `cluster-admin`

Go to `KS-Operator` :

Apply Kubeshark CRD

```
oc apply -f config/crd/bases/apps.kubeshark.co_kubesharks.yaml
```

then

```
make run
```


alongir:
@kcns008 quick question. Does this need to be part of the kubeshark repo or does it make sense to make this its own repo?

kcns008:
@alongir Own repo for kubeshar-operator would be better option.

alongir:
> @alongir Own repo for kubeshar-operator would be better option.

So, let's do that.

alongir:
https://github.com/kubeshark/openshift-operator

PR #1476, opened 2024-01-09
Load pf-ring kernel module in init container

corest:
- Remove all `kmm` references
- Load kernel module via init container
- Update docs

Open questions:
- should we remove from `worker` codebase all the code related to loading kernel module?


alongir:
@corest, this can be a future feature request that comes in addition to this PR, however, one thing that I'm missing is for Kubeshark to NOT use PF-RING if `enabled=false`. Today, if PF-RING is loaded, Kubeshark will use it, even if this feature is disabled. It's not clean. Does this make sense? Is this for now? Is this for later?

alongir:
When I test it, I get this:
```yaml
 load-pf-ring Loading pf_ring module for kernel 5.10.192-183.736.amzn2.x86_64                                                                                                                       │
│ load-pf-ring insmod: ERROR: could not insert module /opt/lib/modules/5.10.192-183.736.amzn2.x86_64/pf_ring.ko: Invalid module format 
```
In any case, we should fail gracefully to AF_PACKET.
Worker goes into crashloop:
image

corest:
@alongir fixed to just log error and fallback into available methods for packet capture

alongir (approved):

alongir:
When running it, worker complained about this: https://github.com/kubeshark/kubeshark/blob/aaeb3ca1eb8d5cfa512fc1f7bc07c90fbb662317/helm-chart/templates/09-worker-daemon-set.yaml#L60 (-no-kernel-module)
When I removed it from the manifest, it didn;t load the PF-RING.

Issue #1467, opened 2023-12-12
Dynatrace / Kubeshark / Openshift compatibility


alongir:
A problem was reported when Kubeshark was installed on Openshift where Dynatrace was installed

This environment where the problem was reported:
```yaml
Client Version: 4.12.9
Kustomize Version: v4.5.7
Server Version: 4.12.19
Kubernetes Version: v1.25.8+37a9a08
```

- Set an environment
- Test compatibility
- Provide detail instructions how to bypass any found issues

tetatetit:
Hi @alongir , 

I'm unable to reproduce an issue  on the simple non-STS single zone (with 2 replicas) [ROSA (https://aws.amazon.com/es/rosa/)](https://aws.amazon.com/es/rosa/) managed OpenShift cluster with below versions on which I installed latest Dynatrace and then Kubeshark
```shell
% oc version                                                  
Client Version: 4.14.3
Kustomize Version: v5.0.1
Server Version: 4.12.19
Kubernetes Version: v1.25.8+37a9a08
```
and as you can see that all Kubeshark and Dynatrace pods are running successfully without any errors:
```shell
% kubectl get pods --all-namespaces | grep -E 'kubeshark|dynatrace'

default                                            kubeshark-front-647bcc7f66-nfr2j                                               1/1     Running            0                73m
default                                            kubeshark-hub-6f68c99d8-xtd8l                                                  1/1     Running            0                73m
default                                            kubeshark-worker-daemon-set-6pkhm                                              2/2     Running            2 (46m ago)      73m
default                                            kubeshark-worker-daemon-set-7qhz4                                              2/2     Running            1 (42m ago)      73m
default                                            kubeshark-worker-daemon-set-k5fwz                                              2/2     Running            1 (26m ago)      68m
default                                            kubeshark-worker-daemon-set-nmntq                                              2/2     Running            0                73m
default                                            kubeshark-worker-daemon-set-pbxjj                                              2/2     Running            4 (9m28s ago)    73m
default                                            kubeshark-worker-daemon-set-vbkd7                                              2/2     Running            1 (70m ago)      73m
default                                            kubeshark-worker-daemon-set-wd999                                              2/2     Running            0                73m
dynatrace                                          dynatrace-operator-77fdbcb56-vb8tp                                             1/1     Running            0                89m
dynatrace                                          dynatrace-webhook-5dd6dcc547-5dvmh                                             1/1     Running            0                89m
[...]

PR #1482, opened 2024-01-16
Option to exclude worker from Dynatrace monitoring

berezins:
#1467


corest:
This option is not needed for now

PR #1514, opened 2024-03-06
Switch to lower ports

corest:
Needs also release of frontend first after merging https://github.com/kubeshark/front/pull/134


corest:
Can be merged after front released

alongir (approved):

PR #1515, opened 2024-03-06
🔨 Disable scripting, targeted pods update & recording via `ConfigMap` keys

tiptophelmet:
Please, refer to the respective disable-features PRs in `hub` & `front`.


alongir:
Works well. One thing is missing though: The PCAP replay should also be disabled.

tiptophelmet:
> @tiptophelmet - Works well. One thing is missing though: The PCAP replay should also be disabled.

We already have `REPLAY_DISABLED` key in `ConfigMap` :ok_hand: 

It is used in `front` & `hub` as well.

alongir (approved):

PR #1517, opened 2024-03-13
✨ Set custom timezone

tiptophelmet:
Please, refer to a corresponding PR for setting custom timezone in `kubeshark/front`.


alongir (approved):

alongir:
I see a different time zone at the footer and at the payload. 
image

image

I realize that you get the payload from the worker. Is there a solution that will sync the timezones? Introduce same code at the worker level?
Also, I think where possible we need to somehow indicate the timezone to avoid confusion.
Lastly, I'm trying to change the timezone and getting only local.

PR #1525, opened 2024-04-01
add net capabilities for tracer

iluxa:
after https://github.com/kubeshark/tracer/pull/50 tracers operates with netlink and requires network capture capabilities enabled


alongir (approved):

mertyildiran:
I would add;

```yaml
    - NET_RAW
    - NET_ADMIN
```

into `ebpfCapture` instead. Because neither `ebpfCapture` nor `networkCapture` are conditional for the `tracer`. `ebpfCapture` is basically the required capabilities for the `tracer`.

PR #1558, opened 2024-07-12
Add debugger container

corest:
Towards https://github.com/kubeshark/worker/issues/152#issuecomment-2181211430

Adds https://github.com/kubeshark/debug-tools this container into worker daemonset if debug is enabled.


corest:
We agreed with @alongir to not add this into helm chart.
I'll deploy this container automatically in CI

PR #1592, opened 2024-08-23
Use env variables from configmap

corest:
Towards https://github.com/kubeshark/kubeshark/issues/1550


corest:
We decided to read this variable from configmap in code

PR #1634, opened 2024-10-28
Fix Helm Chart Annotations and Resource Requests/Limits Configuration

filippolmt:


filippolmt:
Hello @alongir , I made some changes to the chart, could you do a review?

filippolmt:
I performed a rebase, although it would probably have been better to merge this work before releasing a new version.

filippolmt:
@alongir  and @corest  will this PR ever be taken into consideration? Currently, if I try to deploy on my cluster, the daemonsets have the following error, which is due to the annotation.
```
kubeshark-worker-daemon-set-fwkxb   0/2     ImagePullBackOff
```

filippolmt:
I modified the if of the daemonsets.

corest (approved):
changes in values won't be persisted anyway as that file is generated from golang code, lgtm otherwise

filippolmt:
I got confused with the previous comment; I meant to mention @alongir . Can we proceed with merging this PR?

PR #1666, opened 2024-12-11
Add hub metrics port

corest:
Towards https://github.com/kubeshark/hub/issues/233


alongir (approved):

alongir:
@corest do we need also change: https://github.com/kubeshark/kubeshark/blob/master/helm-chart/metrics.md

corest:
@alongir we don't expose any metrics so I don't know if we need to extend that config. Unless we expect users to create dynamic metrics

PR #1671, opened 2024-12-20
Added security capabilities, especially IPC_LOCK

alongir:
to Sniffer in case eBPF traffic capture mechanism is used.


alongir:
@corest I needed to merge this, however, I still need your review.

PR #1726, opened 2025-03-21
:hammer: Create `tap.routing.front.basePath` flag

tiptophelmet:


helm-chart/values.yaml
```diff
@@ -131,6 +131,10 @@ tap:
     host: ks.svc.cluster.local
     tls: []
     annotations: {}
+  routing:
+    front:
+      # Example: /custompath
+      basePath: ""
```
mertyildiran, line 137:
```suggestion
```

I would make it `tap.ingress.path` instead.

helm-chart/templates/11-nginx-config-map.yaml
```diff
@@ -46,6 +46,17 @@ data:
...
+      location {{ .Values.tap.routing.front.basePath }} {
+        rewrite ^{{ .Values.tap.routing.front.basePath }}(.*)$ $1 break;
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;
+        expires -1;
+        add_header Cache-Control no-cache;
+      }
+{{- end }}
+
```
mertyildiran, line 59:
```suggestion
```

Template this path field instead;

https://github.com/kubeshark/kubeshark/blob/529ca63a47aaf11bdbd4c758918e85b2c55ade80/helm-chart/templates/10-ingress.yaml#L31

like:

```
path: {{ .Values.tap.ingress.path }}(.*)
```

or just `{{ .Values.tap.ingress.path }}` and we document it, make it clear that it supports regex.

Also add;

```
  annotations:
    nginx.ingress.kubernetes.io/use-regex: "true"
    nginx.ingress.kubernetes.io/rewrite-target: /$1
```

to annotations either here:

https://github.com/kubeshark/kubeshark/blob/529ca63a47aaf11bdbd4c758918e85b2c55ade80/helm-chart/templates/10-ingress.yaml#L6-L7

or as the default value of `tap.ingress.annotations` here:

https://github.com/kubeshark/kubeshark/blob/529ca63a47aaf11bdbd4c758918e85b2c55ade80/config/configStruct.go#L18

maybe the annotations should be just recommendation since it differs from the ingress controller to ingress controller.

We should seek for @corest's opinion.

mertyildiran (changes requested):

mertyildiran (approved):
Approving to lift the requested changes from me. Don't merge it just because I have approved.

helm-chart/templates/11-nginx-config-map.yaml
```diff
@@ -20,8 +20,8 @@ data:
       client_header_buffer_size   32k;
       large_client_header_buffers 8 64k;
 
-      location /api {
-        rewrite ^/api(.*)$ $1 break;
+      location {{ default "" (((.Values.tap).routing).front).basePath }}/api {
```
tiptophelmet, line 23: @alongir

alongir:
@tiptophelmet can you resolve the conflict?

alongir (approved):

alongir (approved):

PR #1727, opened 2025-03-24
🔧 Use regex-based `api,saml` locations without base path

tiptophelmet:


helm-chart/templates/11-nginx-config-map.yaml
```diff
@@ -46,6 +46,17 @@ data:
         proxy_pass_request_headers on;
       }
 
+{{- if (((.Values.tap).routing).front).basePath }}
```
tiptophelmet, line 49: @alongir

alongir:
@tiptophelmet suggest to close this PR

PR #1736, opened 2025-04-04
Decrease initial delay seconds

corest:
Towards https://github.com/kubeshark/worker/issues/653


alongir:
@iluxa as a reminder we change these values as it took the sniffer a long time to start. I believe we fixed this? Will the new values work?

iluxa:
> @iluxa as a reminder we change these values as it took the sniffer a long time to start. I believe we fixed this? Will the new values work?

Yes, long time start was fixed before. Last change to related code is https://github.com/kubeshark/worker/pull/612
New values should work.

PR #1757, opened 2025-05-21
eBPF capture didn't work in case of persistent storage

iluxa:
tracer and sniffer sync up using datadir
tracer creates sync files in datadir and worker waits for them
in case if files are already exist (non-`emptyDir`), sync mechanism fails

it must be more reliable sync mechanism is required to sync between tracer and sniffer than using filesystem


alongir (approved):

alongir:
fixed with https://github.com/kubeshark/kubeshark/releases/tag/v52.7.7

PR #1770, opened 2025-07-07
:hammer: Add `api2BetaEnabled` helm value

tiptophelmet:
Please, refer to corresponding tasks related to API2 architecture.


alongir (changes requested):
@tiptophelmet please change the name of the variable to something more generic, as we will use this for anything "beta". Feature will go in and out of beta based on maturity. Could be: `betaEnabled`

alongir (approved):

PR #1776, opened 2025-07-25
Adding hostNetwork inside values.yaml instead

noopduck:
Ref: https://github.com/kubeshark/kubeshark/issues/1766

In some situations it's desirable not to run this on the host network, eBPF mounts into /sys anyway.


corest:
Hi @noopduck 
Thx for your contribution.
Currently we do not support running in non-host network. While with eBPF having host-network is not a strict requirement, there are dependencies in other components, which require workers pods running in host network.
Kubeshark wouldn't work properly with that  option changed now. Therefore I'll close this PR and fill internal issue to also fix other dependencies on host-network.

PR #1777, opened 2025-07-29
chore(helm): use mirror.gcr.io for Kubeshark images to avoid Docker Hub pull limits

kahirokunn:
Docker Hub’s pull‑rate limits keep sending the whole world to **429 jail**:

* **Anonymous:** 100 pulls / 6 h / IP
* **Personal (free, logged‑in):** 200 pulls / 6 h ([Docker Documentation][1])

#### What’s changing

```diff
-    registry: docker.io/kubeshark
+    registry: mirror.gcr.io/kubeshark
```

Only the registry prefix changes—tags/digests stay identical.

#### Why `mirror.gcr.io`?

* **No Hub counting** – pulls are served from Google’s cache, so the Docker Hub limit never triggers.
* **Faster delivery** – images come from Google’s CDN‑backed infra.
* **Extra resilience** – decouples deployments from Docker Hub outages.

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -1,7 +1,7 @@
 # find a detailed description here: https://github.com/kubeshark/kubeshark/blob/master/helm-chart/README.md
 tap:
   docker:
-    registry: docker.io/kubeshark
+    registry: mirror.gcr.io/kubeshark
     tag: ""
     tagLocked: true
     imagePullPolicy: Always


corest:
Hi @kahirokunn . Sorry for a long delay with this change. 

You're absolutely right about the Docker Hub rate limiting pain.
However, I have concerns about the timing uncertainty that could impact our users:
The core issue: We push our releases directly to Docker Hub, but Google's mirror sync schedule isn't documented with specific intervals. While the mirror "keeps images in sync" and updates "at regular intervals," we don't have visibility into:

- How long after our Docker Hub push the image becomes available on mirror.gcr.io
- Whether there could be delays during high-traffic periods
- If less popular tags get deprioritized in the sync queue

When we announce a new Helm release, users expect immediate availability. If there's even a 15-30 minute delay between our Docker Hub push and the mirror sync, we'd have frustrated users getting image pull failures right after our release announcement.

As an alternative approach, I'll create issue to extend documentation on using `mirror.gcr.io` when possible to avoid rate limiting with 

```
tap:
  docker:
    registry: mirror.gcr.io/kubeshark
```

in values.yaml

kahirokunn:
Hi @corest .
Thanks for raising this! You’re right that it isn’t spelled out clearly in the docs, but in my own testing `mirror.gcr.io` was able to pull an image roughly **one second after I pushed** it upstream. That behavior strongly suggests it’s operating as a **pull-through cache**.

I also haven’t encountered any **rate limits** when pulling via `mirror.gcr.io`. For what it’s worth, it’s used in the wild—for example by Aqua Security (the folks behind Trivy).

Hope that context helps!

PR #1830, opened 2026-02-05
move the name of the data volume outside of the tap.tls if statement

mudged:
The helm chart will currently fail with the following error if `tap.tls` is set to `false` in the helm values.
```bash
    Error: INSTALLATION FAILED: 1 error occurred:
        * DaemonSet.apps "kubeshark-worker-daemon-set" is invalid: [spec.template.spec.volumes[3].hostPath: Forbidden: may not specify more than 1 volume type, spec.template.spec.containers[0].volumeMounts[2].name: Not found: "data"
```

This is because the start of the data volume is nested inside of the the if statement i.e. the `- name: data` is only output if `tap.tls` value is `true`. Otherwise, the attributes of the data volume are added to the previous volume definition.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,69 @@ clean: ## Clean all build artifacts.
 test: ## Run cli tests.
 	@go test ./... -coverpkg=./... -race -coverprofile=coverage.out -covermode=atomic
 
+test-integration: ## Run integration tests (requires Kubernetes cluster).
+	@echo "Running integration tests..."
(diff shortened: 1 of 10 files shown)


alongir:
Thanks so much @mudged for catching this bug and providing the fix! 🙌 

Great attention to detail in identifying that the volume name was incorrectly nested inside the TLS conditional block. This is exactly the kind of subtle templating issue that can be tricky to debug.

I've created issue #1837 to track this bug.

PR #1853, opened 2026-03-03
Add get_file_url and download_file MCP tools

alongir:
## Summary
- Adds `get_file_url` tool to resolve relative file paths (from tools like `export_snapshot_pcap`) into full download URLs
- Adds `download_file` tool to download files from Kubeshark to the local filesystem
- Adds file download workflow instructions to MCP server initialization
- Adds both tools to `--list-tools` CLI output
- Includes 6 new unit tests covering proxy mode, URL mode, error cases, and download functionality

## Test plan
- [x] `go test ./cmd/ -run TestMCP -v` — all 33 tests pass
- [x] `go build ./...` — compiles cleanly
- [ ] Manual: `kubeshark mcp --list-tools` shows both new tools


cmd/mcpRunner.go
```diff
@@ -706,6 +773,82 @@ func (s *mcpServer) callHubTool(toolName string, args map[string]any) (string, b
...
+	fullURL := strings.TrimSuffix(baseURL, "/") + filePath
+
+	// Determine destination file path
+	dest, _ := args["dest"].(string)
+	if dest == "" {
+		dest = path.Base(filePath)
+	}
+
+	// Download the file
+	resp, err := s.httpClient.Get(fullURL)
```
corest: I think this one is going to fail for large pcap download as default timeout is only 30s. It is better for this particular task to setup different http client with custom http.Transport and response header timeout and tls handshake timeout and then create new client with that transport and use to get fullURL

PR #1901, opened 2026-04-06
fix: add subPathExpr to worker DaemonSet for shared persistent storage

### LINKED ISSUE #1900: EFS persistent storage: workers crash with PebbleDB lock conflict
## Description

When using AWS EFS as persistent storage (`persistentStorage: true` + `persistentStorageStatic: true`), all worker DaemonSet pods share the same PVC and write PebbleDB to the same path `/app/data/captureDB`, causing file lock conflicts.

## Error

```
{"level":"fatal","error":"failed to create pebble db: resource temporarily unavailable","backend":"pebble","message":"Failed to initialize common storage for extensions"}
```

All workers except one enter CrashLoopBackOff with this error.

## Root Cause

The init container `cleanup-data-dir` correctly creates per-node directories (`/app/data/$NODE_NAME`) using `spec.nodeName`, but the `sniffer` and `tracer` containers mount `/app/data` directly without `subPathExpr`. This means all workers on all nodes write to the same `captureDB` path on the shared EFS volume.

With `emptyDir` this is not an issue because each pod gets its own isolated volume. But with a shared PVC (EFS ReadWriteMany), all pods see the same filesystem.

## Proposed Fix

Add `NODE_NAME` env var (from `spec.nodeName`) and `subPathExpr: $(NODE_NAME)` to the `sniffer` and `tracer` volume mounts in `helm-chart/templates/09-worker-daemon-set.yaml`, only when `persistentStorage` is enabled.

## Environment

- Kubeshark chart version: 53.1.0
- Kubernetes: EKS 1.34
- Storage: AWS EFS with aws-efs-csi-driver v2.1.15
- Region: eu-central-1
- Static PV with `volumeHandle` pointing to EFS filesystem ID

theechofive:
Fixes #1900

## Problem

When `persistentStorage` is enabled with a shared volume (e.g. **AWS EFS**), all worker DaemonSet pods mount `/app/data` without per-node isolation. PebbleDB in each sniffer attempts to lock the same `captureDB` path from multiple pods, resulting in:

```
failed to create pebble db: resource temporarily unavailable
```

All workers except one enter CrashLoopBackOff.

## Root Cause

The init container `cleanup-data-dir` already creates per-node directories (`/app/data/$NODE_NAME`), but the `sniffer` and `tracer` containers mount `/app/data` directly — without using `subPathExpr` — so all workers write to the same shared path.

## Changes

- Add `NODE_NAME` env var (from `spec.nodeName`) to `sniffer` and `tracer` containers
- Add `subPathExpr: $(NODE_NAME)` to `/app/data` volume mounts, conditionally only when `persistentStorage` is enabled

This ensures each worker writes to its own subdirectory on the shared volume, preventing file lock conflicts. The `emptyDir` path (when `persistentStorage` is false) is unaffected.

Co-Authored-By: Oz <oz-agent@warp.dev>


corest:
Thx for the contribution @theechofive ! 
We will test it this week and proceed with merging if no issues found

corest (approved):
Confirmed issue and confirmed fix working

corest (approved):
lgtm

PR #1907, opened 2026-04-13
:hammer: Create helm values for nginx proxy-buffer settings

tiptophelmet:


corest:
What is this for? There is no any issue linked. Why do we need this configurable instead of adding into nginx config directly?

PR #1913, opened 2026-04-20
fix(release-pr): sync bumped Chart.yaml to kubeshark.github.io

alongir:
## Summary

Two related improvements to the release Makefile:

### 1. Fix: sync bumped Chart.yaml to kubeshark.github.io

The old `release-pr` switched back to master (and pulled) BEFORE copying `helm-chart/` into `../kubeshark.github.io/charts/chart/`. That reverted the working tree to the **pre-bump** Chart.yaml, so the merged helm PR never contained the version bump. Downstream, the `chart-releaser` GitHub Action then tried to recreate an already-existing release tag (e.g. `kubeshark-53.2.2`) and failed with `422 already_exists`. This happened for v53.2.2 (fixed in kubeshark.github.io#5) and again for v53.2.3 (fixed in kubeshark.github.io#7).

Now the copy happens while the working tree still has the bumped Chart.yaml from `release/vX.Y.Z`, and kubeshark is switched back to master at the very end.

### 2. Split `release-pr` into three rerunnable targets

Before, a failure in any sub-step (sibling tagging, kubeshark PR, helm PR) required rerunning the whole recipe. Some sub-steps had no standalone entry point either — notably tagging `worker`/`hub`/`front` after a docker-image-only rebuild.

New layout:

| Target | What it does |
|---|---|
| `release-siblings` | Tag `worker`, `hub`, `front` with `v$(VERSION)`. Standalone for docker-image refreshes. |
| `release-pr-kubeshark` | Bump `Chart.yaml`, build, open release PR on `kubeshark`. |
| `release-pr-helm` | Sync `helm-chart/` to `kubeshark.github.io`, open helm PR. Requires `release/v$(VERSION)` (created by step 2); fetches it from origin if missing. |
| `release-pr` | Orchestrates all three in order. |

Each is idempotent: sibling tags, branches, commits, pushes, and PR creations all skip when the target state already exists. A shared `_release-check-version` prerequisite validates semver input once per invocation.

## Test plan

- [ ] `make release-siblings VERSION=x.y.z` with tags already present → logs "already on origin — skipping".
- [ ] `make release-pr-kubeshark VERSION=x.y.z` → bumps `Chart.yaml`, pushes `release/vx.y.z`, opens PR.
- [ ] Rerun the same → reuses branch, "nothing to commit", PR already exists.
- [ ] `make release-pr-helm VERSION=x.y.z` → checks out `release/vx.y.z`, copies bumped chart into `kubeshark.github.io`, opens helm PR with updated `Chart.yaml`.
- [ ] `make release-pr-helm VERSION=nonexistent` → errors with clear message pointing to `release-pr-kubeshark`.
- [ ] `make release-pr VERSION=x.y.z` end-to-end → both PRs created with the correct chart version.


Makefile
```diff
@@ -253,52 +253,121 @@ port-forward:
...
+	else \
+		echo "Chart.yaml already at $$NEW"; \
+	fi
 	@$(MAKE) build VER=$(VERSION)
 	@if [ "$(shell uname)" = "Darwin" ]; then \
 		codesign --sign - --force --preserve-metadata=entitlements,requirements,flags,runtime ./bin/kubeshark__; \
 	fi
 	@$(MAKE) generate-helm-values && $(MAKE) generate-manifests
-	@git checkout -b release/v$(VERSION)
+	@if git show-ref --verify --quiet refs/heads/release/v$(VERSION); then \
```
corest, line 304: This will do local check and if there is remote already - it will fall through to push that can fail if there is mismatch between local and remote. So it doesn't make rerun idempotent.

Makefile
```diff
@@ -253,52 +253,121 @@ port-forward:
...
+		echo "ERROR: release/v$(VERSION) branch not found locally or on origin."; \
+		echo "Run 'make release-pr-kubeshark VERSION=$(VERSION)' first."; \
+		exit 1; \
+	fi
+	@git checkout release/v$(VERSION)
+	@cd ../kubeshark.github.io && git checkout master && git pull \
+		&& rm -rf charts/chart && mkdir -p charts/chart \
+		&& cp -r ../kubeshark/helm-chart/ charts/chart/
+	@cd ../kubeshark.github.io && \
+	if git show-ref --verify --quiet refs/heads/helm-v$(VERSION); then \
```
corest, line 336: Same as above

corest (commented):
Won't block as current change improves release but still leaves cases where it can break

Makefile
alongir: Fixed — the branch is now deleted and recreated fresh on each run (`git branch -D` + `git checkout -b`), and the push uses `--force-with-lease` so it always succeeds even if remote diverged. See 6361cc73.

Makefile
alongir: Same fix applied here — delete + recreate `helm-v$(VERSION)` branch and `--force-with-lease` push. See 6361cc73.

corest (approved):

PR #1925, opened 2026-05-09
fix: compare semantic versions numerically

immanuwell:
## What
Fixes SemVersion.GreaterThan so version parts compare as numbers, not strings. Small footgun, but it made v1.16.0 look not greater than v1.9.0, yikes.

## Repro
On old code:
```sh
go test ./semver -run TestGreaterThan -v
```

The bad bit was:
```
1.16.0.GreaterThan(1.9.0) = false, want true
1.16.10.GreaterThan(1.16.9) = false, want true
```

## Checks
`make test` passes.

### DIFF
--- a/kubernetes/provider_test.go
+++ b/kubernetes/provider_test.go
@@ -1,0 +1,23 @@
+package kubernetes
+
+import (
+	"testing"
+
+	"github.com/kubeshark/kubeshark/semver"
+)
+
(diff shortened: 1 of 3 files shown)


corest:
Hi @immanuwell 
Thx for your contribution! We will include this into next release

corest (approved):

alongir (approved):
Correct minimal fix with tests; ignored `Atoi` errors are safe because `Breakdown` only yields digit strings (overflow is the only failure mode, irrelevant here).

PR #1937, opened 2026-05-19
:bookmark: Release v53.3.0

alongir:
Automated release PR for v53.3.0.


.github/workflows/release-tag.yml
```diff
@@ -15,6 +15,7 @@ jobs:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
+          token: ${{ secrets.GH_PAT }}
```
corest: whose token is this? What permissions it has? We agreed to move from static tokens to using application tokens for short-living scope

alongir: Good catch. This is a mistake.

alongir:
Addressed review from @corest:

- **Feedback**: Reviewer asks whose PAT token is used in release-tag.yml, what permissions it has, and notes the team agreed to move from static tokens to GitHub App tokens for short-lived scopes.
  **Action**: rejected
  **Detail**: This is a valid security/architecture concern, but it cannot be resolved with a simple code change. Migrating from `secrets.GH_PAT` to a GitHub App token (e.g., using `actions/create-github-app-token`) requires: (1) a GitHub App to be created and installed on the org with appropriate permissions, (2) the App ID and private key stored as repo/org secrets, and (3) agreement on which App to use. This should be tracked as a follow-up issue and discussed with the team. The current PAT usage works functionally — the token is needed because the default `GITHUB_TOKEN` cannot push tags that trigger other workflows (a known GitHub Actions limitation).

Rejected the feedback as not actionable via code — migrating to GitHub App tokens requires infrastructure setup and team coordination, recommended as a follow-up task.

corest (approved):

PR #1942, opened 2026-06-03
Permissions refactor: tap.auth.* roles + groupMapping

corest:
## Summary

CLI / chart-side counterpart of the hub permissions refactor. Reshapes `tap.auth.*` to the unified role model:

- `tap.auth.rolesClaim` — JWT claim or SAML attribute carrying group memberships.
- `tap.auth.defaultRole` — built-in or custom role applied when no claim value matches. Chart default: `kubeshark-viewer`.
- `tap.auth.groupMapping` — SSO group → role-name map. Built-in roles also identity-match.
- `tap.auth.roles` — operator-defined custom roles. Each declares a `capabilities` set (closed vocabulary) and a `namespaces` scope (comma list with `*` + glob support). Names starting with `kubeshark-` are reserved.
- Drops `AUTH_ROLES` env, drops the pre-refactor per-role action-flag schema (`canDownloadPCAP`, `scriptingPermissions`, etc.).

The top commit removes the cross-repo `plans/*.md` reference from the `AuthConfig.Roles` comment — those plans live as local working docs, not in the repo.

## Test plan

- [x] `make build` succeeds
- [x] Helm install with default values (`tap.auth.enabled: false`) — no regression
- [ ] Helm install with `tap.auth.enabled: true` + SAML — `groupMapping` resolves correctly
- [x] Helm install with `tap.auth.enabled: true` + OIDC + a custom role — custom role applies with namespace scope
- [x] Helm install with a custom role named `kubeshark-foo` — chart / hub rejects at startup


alongir (approved):
Clean permissions refactor — struct changes, defaults, configmap template, and values.yaml are all consistent. The `AUTH_ROLES` env var is preserved with the new `RoleConfig` shape (matching the hub-side refactor), `AUTH_GROUP_MAPPING` is added, and `DefaultFilter` removal from `AuthConfig` is correct (the separate `tap.defaultFilter` path is unaffected). Note: `manifests/complete.yaml` still carries the old schema and may need regeneration, but that's a follow-up, not a blocker.

PR #1946, opened 2026-06-16
cli: mark 'console' as under refactoring and early-exit

corest:
## Summary
The `console` command dials `ws://.../api/scripts/logs`, a hub route that no longer exists since hub PR #464 moved scripting-console log streaming onto a Connect-RPC streaming service (`ScriptLogsDashboard`). As a result the command never streams and spins in its reconnect loop.

This branch:
- Marks the command as non-functional / under refactoring in `Short`/`Long` help.
- Makes `RunE` **early-exit** with the under-refactoring notice instead of warning and then entering the broken reconnect loop, so users don't get stuck in a spinning client.

`runConsole`/`runConsoleWithoutProxy` are left in place (Go allows unused funcs) ready to be re-pointed at the Connect-RPC stream when the refactor lands.

## Test plan
- `go build ./cmd/` passes.
- `kubeshark console` now prints the notice and returns immediately.

### DIFF
--- a/cmd/console.go
+++ b/cmd/console.go
@@ -47,6 +47,7 @@ func init() {
 	consoleCmd.Flags().StringP(configStructs.ReleaseNamespaceLabel, "s", defaultTapConfig.Release.Namespace, "Release namespace of Kubeshark")
 }
 
+//nolint:unused // retained for the in-progress console refactoring; re-wired once the Connect-RPC client lands
 func runConsoleWithoutProxy() {
 	log.Info().Msg("Starting scripting console ...")
 	time.Sleep(5 * time.Second)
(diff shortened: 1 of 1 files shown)


alongir (approved):
Safe, well-scoped stopgap: `RunE` now prints an under-refactoring notice and returns instead of entering a dead reconnect loop. `utils.Yellow` is a valid format string (matches existing usage), all imports remain used, retained `runConsole`/`runConsoleWithoutProxy` are correctly `//nolint:unused` and don't break the build since Go doesn't error on unused functions. Minor (non-blocking): the retained functions and `runProxy` are now dead code that must be re-wired when the Connect-RPC migration lands, and the help text carries the disclaimer only until then.

PR #1949, opened 2026-07-07
feat(helm): add opt-in Prometheus Operator ServiceMonitor for metrics

somaz94:
Adds an opt-in Prometheus Operator `ServiceMonitor` for the worker and hub metrics services, so kube-prometheus-stack users can scrape Kubeshark metrics declaratively instead of hand-writing `additionalScrapeConfigs` (the approach currently documented in `metrics.md`).

It is disabled by default. When enabled, a single `ServiceMonitor` selects both `kubeshark-worker-metrics` and `kubeshark-hub-metrics` (via a new `app.kubeshark.com/metrics` label) and scrapes their `metrics` port. `interval`, `scrapeTimeout`, `path`, `scheme`, `honorLabels`, `relabelings` and `metricRelabelings` are optional passthroughs; empty values fall back to the Prometheus defaults.

`Chart.yaml` version is intentionally left unchanged (handled by the release PRs).

Validation (local, mirrors the helm-chart CI):
- `helm lint`: default + all 4 fixtures pass
- `helm unittest`: 37 tests pass (4 new servicemonitor tests)
- `kubeconform -strict -kubernetes-version 1.35.0`: default + fixtures all valid. The ServiceMonitor stays unrendered by default, so the strict CRD check is unaffected.


somaz94:
Rebased onto `master` to clear the BEHIND state — the only change is the new base; the diff is byte-identical to before (7 files under `helm-chart/`, +145/−0).

Note the check list is empty rather than passing: the push re-armed the fork approval gate, so `Linter`, `Test` and `CodeQL Advanced` are all sitting at `action_required` on the new head sha. Nothing has actually run.

Locally the chart tests pass — `helm unittest helm-chart -f 'tests/*_test.yaml'` → 4 suites, 37 tests, including the new `servicemonitor_test.yaml`. Could a maintainer approve the workflow run?

corest (changes requested):
MetricsConfig in config/configStructs/tapConfig.go:305 has only Port, and the PR does not add a ServiceMonitor field. Two consequences:

- Users installing via kubeshark tap cannot enable it, and serviceMonitor.enabled: true in ~/.kubeshark/config.yaml is dropped silently.
- values.yaml is generated from the struct (make generate-helm-values), so the next regeneration erases the whole serviceMonitor block from the chart.

somaz94:
Good catch, thanks. I added a ServiceMonitorConfig to MetricsConfig, so the block now comes from the struct and reaches the chart through kubeshark tap and config.yaml as well. Regenerating values.yaml locally with the new binary gives a file identical to the committed one, and a new test in config_internal_test.go fails if the field is ever dropped again.

corest (approved):

alongir (approved):
Opt-in ServiceMonitor is correctly gated, selector/port names match the rendered metrics services, config struct round-trips through values generation (new Go test guards it), and no import cycle is introduced; `manifests/complete.yaml` isn't regenerated for the new service label, but that's produced by the release target, not CI.

PR #1955, opened 2026-08-07
helm: gate the worker hub token on the same condition as AUTH_ENABLED

### LINKED ISSUE #1954: demoModeEnabled enables hub auth but not the worker hub token: all worker->hub calls 401
## Summary

`demoModeEnabled: true` makes the hub enforce auth, but the worker's internal hub token is gated on a *different* condition. With demo mode on and `tap.auth.enabled` left at its default (`false`), the hub demands a bearer token that the workers were never issued, and every worker → hub call returns **401**.

Capture and streaming keep working, so nothing looks broken — but tracer target refresh and name resolution both stop, silently.

## The mismatch

`templates/12-config-map.yaml` (L21-25) — `AUTH_ENABLED` is `demoModeEnabled` **OR** `tap.auth.enabled`:

```yaml
AUTH_ENABLED: '{{- if and .Values.cloudLicenseEnabled (not (empty .Values.license)) -}}
                    {{ (default false .Values.demoModeEnabled) | ternary true (...) }}
              {{- else -}}
                    {{ .Values.cloudLicenseEnabled | ternary "true" ((default false .Values.demoModeEnabled) | ternary "true" .Values.tap.auth.enabled) }}
              {{- end }}'
```

`templates/09-worker-daemon-set.yaml` (L149, L242, L442) — the token is gated on `tap.auth.enabled` **alone**:

```yaml
{{- if (((.Values.tap).auth).enabled) }}
- name: HUB_INTERNAL_TOKEN_PATH
  value: /var/run/secrets/kubeshark/hub-token/token
{{- end }}
```

...and likewise for the volumeMount and the projected `serviceAccountToken` volume.

So `demoModeEnabled: true` + `tap.auth.enabled: false` (the default) produces a hub that requires a token and workers that have none.

## Symptoms

```
ERR kubernetes/resolver/resolver.go:351 > Failed to handle kubeshark configmap modified
  error="failed to ensure name resolution history: hub returned HTTP 401
         for url=http://kubeshark-hub:80/resolver/history (auth/config?): <empty>"
```

Verified live from inside a worker — both internal endpoints reject:

```
$ wget -S -O /dev/null http://kubeshark-hub:80/resolver/history   -> HTTP/1.1 401 Unauthorized
$ wget -S -O /dev/null http://kubeshark-hub:80/pods/all           -> HTTP/1.1 401 Unauthorized
```

It recurs on every sync cycle, not just at startup.

### The `/pods/all` failure is disguised

The 401 comes back with an empty body, which the tracer then tries to parse as JSON, so it surfaces as a completely unrelated-looking error:

```
ERR tracer/pkg/kubernetes/target.go:77 > Failed to get all targeted pods
  error="failed unmarshalling list of target pod from url=http://kubeshark-hub:80/pods/all:
         unexpected end of JSON input"
```

Nothing there points at auth. Worth handling a non-2xx status before unmarshalling, independently of the gating fix.

## Impact

- Tracer cannot refresh its targeted-pod list → TLS hooking does not pick up newly started pods.
- Name resolution history never syncs → IPs may not resolve to pod/service names in the UI.
- Packet capture and streaming are unaffected, which is what makes it easy to miss: the product looks alive and healthy.

## Suggested fix

Gate the worker token wiring on the same condition that drives `AUTH_ENABLED`, rather than on `tap.auth.enabled` alone. A shared helper would keep the two from drifting again, e.g.:

```yaml
{{- define "kubeshark.authEnabled" -}}
... single source of truth ...
{{- end -}}
```

consumed by both `12-config-map.yaml` and `09-worker-daemon-set.yaml`. Any deployment that turns auth on by a route other than `tap.auth.enabled` will hit this same class of bug otherwise.

## Workaround

Set both:

```yaml
demoModeEnabled: true
tap:
  auth:
    enabled: true
```

Confirmed: this adds only `HUB_INTERNAL_TOKEN_PATH`, the mount and the projected `serviceAccountToken` (audience `kubeshark-hub`) to the worker DaemonSet — every `AUTH_*` config value renders identically. After applying, the same request with the mounted token returns `HTTP/1.1 200 OK` and the 401s stop.

## Environment

- Chart from `master` (`helm-chart/`), `Chart.yaml` 53.3.0; hub/worker/front built from `master`
- EKS 1.36
- `demoModeEnabled: true`, `cloudLicenseEnabled` false, `tap.auth.type` resolves to `default`
- `tap.auth.enabled` unset (chart default `false`)

corest:
Fixes #1954 (chart half).

## The bug

The hub's `AUTH_ENABLED` is derived from `cloudLicenseEnabled` / `demoModeEnabled` / `tap.auth.enabled`, but the worker DaemonSet gated `HUB_INTERNAL_TOKEN_PATH`, its volumeMount and the projected `serviceAccountToken` on `tap.auth.enabled` **alone**. With `demoModeEnabled: true` and `tap.auth.enabled` unset the hub demanded a bearer token the workers never received, and every worker → hub call returned 401.

Note this is not limited to demo mode: `cloudLicenseEnabled` defaults to `true`, so `helm template` with stock values already produces `AUTH_ENABLED: 'true'` and no token.

## Changes

1. **`kubeshark.authEnabled` helper** — the `AUTH_ENABLED` expression extracted verbatim into `_helpers.tpl`, consumed by both `12-config-map.yaml` and `09-worker-daemon-set.yaml` so the two cannot drift again.
2. **Token wiring in the tracer container too** — the tracer polls `/pods/all` and `/pods/targeted` on every sync cycle, but only the sniffer container got the env var and mount.

## Draft, because

The tracer container change is inert without kubeshark/tracer2#68, which teaches the tracer to send the header. The helper change stands on its own and fixes the resolver 401s.

## Testing

- `helm template` across all 48 combinations of `cloudLicenseEnabled` × `license` × `demoModeEnabled` × `tap.auth.enabled` × `tap.auth.type`: `AUTH_ENABLED` renders byte-identical to master in every one. Only the worker token wiring changes.
- With `tap.tls=true` both containers get `HUB_INTERNAL_TOKEN_PATH` + the mount; with `tap.tls=false` only the sniffer, as before.
- `helm lint` clean.

### DIFF
--- a/helm-chart/templates/09-worker-daemon-set.yaml
+++ b/helm-chart/templates/09-worker-daemon-set.yaml
@@ -284,6 +284,10 @@ spec:
             value: '{{ (include "sentry.enabled" .) }}'
           - name: SENTRY_ENVIRONMENT
             value: '{{ .Values.tap.sentry.environment }}'
+          {{- if eq (include "kubeshark.authEnabled" .) "true" }}
+          - name: HUB_INTERNAL_TOKEN_PATH
+            value: /var/run/secrets/kubeshark/hub-token/token
+          {{- end }}
           resources:
             limits:
               {{ if ne (toString .Values.tap.resources.tracer.limits.cpu) "0" }}
@@ -359,6 +363,11 @@ spec:
               mountPropagation: HostToContainer
               name: root
               readOnly: true
+            {{- if eq (include "kubeshark.authEnabled" .) "true" }}
+            - mountPath: /var/run/secrets/kubeshark/hub-token
+              name: hub-internal-token
+              readOnly: true
+            {{- end }}
       {{- end }}
       dnsPolicy: ClusterFirstWithHostNet
       hostNetwork: {{ .Values.tap.hostNetwork }}
--- a/helm-chart/templates/09-worker-daemon-set.yaml
+++ b/helm-chart/templates/09-worker-daemon-set.yaml
@@ -146,7 +146,7 @@ spec:
             value: '{{ (include "sentry.enabled" .) }}'
           - name: SENTRY_ENVIRONMENT
             value: '{{ .Values.tap.sentry.environment }}'
-          {{- if (((.Values.tap).auth).enabled) }}
+          {{- if eq (include "kubeshark.authEnabled" .) "true" }}
           - name: HUB_INTERNAL_TOKEN_PATH
             value: /var/run/secrets/kubeshark/hub-token/token
           {{- end }}
@@ -239,7 +239,7 @@ spec:
 {{- if .Values.tap.persistentStorage }}
               subPathExpr: $(NODE_NAME)
 {{- end }}
-            {{- if (((.Values.tap).auth).enabled) }}
+            {{- if eq (include "kubeshark.authEnabled" .) "true" }}
             - mountPath: /var/run/secrets/kubeshark/hub-token
               name: hub-internal-token
               readOnly: true
@@ -439,7 +439,7 @@ spec:
           emptyDir:
             sizeLimit: {{ .Values.tap.storageLimit }}
 {{- end }}
-        {{- if (((.Values.tap).auth).enabled) }}
+        {{- if eq (include "kubeshark.authEnabled" .) "true" }}
         - name: hub-internal-token
           projected:
             sources:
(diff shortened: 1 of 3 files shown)


alongir (approved):
Correct, well-tested fix — extracts `AUTH_ENABLED` into a shared `kubeshark.authEnabled` helper so the worker token wiring and the hub's auth flag can't drift; gate/value are byte-consistent and volume/mount conditions align in all combinations. Tracer container change is inert until tracer2#68 (as noted for the Draft) but harmless.

PR #1960, opened 2026-08-21
docs: explain epoch conversion in network RCA skill

### LINKED ISSUE #1923: Add epoch conversion hint to network-rca skill data boundaries section
### Description

The `network-rca` skill's "Check Data Boundaries" section (around line 143) should include an explicit hint about converting Unix epoch milliseconds returned by `get_data_boundaries`.

Currently, there's no guidance on how to interpret the epoch timestamps, which leads to off-by-one timezone errors that silently shift the investigation window by hours.

**Proposed addition** — insert after the example response table (line 162), before the data rotation note:

> **Epoch conversion rule**: `get_data_boundaries` returns Unix milliseconds.
> Always convert programmatically — never estimate or do mental math:
>
> ```bash
> python3 -c "from datetime import datetime, timezone, timedelta; ts=EPOCH_MS/1000; local=timezone(timedelta(hours=OFFSET)); print(datetime.fromtimestamp(ts, tz=local))"
> ```
>
> Off-by-one timezone errors silently shift the investigation window by hours.

### Why Is This Needed?

When an AI agent receives epoch milliseconds from `get_data_boundaries`, it may attempt mental math or rough estimation for timezone conversion, resulting in queries that target the wrong time window. This is a silent failure — the query succeeds but returns irrelevant or empty results, derailing the investigation. An explicit conversion rule prevents this class of errors.

kVinsom:
## Summary

- document that `get_data_boundaries` returns Unix epoch milliseconds
- add a programmatic timezone-conversion example
- warn that manual conversion can silently shift the investigation window

## Testing

- `git diff --check`

Closes #1923

### DIFF
--- a/skills/network-rca/SKILL.md
+++ b/skills/network-rca/SKILL.md
@@ -168,6 +168,15 @@ Per node:
   └─────────────────────────────┴───────────────────────────────┴───────────────────────────────┘
 ```
 
+> **Epoch conversion rule**: `get_data_boundaries` returns Unix milliseconds.
+> Always convert programmatically — never estimate or do mental math:
+>
+> ```bash
+> python3 -c "from datetime import datetime, timezone, timedelta; ts=EPOCH_MS/1000; local=timezone(timedelta(hours=OFFSET)); print(datetime.fromtimestamp(ts, tz=local))"
+> ```
+>
+> Off-by-one timezone errors silently shift the investigation window by hours.
+
 If the incident falls outside the available window, the data has been rotated
 out. Suggest increasing `storageSize` for future coverage.
 


kVinsom:
Hi! Friendly ping when you have a moment. The workflows are still awaiting maintainer approval. I’m happy to address any feedback or requested changes. Thanks!

corest (changes requested):
- The same guidance applies to get_l7_data_boundaries, which uses the identical oldest_ts/newest_ts fields. The note should cover both tools, or sit under a shared heading, rather than only the L4 section.
- The "Example response" block above it should be relabelled as presented output, or replaced with the real JSON shape, otherwise the two blocks contradict each other (one shows formatted strings, the next says the tool returns epoch ms).

PR #1961, opened 2026-08-21
feat: validate configuration schema version

### LINKED ISSUE #1378: Add a version to the config file and alert on backward compatibility
Yeah, just something to say "This version of CLI doesn't work with this config file" and spit out the command to regenerate it

[Slack Message](https://kubeshark.slack.com/archives/C04BNKU5SNM/p1688409147126669?thread_ts=1688408752.479139&cid=C04BNKU5SNM)

kVinsom:
## Summary

- add an explicit configuration schema version independent of the Kubeshark release version
- accept unversioned legacy files as schema v1 and reject older, newer, or malformed configurations with regeneration guidance
- decode into a temporary copy so validation is atomic and each file is parsed only once
- allow `kubeshark config -r` to replace malformed or incompatible configuration without loading it first
- keep Helm values synchronized and document compatibility and schema-evolution rules
- add table-driven coverage for validators, constructors, writing, loading, regeneration, Helm values, and CLI errors

## Testing

- `go test ./config`
- `go test ./...`
- `go vet ./...`
- `go test ./config -cover` — 76.4%

Closes #1378

### DIFF
--- a/config/README.md
+++ b/config/README.md
@@ -1,0 +1,28 @@
+# Configuration schema versioning
+
+The top-level `version` value identifies the CLI configuration schema. It is
+independent of the Kubeshark release version: compatible application releases
+must continue to use the same schema version.
+
+## Compatibility rules
+
+- Files without `version` predate schema versioning and are treated as schema
+  version 1.
+- The CLI loads only the schema version it explicitly supports. An incompatible
+  file produces an error that points to `kubeshark config -r`.
+- `kubeshark config -r` does not load the file it replaces, so regeneration also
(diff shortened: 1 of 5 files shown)


kVinsom:
Hi! Friendly ping when you have a moment. The workflows are still awaiting maintainer approval. I’m happy to address any feedback or requested changes. Thanks!

corest:
Internal expectation on this validation has changed so #1378 became slightly outdated. Because of that we can't accept this PR as is for now. We would need to update initial issue description first.

PR #1962, opened 2026-08-25
Add support for extra objects in helm chart

corest:
Allow extra object deployments together with predefined resources


alongir (approved):
Clean, well-tested addition of `tap.extraObjects` — the template correctly handles both mapping and raw-string entries with root context (`$`), the Go config field is properly defaulted, and the change is backed by unit tests plus lint/kubeconform CI validation with no correctness, security, or breaking-change concerns.

PR #1969, opened 2026-09-21
Link testing methodology from release notes

corest:
Adds a public Benchmark Methodology link to the release-body template so newly generated release notes include it. The text identifies its scope as Hub streaming benchmarks and does not imply a validation pass.

Verified the documentation URL returns HTTP 200 and contains the expected title. Confirmed the Makefile renders this template into bin/README.md, which the release workflow uses as its body. git diff --check passed.


corest:
Closing because we update the release notes after the release is published, so there is no need to update this initial template. The testing methodology link is included in the perfshark release-notes example for use when composing the final release notes.

PR #1970, opened 2026-09-23
chart: add endpoint and path-style options for S3-compatible snapshot storage

corest:
## Summary
Adds `tap.snapshots.cloud.s3.endpoint` and `tap.snapshots.cloud.s3.forcePathStyle` so snapshots can be stored on S3-compatible object stores such as NetApp StorageGRID, Dell ECS and MinIO. They map to `SNAPSHOT_AWS_ENDPOINT` and `SNAPSHOT_AWS_FORCE_PATH_STYLE`, which the hub reads as of kubeshark/hub#857.

- CLI config struct, `values.yaml` regenerated with `bin/kubeshark__ config`
- cloud-storage ConfigMap template and the hub deployment's inline-config condition
- chart README, `docs/snapshots_cloud_storage.md` (new S3-compatible section), install skill references
- two chart unit tests

## Test plan
- [x] `helm unittest ./helm-chart` (85 tests) and `helm lint`
- [x] `go test ./config/...`
- [x] Rendered with StorageGRID-style values: ConfigMap carries the endpoint and `SNAPSHOT_AWS_FORCE_PATH_STYLE: "true"`, hub gets `-cloud-storage-provider s3` and both envFrom refs
- [x] End to end with MinIO on an EKS sandbox cluster and the hub branch image: rendered ConfigMap carried endpoint and path-style flag, hub initialised the provider against MinIO and uploaded a snapshot


alongir (approved):
Purely additive; follows the existing conditional-ConfigMap pattern exactly, defaults (`""` / `false`) keep current renders unchanged, and both the ConfigMap and hub `envFrom` guards were updated in step — no showstoppers. Only nit: `--set-string tap.snapshots.cloud.s3.forcePathStyle=false` would render `SNAPSHOT_AWS_FORCE_PATH_STYLE: "true"` (non-empty string is truthy in the template), same as every other bool in this chart.

